PR #19, opened 2022-07-05
feat: Added Chinese (Simplified) translations

e5u:


Sculas:
The manager is going to be rewritten, meaning all translations will be invalidated. So I'm going to close this PR. Thank you for your contribution!

PR #21, opened 2022-07-05
feat: recode of the manager

Aunali321:

PR #22, opened 2022-07-06
feat: move patching to foreground service

danthe1st:
Currently, ReVanced Manager uses the `viewModelScope` in order to do patching in `PatcherSubscreen.startPatcher`.
While parts of the patching use coroutines, not all do. This would result in the whole `viewModelScope` being blocked when Androlib or other blocking functions (that is not using coroutines) are invoked.
Furthermore, this is not done in a foreground service. This means that the whole patching process is interrupted as soon as the user opens another app and Android decides to kill the app.

This PR moves patching to a foreground service so that patching is possible in the background without needing to worry about Android destroying the activity.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -9,6 +9,7 @@
     <uses-permission android:name="android.permission.REQUEST_DELETE_PACKAGES" />
     <uses-permission android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" />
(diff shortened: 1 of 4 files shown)


danthe1st:
included in https://github.com/Aunali321/revanced-manager

danthe1st:
Duplicate of #21

Canny1913:
image
uhhhh

danthe1st:
Nevermind. However, that still might cause conflicts since all commits from this branch are also present in the other one.

danthe1st:
https://github.com/Aunali321/revanced-manager/pull/3 added an improved version of this PR leading to this being a duplicate of #21 for real.

PR #32, opened 2022-07-27
Fix disabled resourcepatching

danthe1st:
If resource patching is disabled, a NPE is thrown in the line `file.copyEntriesFromFileAligned(ZipFile(result.resourceFile!!), ZipAligner::getEntryAlignment)` in `PatcherWorker`. This PR fixes that using a `?.let` construct.


danthe1st:
wrong repository

PR #128, opened 2022-09-16
feat: added german translation

Janik313:


Ushie:
Translations for ReVanced Manager can be done on [Crowdin](https://translate.revanced.app/)

PR #137, opened 2022-09-16
feat: UI improvements

55nknown:


55nknown:
actually, it's edgeToEdge what I intended, not immersive

PR #179, opened 2022-09-17
Update build.gradle

MikeAinslie20:


Aunali321:
Manager doesn't work on older android versions.

PR #196, opened 2022-09-18
fix: delete tests

55nknown:
Flutter tests aren't used currently.


ponces:
Meh... Why is this needed?

55nknown:
Why have tests if you're not using them?

ponces:
Meh... Default Flutter template creates them and we should have in the future...
Anyway, I can merge this. Please amend commit message to use `fix` scope instead of `misc`

55nknown:
Done.

ponces:
Thanks!

PR #197, opened 2022-09-18
fix: navigation bar color

55nknown:
## Before:
image

## After:
image


ponces:
What is the result of these changes on a three or two buttons system navbar?

55nknown:
I also tested them, but it won't look the same, since android defines partially opaque background for the navbar, even when telling it to be transparent.

## This is how it looks:
Screenshot

ponces:
Thanks!

PR #208, opened 2022-09-18
feat: CustomScrollView in app and patches selector views

55nknown:
## Back button and title:
image

## When scrolled:
image

### DIFF
--- a/assets/i18n/en.json
+++ b/assets/i18n/en.json
@@ -66,11 +66,13 @@
         "widgetSubtitle": "We are online!"
     },
     "appSelectorView": {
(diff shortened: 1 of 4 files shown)


ponces:
Not sure we want this...
What do you think @Aunali321?

Aunali321:
I dont like it either but every material 3 app has it.

55nknown:
The primary reason for this is to have a way to navigate back without using the system navigation bar. Also more padding at the top looks better in my opinion.

ponces:
Thanks!

PR #211, opened 2022-09-18
feat: combine org and repo in source settings

55nknown:
- combined the source org and repo into a single input field

## Example:
image


ponces:
This was implemented like this before but this is not so versatile as the user might want to use a different org for one and another source.
Thank you, but this limits current feature. Closed.

55nknown:
> This was implemented like this before but this is not so versatile as the user might want to use a different org for one and another source.
> Thank you, but this limits current feature. Closed.

You can still specify different organisations (or GitHub users) as sources, because the patches and integrations are still separated.

Aunali321:
@ponces should we merge this

PR #223, opened 2022-09-19
Create ko.json

DDinghoya:
Added  "koreaOption": "한국어", of 108 lines for Korean

Need "koreaOption": "Korean" between 107 and 108 lines on en.json

### DIFF
--- a/assets/i18n/ko.json
+++ b/assets/i18n/ko.json
@@ -1,0 +1,154 @@
+{
+    "okButton": "확인,
+    "cancelButton": "취소",
+    "enabledLabel": "활성화",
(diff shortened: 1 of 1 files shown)


KiGamji:
translations would be approved for crowdin. languages should be each in their own language. ty for language string btw, i will add it to my languages options commit

DDinghoya:
> translations would be approved for crowdin. languages should be each in their own language. ty for language string btw, i will add it to my languages options commit

Thank you

Next time, I will work on crowdin

Ushie:
Translations for ReVanced Manager can be done on [Crowdin](https://translate.revanced.app/)

PR #241, opened 2022-09-20
feat: show warning when apk min-sdk is higher than device sdk

shabbyplace:


shabbyplace:
Only YouTube requires Android 8.0 and up. Older Android versions are still supported by YouTube Music, Twitter, and Tiktok. However, because the manager sdk is configured for Android 8.0. Users of older devices cannot use the patcher to patch the these apps.

Aunali321:
@ponces what do you think about this?

ponces:
Why does this have an impact of min sdk of manager itself?

shabbyplace:
There are still Android 6 and 7 devices used by some. TikTok, for example, continues to support these versions. If the manager only supports Android 8 and up, these devices will be unable to install the manager. As a result, they are unable to patch other apps that still support their Android version.

Ushie:
> There are still Android 6 and 7 devices used by some. TikTok, for example, continues to support these versions. If the manager only supports Android 8 and up, these devices will be unable to install the manager. As a result, they are unable to patch other apps that still support their Android version.

ReVanced Manager only works with SDK 26 and up, otherwise it'd of already had a lower min SDK

PR #296, opened 2022-09-25
fix: tweak card appearances

afnzmn:


mnrsuristic:
I've a suggestion about the when you click to update Vanced manager app it doesn't show you the process bar at all hence I don't know whether it finished updating or not

PR #375, opened 2022-10-09
fix: use permanent keystore path if available

aliernfrog:
Keystore files are currently located in cache, which gets deleted often.
This PR solves it by saving keystores in `externalStorageDirectory` (`Android/data/app.revanced.manager.flutter/files`), also gives users ability to backup keystore files.


Aunali321:
Tested it and it works as intended. Thank you!

Issue #140, opened 2022-09-16
Option to backup (or alternatively, export packages), post-build


Thisisauser6443:
### Type

Functionality

### Branch

Flutter

### Issue

In my opinion, it's not so much as one of the **largest** problems, but while building an APK is an easy process, in some cases, especially my experience with YouTube (`17.33.42`, at the time), the process does take a while with attempting to apply a multitude of patches, to it, and can be tedious for some _(Including myself, lol_).

### Feature

What I think could be implemented is a way to simply **backup/export the packages to the internal storage**.

Let's say, after attempting to see if a patch can function with an unrecommended build number of YT, it fails to launch/has major usability bugs, I could simply revert to the _last known good package_, without needing to rebuild, again, in less than a minute (give or take), rather than needing to try and remember all of the patches you had/didn't have, off the top of your head

And believe me, with the few braincells I have left, it's getting harder and harder

### Motivation

Truth is, a lot of us are, without admitting to it, impatient, and, like I said, this would help with interchangeability between older/newer builds of patched programs. The only issue with my request, AFAIK, is whether this could probably lead to people distributing packages, but I'm wondering what others will make of this

nkitsaini:
I agree with you. Having an export is an easy option which I was surprised to see is not present. I'll take a look at the code and possibly open a PR if others agree.

Aunali321:
@nkitsaini do you still planning on opening a PR?

nkitsaini:
Yes, I'm most probably create the PR by the coming week. @Aunali321

PR #439, opened 2022-10-25
feat: Adding support for Exporting APK

nkitsaini:
Allows users to export APK once it's patched. 
I'm using [ACTION_CREATE_DOCUMENT](https://developer.android.com/reference/android/content/Intent#ACTION_CREATE_DOCUMENT) intent for export. This has it's benefits.

* We get to populate initial name for exported file, but user can edit before exporting
* We don't need to care about permissions and folder locations. We get a ContentUri and we just have to write apk there.
* Export is not limited to local filesystem. APK can be exported to other implementations like Google Drive out of the box.

Fixes: #140

### DIFF
--- a/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
+++ b/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
@@ -1,8 +1,5 @@
 package app.revanced.manager.flutter
 
-import android.app.Activity
-import android.content.Intent
-import android.net.Uri
 import android.os.Build
 import android.os.Handler
 import android.os.Looper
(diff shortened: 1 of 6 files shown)


proxykingdev:
fantastic. i was waiting for this. very useful

Ushie:
is this still being worked on?

nkitsaini:
@Ushie Yes. I've cleaned up the PR. It is ready for review.

Aunali321:
@nkitsaini any reason for using platform channels (kotlin) instead of dart? i am pretty sure the feature you're trying to implement is possible in dart. The reason we avoid platform channels is because they are very slow compared to dart.

nkitsaini:
@Aunali321 I'm relatively new to android/flutter development so please correct me if I'm wrong. But as I understand it, if you need to get the Activity Result then you have to do it using native code.

Ideally there should a flutter library that wraps this whole logic and exposes via a dart api but I couldn't find it.

There's [android_intent_plus](https://pub.dev/packages/android_intent_plus), but it only seems to launch the `Intent` but not return the result.
I was able to find one library [cr_file_saver](https://pub.dev/packages/android_intent_plus) that let's you export files from dart but it's very new (less then a month old), so not sure about its stability.

I'll search more about Android Activity and how they work with flutter.

Aunali321:
can you test if `cr_file_saver` works? if so we can avoid platform channels. and using new packages isnt a problem in alpha. we can always revert hence "alpha". Thank you!

nkitsaini:
@Aunali321

It mostly works. There's just a minor bug around initial file name for export but I've added a [work around](https://github.com/revanced/revanced-manager/pull/439/files#diff-e0ef5e8d22a0beaf4e08655dcc8fe9a39b15a73d908b33e1f911c7225513e1d8R238) for now.


Overall the implementation is much simpler and I'll also try to get this bug fixed on `cr_file_saver`.

Aunali321 (approved):

PR #466, opened 2022-11-02
feat: show selected patches count in patches selector view & patch selector card

Mipirakas:
This is nice to have when updating youtube revanced and you know how many patches you currently have applied.
It's easy to see if you forgot one or not instead of manually counting.
Feedback on how to make the code better very much appreciated.


Aunali321:
Thank you!

PR #481, opened 2022-11-04
fix: delete cached apk files when picking new one

Mipirakas:
When picking an apk file with FilePicker from storage, the file gets cached. This is fine, but when you select a newer version of that file (because the patches now support a newer version of eg. youtube), the old apk file does not get removed from cache. I accumulated several youtube apks in my cache taking up a considerable amount of storage.

This PR simply deletes all other apks in the cache when selecting a new one.


Aunali321 (approved):

Aunali321:
Thank you!

PR #482, opened 2022-11-05
fix: update workflow actions

ThexXTURBOXx:
This PR updates nearly all workflow actions (except for one which is abandoned, namely `marvinpinto/action-automatic-releases`).
Why is this needed?
Node.js 12 actions are deprecated and support for them could be removed at any point, see warnings like in [this](https://github.com/revanced/revanced-manager/actions/runs/3397952348) workflow.

I will also create PRs like this one in the other repos using outdated actions.

### DIFF
--- a/.github/workflows/crowdin.yml
+++ b/.github/workflows/crowdin.yml
@@ -16,10 +16,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


ThexXTURBOXx:
I finished updating all the important repos. Feel free to merge them or ignore them :)

PR #499, opened 2022-11-10
feat: Add spanish support

mauriciabad:
Add Spanish (from Spain) translations to i18n.

> So far, I've only added half of the strings. I'll finish it latter.


Aunali321:
Hello!, translations for manager are being accepted at https://translate.revanced.app/ .

mauriciabad:
@Aunali321 Okay, I'll contribute there then.

To prevent other people doing the same as me, you should explain how to add translations, or if it's already explained make it more discoverable. 

I would expect this information to be in the readme, inside a "contribute" section. Or having a readme inside the i18n folder.

PR #500, opened 2022-11-11
Add files via upload

wangyi1573:
I translated the basic English into Chinese and made a JSON file, hoping to achieve localization as soon as possible. Some of them did not know how to translate, so they kept the original English. There will be better translations in the future.


kazimmt:
Upload it to ReVanced translation site
https://translate.revanced.app

PR #521, opened 2022-11-18
fix: update broken hint

milksense:


JuniorWolfgamingDE:
You need to change "Update settings_manage_api_url.dart" to "fix: update settings_manage_api_url.dart" or else it won't show as compatible by Semantic PR.

PR #535, opened 2022-11-24
fix: the issue that makes `manager` apply the wrong patch when there are duplicate patch names.

d4rkk3y:
This PR fix a issue that manager possibly apply wrong patch when there are 2 or more patch intended for different package name but have same patch name.
Issue: https://github.com/revanced/revanced-patches/pull/1082
If this issue was fixed then this https://github.com/revanced/revanced-patches/pull/1085 can be merged.

### DIFF
--- a/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
+++ b/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
@@ -11,6 +11,7 @@ import app.revanced.manager.flutter.utils.zip.ZipFile
 import app.revanced.manager.flutter.utils.zip.structures.ZipEntry
 import app.revanced.patcher.Patcher
 import app.revanced.patcher.PatcherOptions
+import app.revanced.patcher.extensions.PatchExtensions.compatiblePackages
 import app.revanced.patcher.extensions.PatchExtensions.patchName
 import app.revanced.patcher.logging.Logger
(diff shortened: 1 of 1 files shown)


Canny1913:
Was this PR tested?

d4rkk3y:
> Was this PR tested?

I tested with current official patch file and it worked fine. I did not test with patch from https://github.com/revanced/revanced-patches/pull/1085 because I did not know how to import custom patch file to revanced manager. How can I do that?

Canny1913:
> > Was this PR tested?
> 
> I tested with current official patch file and it worked fine. I did not test with patch from [revanced/revanced-patches#1085](https://github.com/revanced/revanced-patches/pull/1085) because I did not know how to import custom patch file to revanced manager. How can I do that?

Use the custom sources option in settings and change the repo there

d4rkk3y:
> Use the custom sources option in settings and change the repo there

After a hour struggling with semantic release, I tested it and it works as expected with https://github.com/revanced/revanced-patches/pull/1085 which have youtube and tiktok settings patch name 'settings'.

Canny1913:
> > Use the custom sources option in settings and change the repo there
> 
> After a hour struggling with semantic release, I tested it and it works as expected with [revanced/revanced-patches#1085](https://github.com/revanced/revanced-patches/pull/1085) which have youtube and tiktok settings patch name 'settings'.

Yeah semantic is absolute pain.

PR #544, opened 2022-11-26
fix(dashboard): stop showing patched apps

FineFindus:
@Canny1913 asked me to pr this.

This removes the patched apps in the dashboard, since it generates a lot of spam issues.


Aunali321:
I do not agree with this @Canny1913.

Aunali321:
I will probably fix this next when i am free.

PR #631, opened 2023-01-05
Added Guide for Dummies

dsouzavivian13:
This is a guide posted on Reddit and I think it would be really helpful for Noobs like me to install app patches through the ReVanced Manager app.


oSumAtrIX:
Adding a PDF is unconventional. The docs branch should be used.

Issue #563, opened 2022-12-05
Export keystore


SodaWithoutSparkles:
### Type

Functionality

### Issue

When reverting to previous versions of the manager you will need to delete the app and then install the old version again, thus deleting the Keystore. 
People who are migrating between phones might not have the original Keystore moved as well, so they will have to delete patched apps to update again, losing all configs
On rare cases where the Keystore was lost/deleted for whatever reason, this could serve as a backup to not lose the  Keystore

### Feature

Under "Import & Export", allow the keystore to be imported/exported with/without a password

### Motivation

Solves issues mentioned above in the Issue section

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

SodaWithoutSparkles:
It seems that the branch custom-keystore is here but has gone stale for 2 months. Doesn't it have a PR associated?

Issam28:
Yeah this would be great. It's the only reason preventing me from using the manager as of now. Having the ability to import Keystores would be great.

GuyInCanada:
This would be great. I use Revanced-Builder (reisxd/revanced-builder) and keep a copy of revanced.keystore on OneDrive.

SodaWithoutSparkles:
https://www.reddit.com/r/revancedapp/comments/105rqat/exporting_the_keystore/?utm_term=1

Tasin5541:
I had to clear the app data to fix compiling errors on 2 occasions. This feature would be a great addition

Aunali321:
Added in https://github.com/revanced/revanced-manager/commit/dca2d4fe126a6966a094d335e0f27bb62d76c5e8

PR #641, opened 2023-01-14
feat: Export keystore

cvphat:
For #563


lib/services/manager_api.dart
```diff
@@ -129,9 +129,14 @@ class ManagerAPI {
     }
   }
 
+  Future<File> getKeyStoreFile() async {
```
oSumAtrIX, line 132: When this is merged, the previous Keystore file will be lost. Migration code is required.

oSumAtrIX (changes requested):

lib/services/manager_api.dart
cvphat: I have no idea to migrate the previous keystore. This code is just created from `initialize` of `patcher_api.dart` to make sure patcher, import, export and delete is using from the same file

oSumAtrIX: If we merge this code, will the previously saved keystore remain in tact?

cvphat: Yes

oSumAtrIX: How does it guarantee that? The previous location is `/sdcard/Android/data/app.revanced.manager.flutter/files/revanced-manager.keystore` but the new method `getKeyStoreFile()` uses `getTemporaryDirectory()` or `getExternalStorageDirectory()`. Does `getExternalStorageDirectory()` point to `/sdcard/Android/data/app.revanced.manager.flutter/files/`?

cvphat: 1. `/sdcard/Android/data/app.revanced.manager.flutter/files/` only used in `deleteKeystore` method
2. the result of `getExternalStorageDirectory()` is `/storage/emulated/0/Android/data/app.revanced.manager.flutter/files` in my phone. it is same with `/sdcard/Android/data/app.revanced.manager.flutter/files`

Aunali321: https://github.com/revanced/revanced-manager/tree/custom-keystore

Aunali321:
How is this different from https://github.com/revanced/revanced-manager/tree/custom-keystore?

oSumAtrIX:
Superseded by #776.

Issue #532, opened 2022-11-24
System theme.


ghost:
### Type

Cosmetic

### Issue

The app should automatically use the system theme.

### Feature

ReVanced Manager uses a system theme.
![Screenshot_20221124-103351_Vanced Manager](https://user-images.githubusercontent.com/112704336/203711076-8a2dfeaa-bce3-4591-9ba0-0196b6bf3962.jpg)

### Motivation

I think it's a good feature.

### Additional context

No

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

bagarwa:
Not only just the Light or Dark theme; but it'd be cool if the app can use Material You theme. (i.e. Monet)

Ushie:
#532

ILoveOpenSourceAppications:
> #532

Did you mean #276 ?

Ushie:
> > #532
> 
> Did you mean #276 ?

Yes, edited thanks

validcube:
Already fixed

Issue #276, opened 2022-09-22
feat: Add a Dynamic Dark theme


lopez05656:
### Type

Cosmetic

### Manager Branch

Flutter

### Issue

ReVanced Manager's Dark theme doesn't respect my phone's dark theme settings and has to be set manually within the app.

### Feature

Replace the Dark Theme toggle with a dialog that has 3 options:
Light Theme
Dark Theme
System Theme (Should be the default option)

### Motivation

If this feature is considered, it'll prevent people from blinding themselves when they open the ReVanced Manager at night and it'll also respect your Phone's dark theme settings as an added bonus.

ponces:
Manager already does this by default.
The first time it is opened, it uses system theme brightness to set dark mode or not.
Which version of the Manager you are using?

lopez05656:
I'm Using 0.0.19. But what I'm saying here is the app shouldn't just respect it on the first launch. It should respect it even after the first launch.

ILoveOpenSourceApplications:
Closing this in favour of #923

PR #680, opened 2023-02-05
style: use the correct m3 theming (partially)

EvadeMaster:
# 💄 Style

> **Note**: This PR only documented style(navbar, switch).

## Style(navbar)
Follow the Material 3 specification for the navigation bar: https://m3.material.io/components/navigation-bar/specs#072e0aaf-cb7e-4ac1-9bec-356f40cce58f

Old
image
New
image

## Style(switch)
Improve accessibility when the toggle is in `dark mode` && `off state`, this does not follow the Material 3 specification due to a problem with accessibility. (insufficient contrast)

Old
image
New
image

###### Feel free to share your opinion.


ILoveOpenSourceAppications:
Hey @EvadeMaster, can you also look into #532 and #276, which both seems like a similar request and a theming request for the manager. Thanks!

PR #685, opened 2023-02-08
feat: clarify microG needs to be selected

SodaWithoutSparkles:
Many support tickets are related to "non-root install is not possible". This aims to reduce similar tickets by clarifying what patch is needed for the non-root install to work.


oSumAtrIX:
This is a change into the wrong direction. The ReVanced Manager app has nothing to do with patches.

SodaWithoutSparkles:
Yes, it has more relation to the patches. But the manager is where the average user is going to see the error message. I think it's better to tell the end user that what (most likely) went wrong and how they should fix it, instead of letting them scratching their head on "why this doesn't work". 

This "recommendation" that pops up cannot be added in the patches repo. Most people dont read the fine text anyway. From a developer's perspective, this is not ideal. But from a UX perspective, this helps a lot.

PR #686, opened 2023-02-08
feat: fix patch bundle version no longer displayed, closes #670

### LINKED ISSUE #670: bug: version of patch is no longer displayed
### Bug description

The version of patch is no longer displayed as it was before. Please see the screenshot:

IMG_20230130_211732

EvadeMaster:
### Info
Add a fixed height of 500px to the textbox (apparently that works?)

### DIFF
--- a/lib/ui/views/patches_selector/patches_selector_view.dart
+++ b/lib/ui/views/patches_selector/patches_selector_view.dart
@@ -63,7 +63,6 @@ class _PatchesSelectorViewState extends State<PatchesSelectorView> {
               ),
               actions: [
                 Container(
-                  height: 500,
                   margin: const EdgeInsets.only(top: 12, bottom: 12),
                   padding:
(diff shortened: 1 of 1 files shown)


Ushie:
Is the reason for this issue known?

Mipirakas:
A container's default behavior is to take on the width and height of its child.
Setting the width and height actually enforces these as maximum values so a height of 2px should have never worked i think?
I think a better solution is to simply remove the height property since the actual height of the container is only about 40px high.
I have tested without the height property and it works just fine as it is scaled to the Text child height and width.

Ushie:
@EvadeMaster Can your PR be adjusted to do the above?

EvadeMaster:
> @EvadeMaster May you adjust your PR to do the above?

Just updated my PR, via a mobile phone!

So I would recommend that you probably should test it.

It's actually 4am (soon-to-be 5) at the time of this discussion.

Mipirakas:
This is what I tested and worked fine.

PR #691, opened 2023-02-10
POC: use flutter's built in localization feature

2shrestha22:
### What this PR does?
 - It shows how we can use Flutter's built-in localization feature. (https://docs.flutter.dev/development/accessibility-and-localization/internationalization)
 - It shows how to add a language with multiple variant. (`es_ES`, `es_MX`).
 - It shows how to fallback to default locale (`en`). App will fallback to `en` when English localization for supported country (e.g. `en_US`, `en_GB`) is not present.
 - It shows how type safe it is to implement localization.
 - It show how it automatically lists untranslated strings. (`lib/l10n/untranslated-messages.txt`)
 
### TypeSaftey:
Flutter does all code generation to make localization type safe. We never write wrong key again.
#### Before:
```dart
 I18nText(
    'homeView.patchedSubtitle', // This is a String and it can be wrong since there is no static analysis for this.
    child: Text(homeView.widgetTitle),
)
```
#### After:
```dart
Text(context.l10n().homeView_widgetTitle) // This is a getter generated by Flutter.
```


Aunali321:
I will check this out tonight. Thanks!

oSumAtrIX:
@revanced/manager Please review this. It adds translations, not sure if this is the correct approach, though.

YarostheLaunchpadder:
This would indeed be the correct way to do localization in Flutter, referencing the docs.

Issue #545, opened 2022-11-27
bug: heavy load on api


oSumAtrIX:
### Type

Error while running the manager

### Bug description

The ReVanced Manager hits the API multiple times when not necessary. Relevant code is used in multiple places:
https://github.com/revanced/revanced-manager/blob/f1261398e927903f6c75c44f5d03fdff562bda47/lib/services/revanced_api.dart#L83-L100

### Screenshots or videos

[All of these call the API each time](https://github.com/revanced/revanced-manager/blob/0faf86c9e949f82e8fdc6e33ae561cc5cca5b42f/lib/services/manager_api.dart):
![image](https://user-images.githubusercontent.com/13122796/204113521-fc920ddc-3332-4821-8ac4-5c9c73b5d7f3.png)

### Solution

Instead of hitting the API multiple times, it is only necessary to do so once and then read the fields from the parsed model.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

rootback123:
that must be why the app doesn't open if offline, right? and I also wanted to ask if the application can work in the future without any connection, of course, except for updating. Thank you

Ushie:
> that must be why the app doesn't open if offline, right? 

Yes

Ushie:
Closing as ReVanced API is fine with burst-requests and dio_cache_interceptor is now being used to cache responses anyways

PR #774, opened 2023-04-09
fix: Reduced api hits

Foxify52:
Removed all unnecessary extra hits on the api as detailed in https://github.com/revanced/revanced-manager/issues/545

### DIFF
--- a/lib/services/revanced_api.dart
+++ b/lib/services/revanced_api.dart
@@ -12,6 +12,12 @@ import 'package:revanced_manager/models/patch.dart';
 import 'package:revanced_manager/utils/check_for_gms.dart';
 import 'package:timeago/timeago.dart';
 
+Map<String, List<dynamic>> contributors = {};
+List<dynamic> patches = [];
+String extension = '';
+String repoName = '';
+List<dynamic> tools = [];
+
 @lazySingleton
 class RevancedAPI {
   late Dio _dio = Dio();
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Thanks, has this been tested and confirmed to reduce hits?

Foxify52:
I've fixed one small issue that I've found during testing. Everything is now fully tested and will only hit the API a single time and use the saved response on any consecutive calls.

oSumAtrIX:
Did you check for a difference? Iirc the current code used a cache so it didn't hit the API anyways.

lib/services/revanced_api.dart
```diff
@@ -12,6 +12,12 @@ import 'package:revanced_manager/models/patch.dart';
 import 'package:revanced_manager/utils/check_for_gms.dart';
 import 'package:timeago/timeago.dart';
 
+Map<String, List<dynamic>> contributors = {};
```
oSumAtrIX, line 15: I am not sure if it is traditional to abuse a map in this file for caching purposes. Instead a dedicated middleware should cache the REST api calls and IIrc there is already one. The question would be why it doesn't work.

Sculas:
@alexandreteles Does the API support ETags? If so, you could use that for caching the requests via an ETag-compatible DIO middleware. Of course, this requires the server to have a cache layer for the entities it's returning (it **should** already have that!). The good thing about this is that if the data ever becomes stale, it will automatically be updated. Otherwise, a `304 Not Modified` is returned with no body, which saves on bandwidth as well.

This PR works as well, though not the cleanest option in my opinion.

oSumAtrIX:
@Sculas I agree, any ideas on https://github.com/revanced/revanced-manager/pull/774#discussion_r1161267841

Sculas:
The server already returns ETags, so literally, all that needs to be done is that this library needs to be added: [dio_cache_interceptor](https://pub.dev/packages/dio_cache_interceptor).

oSumAtrIX:
@Sculas do you happen to know why the current cache fails to work?

Sculas:
I'm not sure why the existing cache impl doesn't work. But using the one I sent should work, though.

Aunali321:
Because it does not use dio_cache_interceptor as @Sculas said. It's using another library. But it's been replaced in this commit https://github.com/revanced/revanced-manager/commit/c2021d508eb05bfd48d3b65d8ff440dd41311273 of the `improve-ux` branch which is not yet in production.

oSumAtrIX:
Another library or not, it's a cache, but it does not work as expected leading to my question

Aunali321:
The older library did not support Etags.

oSumAtrIX:
And how was it intending to cache?

Aunali321:
> And how was it intending to cache?

https://pub.dev/packages/dio_http_cache

oSumAtrIX:
I don't follow, there's a cache implementation but it does not cache, which is my question

Foxify52:
Perhaps the reason it doesn't cache is because it was never set up right to begin with.

oSumAtrIX:
In that case, this PR would be working around that issue. Instead, it should either replace the incomplete implementation, or fix it.

PR #776, opened 2023-04-11
feat: add option to import/export keystore

andrisas:
add option to import/export keystore used to sign apks


Ushie:
@oSumAtrIX @Aunali321 how should ReVanced Manager ask for the imported keystore's password? should it just be an extra textbox in the popup or should it be a followup popup?

andrisas:
I do not think a password is necessary unless you use a keystore that is not generated by revanced

Aunali321:
> I do not think a password is necessary unless you use a keystore that is not generated by revanced

What if i want to use keystore from PC which i generated using ReVanced Cli with different password?

andrisas:
added option to change the keystore password

oSumAtrIX:
@revanced/manager Should be ready for a review.

Aunali321 (approved):
Works as expected, Thanks!

Issue #753, opened 2023-03-21
bug: Monochrome Icon looking quite different from the normal icon 


ILoveOpenSourceApplications:
### Type

Cosmetic

### Bug description

The title says it and the screenshots below shows it.

ReVanced Manager icons currently:
![Screenshot_20230321-203444~2](https://user-images.githubusercontent.com/117499019/226650996-a3e5f18b-8767-4756-84d9-f24146598c59.png)

![Screenshot_20230321-203453~2](https://user-images.githubusercontent.com/117499019/226651019-e5a0adee-efe1-4fe0-a393-2b0bf050a8f5.png)

How it should be looking (Motivation):

![Screenshot_20230321-203444~3](https://user-images.githubusercontent.com/117499019/226651414-1522a9e5-929f-431c-a546-dce6db79e741.png)

![Screenshot_20230321-203527~2](https://user-images.githubusercontent.com/117499019/226651546-61caa0e8-11c6-4539-b78f-8f10fe77fffe.png)

### Steps to reproduce

Welp...

### Android version

Android 13 

### Manager version

0.0.57

### Target package name

ReVanced Manager 

### Target package version.

Future Versions

### Installation type

Non-root

### Patches selected.

Nil

### Device logs (exported using Manager settings).

```shell
Nil
```

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager and not related to patching errors or patches

ILoveOpenSourceApplications:
I guess this can be closed as completed now.

PR #789, opened 2023-04-17
fix: resized monochrome icon to match the original

YarostheLaunchpadder:
I have resized the monochrome icon to match the original foreground icon, and also changed the fill to black as it doesn't affect the appearance but reduces size slightly.

This fixes issue #753

Result:


Aunali321:
Thanks!

PR #791, opened 2023-04-18
fix: open contributor links externally

YarostheLaunchpadder:
Opening links externally allows them to open in GitHub app if the user has it installed, if not, it opens in the browser.


Aunali321:
Thanks!

PR #813, opened 2023-04-24
feat: add checksums verification to Gradle

EvadeMaster:
## 🔧 Add checksums check to Gradle

_Allows for verification for the downloaded binary using SHA-256 hash comparison. This increases security against targeted attacks by preventing a man-in-the-middle attacker from tampering with the downloaded Gradle and in rare case, attacker gets access to the Gradle servers._

- add `distributionSha256Sum` to Gradle properties

> **Note** <br>
> * The checksum is provided here - https://gradle.org/release-checksums/

-----------------

### Resource

- [Gradle documentation - **Configuring SHA-256 checksum verification**](https://docs.gradle.org/current/userguide/gradle_wrapper.html#downloading_the_sha_256_file)


EvadeMaster:
I love git

SodaWithoutSparkles:
while we are here, should we add `sha256` checksums to revanced manager's release?

Issue #818, opened 2023-04-27
feat: improve suggested app version text


Ushie:
### Type

Cosmetic

### Issue

Right now, ReVanced Manager shows a `recommended version` when selecting an application, the wording may be a bit confusing to the user as it's only a recommendation

### Feature

Switch `recommended` for `supported`, this will clarify that things are likely NOT going to work if you don't follow the supported version

### Motivation

Clarification for which version of the app should be patched and reduce the number of people confused due to their patched application not working

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

KobeW50:
I'm just confused as to why in 1.1.0 it says "Suggested" instead of "Supported" even after this was marked completed.

Suggested is a soft word.

Ushie:
> I'm just confused as to why in 1.1.0 it says "Suggested" instead of "Supported" even after this was marked completed.
> 
> Suggested is a soft word.

You can read the associated issue where discussion happened

PR #822, opened 2023-04-28
feat: improve suggested app version text

SodaWithoutSparkles:
Addressing issue raised by @Ushie in #818, clarify that things might (and likely will) not work as intended if you don't use the version stated.

Maybe using the phrase `Latest supported` would be better, but I don't want to give any possible illusion of "huh, it might still work, maybe its just not the latest" then at best wasting their own time, at worst wasting everyone else time to do tech support for him/her.

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -57,7 +57,7 @@
         "widgetSubtitle": "No application selected",
         "noAppsLabel": "No applications found",
         "currentVersion": "Current",
-        "recommendedVersion": "Recommended",
+        "recommendedVersion": "Supported",
         "anyVersion": "any"
     },
     "patchSelectorCard": {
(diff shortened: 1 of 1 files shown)


Ushie:
I agree with latest supported

SodaWithoutSparkles:
So I should use the phrase "latest supported" instead?

Ushie:
Popup sounds good, I'll create a new issue, in the meantime @SodaWithoutSparkles you can change the wording to Suggested version, and if you're willing and are able to, try adding a popup for users patching non-suggested versions

SodaWithoutSparkles:
I probably don't have enough confidence on flutter/dart commit to a public repo. @Ushie could you make another PR on the popup?

"Suggested" is noe used instead of "Latest supported"

PR #857, opened 2023-05-08
chore: merge `dev` to `main` 

Aunali321:


Aunali321:
If you go with restricting, you should account for the entire compatibility list, as the latest suggested version isnt the only supported one

Aunali321:
If you go with restricting, you should account for the entire compatibility list, as the latest suggested version isnt the only supported one

PR #859, opened 2023-05-08
fix: use correct version in update download dialog

HokoraYinphineMPP:


HokoraYinphineMPP:
While downloading from 1.0.0 to 1.1.0, message says downloading 1.0.0
image

HokoraYinphineMPP:
dart.yml was unintentionally pushed, trying to figure out github for the first time, sorry :/
trying to remove the commit now.

lib/ui/views/home/home_viewmodel.dart
```diff
@@ -39,13 +39,13 @@ class HomeViewModel extends BaseViewModel {
...
-  String _managerVersion = '';
+  String? _latestManagerVersion = '';
 
   Future<void> initialize(BuildContext context) async {
-    _managerVersion = await AboutInfo.getInfo().then(
+    _latestManagerVersion = await AboutInfo.getInfo().then(
       (value) => value.keys.contains('version') ? value['version']! : '',
     );
-    _managerVersion = await _managerAPI.getCurrentManagerVersion();
+    _latestManagerVersion = await _managerAPI.getCurrentManagerVersion();
```
shabnix, line 48: This renaming does not make sense. Because the field represents the current version of the manager.

lib/ui/views/home/home_viewmodel.dart
```diff
@@ -115,15 +115,14 @@ class HomeViewModel extends BaseViewModel {
...
-    final String? latestVersion = await _managerAPI.getLatestManagerVersion();
     String currentVersion = await _managerAPI.getCurrentManagerVersion();
 
     // add v to current version
     if (!currentVersion.startsWith('v')) {
       currentVersion = 'v$currentVersion';
     }
 
-    if (latestVersion != currentVersion) {
+    if (_latestManagerVersion != currentVersion) {
```
shabnix, line 125: You are now comparing the current version of the manager (see comment above) with the current version of the manager.

lib/ui/views/home/home_viewmodel.dart
```diff
@@ -194,7 +193,7 @@ class HomeViewModel extends BaseViewModel {
                     ),
                     const SizedBox(width: 8.0),
                     Text(
-                      'v$_managerVersion',
+                      'v$_latestManagerVersion',
```
shabnix, line 196: Get this value from _managerAPI.getLatestManagerVersion() method

lib/ui/views/home/home_viewmodel.dart
shabnix: This is already in the code base but why is this field assigned twice? The managerAPI class suggests (by name) that it does API stuff so why does it know about the current manager version?

shabnix (changes requested):

lib/ui/views/home/home_viewmodel.dart
HokoraYinphineMPP: ah, on 48, i meant to change to getLatestManagerVersion()

lib/ui/views/home/home_viewmodel.dart
HokoraYinphineMPP: what do you mean by `_managerAPI.getLatestManagerVersion()`?
there is no method by that name

shabnix: There is, it was used on line 118 but you removed it.

HokoraYinphineMPP: i removed it in favor of the `_latestManagerVersion` variable that calls the `_managerAPI.getLatestManagerVersion()` method in initialize

lib/ui/views/home/home_viewmodel.dart
```diff
@@ -39,13 +39,13 @@ class HomeViewModel extends BaseViewModel {
   bool showUpdatableApps = false;
   List<PatchedApplication> patchedInstalledApps = [];
   List<PatchedApplication> patchedUpdatableApps = [];
-  String _managerVersion = '';
+  String? _latestManagerVersion = '';
 
   Future<void> initialize(BuildContext context) async {
-    _managerVersion = await AboutInfo.getInfo().then(
+    _latestManagerVersion = await AboutInfo.getInfo().then(
```
shabnix, line 45: I think you should delete this since assigning this field twice is unnecessary, and it also assigns the current manager version, which is misleading.

lib/ui/views/home/home_viewmodel.dart
HokoraYinphineMPP: the variable in line 118 was only within the scope of that function, i didnt want to do another api request since alex will bonk me.
i deleted it and made a variable on init instead so it calls for it once instead.
the `managerVersion` variable was used only for the text previously anyways, so i went ahead and deleted that too since the text would be calling from the new variable

lib/ui/views/home/home_viewmodel.dart
HokoraYinphineMPP: oh lol i didnt made that part, since its overwritten right on the next line, ill go ahead and delete it

Ushie (changes requested):
~~What is this for?~~

left review on entire PR rather than specific line by mistake

lib/ui/views/home/home_viewmodel.dart
```diff
@@ -300,6 +296,10 @@ class HomeViewModel extends BaseViewModel {
     return _managerAPI.getLatestManagerReleaseTime();
   }
 
+  Future<String?> getLatestManagerVersion() {
```
Ushie, line 299: What is this for?

Ushie (changes requested):

lib/ui/views/home/home_viewmodel.dart
HokoraYinphineMPP: unintentional test code, thought it was marked out with // but i guess i forgor 💀 
ill remove it completely

PR #873, opened 2023-05-17
chore: remove coreLibraryDesugaring

kitadai31:
coreLibraryDesugaring is enabled, but it is actually not used because this app is minSdkVersion 26+.

Also remove WindowManager library which was added in the same commit (https://github.com/revanced/revanced-manager/commit/197770b68b0a2f93f91be746f1117699d47c01d6).
(He probably read [this page](https://pub.dev/packages/flutter_local_notifications#gradle-setup) and added it to avoid crashing.)

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -30,7 +30,6 @@ android {
     ndkVersion flutter.ndkVersion
 
     compileOptions {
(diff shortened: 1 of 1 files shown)


Aunali321:
Yop, those are not needed in newer sdks, thanks!

PR #880, opened 2023-05-19
chore(deps): bump `revanced-patcher` to `8.0.0`

Ushie:
## TODO:

- [ ] Notify and agree about bumping before merge to time merging patches

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -71,7 +71,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
     // ReVanced
-    implementation "app.revanced:revanced-patcher:7.1.1"
+    implementation "app.revanced:revanced-patcher:8.0.0"
 
     // Signing & aligning
     implementation("org.bouncycastle:bcpkix-jdk15on:1.70")


oSumAtrIX:
Keep in mind that "chore" is a non-releasing commit. You may want something like this:

https://github.com/revanced/revanced-cli/commit/69fe59f97bf8c424422fee873905fa231627c383

Ushie:
So in that case, we'd use `build(deps): bump \`revanced-patcher\` to \`8.0.0\``?

oSumAtrIX:
The commit type `build` is also not a releasing commit. If you make your `build` commit a breaking change, it will release with a major version bump.

oSumAtrIX:
Regarding the custom CI script, an example commit on its usage:

https://github.com/revanced/revanced-cli/commit/995e1712bc05835437ee99ee4a1a9cda9cfdf93d

Ushie:
oh i see, i didnt focus on the scope on your first message

Ushie:
well, do we want revanced manager to jump versions like crazy? i feel like jumping an entire major version over a custom fingerprint change is well, too much

oSumAtrIX:
A breaking change is considered a "major" version bump as specified by SemVer specifications:

image

ReVanced Manager stops supporting patches that work on v7 of ReVanced Patcher, thus a breaking change is involved.

oSumAtrIX:
v8 introduces a breaking change, ReVanced Manager will not be compatible with the newest version of patches if we merge.

kitadai31:
> Keep in mind that "chore" is a non-releasing commit. You may want something like this:

ReVanced Manager does not use semantic-release, so commit type is not a problem.

oSumAtrIX:
ReVanced Manager [uses semantic-release](https://github.com/revanced/revanced-manager/blob/main/.github/workflows/release.yml#L68). This commit needs to be released, thus above recommended steps should be taken.

kitadai31:
That workflow seems to be not used.
but Ushie probably knows more about this subject, so I won't say any more.

oSumAtrIX:
dev, then dev to main. It indeed turns out that this repo does not use semantic release yet and releases are created when pushing tags.

Ushie:
> That workflow seems to be not used. but Ushie probably knows more about this subject, so I won't say any more.

You're right, semantic releases are unused atm due to incomplete implementation (cc: PalmDevs), however we still (try to) follow SemVer compliant commits and releases, although I wouldn't say I like the idea of bumping the major version for this

oSumAtrIX:
@Ushie By SemVer conventions, bumping major shall be done whenever a breaking change is introduced. Gradle and co depend on this standard as well for their dependencies. Since patcher is introducing API changes for patches, old versions of patches won't be supported any more, thus bumping this dependency causes a breaking change for ReVanced Manager. I hope this makes sense to you, if you have any objections to this, please let me know.

Ushie:
Yeah I understand completely why that is, I just don't personally like it, I'm not opposed or objecting

Ushie:
Yeah, that's also how I view it

I look at the versioning as progress of the project, 2.0.0 would be a fundemental upgrade over 1.0.0, 1.0.0 barely just released two merges ago, there's little to no changes and 2.0.0 is already coming up, which makes me feel exactly how you stated abovew

Ushie:
however is we continue to follow SemVer, we're practically obligated to follow that versioning scheme

Patcher v8 is a dependency upgrade that breaks compatibility of a significant part of Manager, hence its a breaking change which in SemVer is a major bump

oSumAtrIX:
@Ushie  It is important to not miss-understand that a major version represents. It is not meant to display "significant" changes soley. It is meant to display changes according to SemVer. The false image of the major version is created due to being familiar with other versioning schemes, but it is important to not confuse them with SemVer semantically. The point of SemVer is to standardize a certain way how to version, while a major version bump from 1 to 2 may feel like it implies something significant, you correctly understood that SemVer intends it for other certain purposes such as breaking changes.

@LisoUseInAIKyrios I think it isn't really necessary to display a friendly version as the version itself doesn't really mean much to the end user. While a major bump may imply the wrong thing for many users, after the first or second time confusion will clear itself.

oSumAtrIX:
Apparently ReVanced Manager is not using SemVer and currently is bumped manually

0xrxL:
> Need to pick a day/time to do this.
> 
> How about today?

Yeah...everything it's blocked now (and who knows how long it will be).

Ushie:
This is already resolved in revanced-patcher in dev, it'll be ready to release once that's merged to main

We're ready to merge this whenever anyways, even if revanced-patcher dev wasn't merged before then, iirc revanced-patches weren't ready to merge

oSumAtrIX:
Imo this is ready as is because it is a merge to `dev`, not `main`.

PR #889, opened 2023-05-24
feat: add fastlane for F-Droid

Aunali321:


android/fastlane/metadata/android/images/icon.png
EvadeMaster, line 1: 🤔 This logo is not up-to-date with the current ReVanced Manager's icon.

Issue #894, opened 2023-05-24
bug: update download screen shows two "v"s


HokoraYinphineMPP:
### Type

Error while running the manager

### Bug description

Download Update screen shows `vv1.3.0` instead of `v1.3.0`

### Steps to reproduce

1) Update Manager
2) Bug visible

### Android version

11

### Manager version

1.2.0

### Target package name

N/A

### Target package version.

N/A

### Installation type

Non-root

### Patches selected.

N/A

### Device logs (exported using Manager settings).

```shell
N/A
```

### Screenshots or videos

![image](https://github.com/revanced/revanced-manager/assets/54823358/be62a7cf-71a1-4664-abe5-d685c0dbeae3)

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager and not related to patching errors or patches

EvadeMaster:
🎉 This issue have been fixed by https://github.com/revanced/revanced-manager/pull/895, please wait for the next version of ReVanced Manager.

You are welcome to reopen this issue if the problem persists.

PR #895, opened 2023-05-24
fix: remove redundant "v" in the downloader screen

MrAngelos6:
Fix: #894
There is already a "v" before the version number thanks to the workflow. So now it's useless to write it in the string.

### DIFF
--- a/lib/ui/views/home/home_viewmodel.dart
+++ b/lib/ui/views/home/home_viewmodel.dart
@@ -190,7 +190,7 @@ class HomeViewModel extends BaseViewModel {
                     ),
                     const SizedBox(width: 8.0),
                     Text(
-                      'v$_latestManagerVersion',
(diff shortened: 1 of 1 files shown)


HokoraYinphineMPP:
Pushing into main?

Ushie (changes requested):
Does latestManagerVersion always include `v` in the string?

EvadeMaster:
> Does latestManagerVersion always include `v` in the string?

Yes indeed, ReVanced Manager get `latestManagerVersion` from [ReVanced API Tools](https://releases.revanced.app/tools), the response data from ReVanced API already included "v", so there's no reason to include it.

https://github.com/revanced/revanced-manager/blob/8af62b917ce284b907da048b30f28bd94087f370/lib/services/revanced_api.dart#L100-L119

Ushie (approved):

PR #915, opened 2023-06-02
refactor: clarify architecture in about section

EvadeMaster:
This commit change `Arch` to `Supported Arch`

### DIFF
--- a/lib/ui/widgets/settingsView/about_widget.dart
+++ b/lib/ui/widgets/settingsView/about_widget.dart
@@ -29,7 +29,7 @@ class _AboutWidgetState extends State<AboutWidget> {
                         text: 'Version: ${snapshot.data!['version']}\n'
                             'Model: ${snapshot.data!['model']}\n'
                             'Android Version: ${snapshot.data!['androidVersion']}\n'
-                            'Arch: ${snapshot.data!['arch']}\n',
+                            'Supported Arch: ${snapshot.data!['arch']}\n',
                       ),
                     );
                     ScaffoldMessenger.of(context).showSnackBar(
@@ -84,7 +84,7 @@ class _AboutWidgetState extends State<AboutWidget> {
                         ),
                       ),
                       Text(
-                        'Arch: ${snapshot.data!['arch']}',
+                        'Supported Arch: ${snapshot.data!['arch']}',
                         style: const TextStyle(
                           fontSize: 13,
                           fontWeight: FontWeight.w300,


Ushie:
Could the text acknowledge the existence of multiple architectures? so if archs.length > 1, "Supported archs" or so

EvadeMaster:
> Could the text acknowledge the existence of multiple architectures? so if archs.length > 1, "Supported archs" or so

Done! 1ca3a1bbe4f8727f0bc96688cdc916bfc65ffc2d

lib/ui/widgets/settingsView/about_widget.dart
```diff
@@ -84,7 +84,9 @@ class _AboutWidgetState extends State<AboutWidget> {
                         ),
                       ),
                       Text(
-                        'Arch: ${snapshot.data!['arch']}',
+                        snapshot.data!['supportedArch'].length > 1
+                            ? 'Supported Archs: ${snapshot.data!['supportedArch'].join(", ")}'
+                            : 'Supported Arch: ${snapshot.data!['supportedArch'].join(", ")}',
```
Ushie, line 89: What's the .join() for? you already check if it's a single architecture

lib/ui/widgets/settingsView/about_widget.dart
```diff
@@ -29,7 +29,7 @@ class _AboutWidgetState extends State<AboutWidget> {
                         text: 'Version: ${snapshot.data!['version']}\n'
                             'Model: ${snapshot.data!['model']}\n'
                             'Android Version: ${snapshot.data!['androidVersion']}\n'
-                            'Arch: ${snapshot.data!['arch']}\n',
+                            'Supported Arch: ${snapshot.data!['supportedArch'].join(", ")}\n',
```
Ushie, line 32: Where is this used? it might also need the plural support

Ushie (changes requested):

lib/ui/widgets/settingsView/about_widget.dart
EvadeMaster: formatting, it's there for displaying `Supported Archs: arm-v8a, armabi-v7a` instead of `Supported Archs: [arm-v8a, armabi-v7a]`.

EvadeMaster: formatting, it's there for displaying `Supported Archs: arm-v8a, armabi-v7a` instead of `Supported Archs: [arm-v8a, armabi-v7a]`.

lib/ui/widgets/settingsView/about_widget.dart
EvadeMaster: When you long-press on the About list, it will copy the value to the clipboard.

As for plural support, will do!

EvadeMaster: When you long-press on the About list, it will copy the value to the clipboard.

As for plural support, will do!

EvadeMaster: Resolved in 69f325337ada87a616c389e7e07e289939e30bdb

lib/ui/widgets/settingsView/about_widget.dart
Ushie: Do you need `.join()` for non-plural? because it would be `Supported arch: arm-v8a` for example, the `.join()` will never be required because its never above 1

EvadeMaster: I realised that. It shouldn't affect anything but would be nice to fix. f94879c17195c8c0a7f8a2c291030f9be2078ca9

Ushie (approved):

PR #919, opened 2023-06-06
feat: update icons for the new logo

### LINKED ISSUE #918: fix: update branding
## About

ReVanced brandings have updated

## Issue

The current branding is outdated

## Solution

Update the current branding to the new one according to the following:
https://github.com/revanced/revanced-branding

YarostheLaunchpadder:
fixed #918

The icon is adaptive according to the docs: https://developer.android.com/develop/ui/views/launch/icon_design_adaptive


Ushie:
Could you use the no-ring variant? There are more shapes than circle to count for

YarostheLaunchpadder:
Sure. Will do so. But only for the launcher icon? Or for the notification icon as well? @Ushie

EvadeMaster (approved):
:arrow_up_small: Look good to me, 

would be great if we switch back to the no-ring variant like Ushie suggest, that way it should make it distinguish from YouTube ReVanced which uses ring variant.

YarostheLaunchpadder:
Perfect. I will make the new SVG variant without the ring, so it will look on all devices. 
But should I leave the legacy icon (.png) with the ring as it can't be another shape?

Aunali321 (approved):

YarostheLaunchpadder:
@Ushie Yes, monochrome should be updated and work.

PR #921, opened 2023-06-06
fix: Return fetched patch version if non default patch repo is used

1fexd:
Looks like the return statement has been forgotten when a non-default repo is used, resulting in an NPE


EvadeMaster:
Superseded by #922

Issue #876, opened 2023-05-18
feat: improve UX in regards of displaying the suggested version in app list


oSumAtrIX:
### Type

Error while running the manager

### Bug description

ReVanced Manager shows the suggested version in the app list. This string can be interpreted as the app's version currently installed on the device.

### Manager version

1.1.0

### Installation type

Non-root

### Solution

Redesign the app card content. For example, the current app version may be shown next to the app name. The package name should be hidden as it has no actual use for the end user, and instead, the app card should clarify that the suggested version is a suggested version.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager and not related to patching errors or patches

Ushie:
This issue is speaking about here 
![Screenshot_20230518_214306](https://github.com/revanced/revanced-manager/assets/79272171/4cc945c5-32f3-441a-b6bc-b64ff1fedc4c)

By the way, 1.0.0+ does still have what you're portraying 
![Screenshot_20230518_214255](https://github.com/revanced/revanced-manager/assets/79272171/c3dad671-8b0d-4bad-a00d-87edb7793747)

theofficialgman:
where is the feature? there is only one version displayed. what is that version? is that suggested or current?
that is what this issue is about. the UI does NOT specify what the version is depicting.

Ushie:
yes... that's what the issue is about

theofficialgman:
> By the way, 1.0.0+ does still have what you're portraying ![Screenshot_20230518_214255](https://user-images.githubusercontent.com/79272171/239351822-c3dad671-8b0d-4bad-a00d-87edb7793747.jpg)

you can't get to this page without first selecting an apk from storage. negating the entire purpose of checking revanced manager for current vs recommended version.

this was a feature without selecting any apk from storage on pre-1.0.0 and is no longer the case on 1.0.0+, meaning this is a regression, a feature that was removed

Ushie:
[...]

PR #931, opened 2023-06-10
feat: Show app name and icon for the apps that are not installed on the device.

TheAabedKhan:


Aunali321:
Please do not merge until i approve it. This needs to be discussed with team. Because the next api is release is going to have this functionality.

Ushie:
Please separate your commit regarding #876 into a new PR, it could be merged before this

lib/services/google_play_api.dart
```diff
@@ -0,0 +1,26 @@
+import 'dart:convert';
+import 'package:http/http.dart' as http;
+
+Future<Map<String, dynamic>> getPackageInfo(String pkgName) async {
+  String result = '{}';
+  if (pkgName == 'com.vanced.android.youtube') {
```
oSumAtrIX, line 6: This breaks design. ReVanced Manager should not make an exception for an app that has nothing to do with ReVanced Manager

lib/services/google_play_api.dart
```diff
@@ -0,0 +1,26 @@
...
+    result =
+        '{"name":"YouTube Vanced","image":"https://downloadr2.apkmirror.com/wp-content/uploads/2019/02/youtube-vanced.jpg", "pkgName":"com.vanced.android.youtube"}';
+  } else {
```
oSumAtrIX, line 9: Before hitting any additional IP a warning has to be displayed to the user. This also goes for the sources or API links. Additionally this implementation forces using Google Play Store. What if I don't want to use Google Play Store? This should be modularized enough, so that the provider for app names and icons can be switched out through an interface.

oSumAtrIX (changes requested):

lib/services/google_play_api.dart
TheAabedKhan: No worrries, I will remove the exception. I can also try to find the name and icon on the APKMirror if Google Play Store doesn't have the app. So, two sources.

lib/services/google_play_api.dart
oSumAtrIX: If you design an interface, you can implement both or one of them. The important part is that the interface exists so if necessary the provider can be switched out.

TheAabedKhan: OK. So, can I close this pull and create a new one after adding the requested changes?

Aunali321: > OK. So, can I close this pull and create a new one after adding the requested changes?

It will be alot of effort. I do not think it is worth spending time on. Current manager will replaced by Compose rewrite where this is already in the works.

PR #938, opened 2023-06-12
feat(updater): download successful dialog

TheAabedKhan:
When a user knowingly or unknowingly cancels or closes the installer window, there must be a dialog box which allows the user to launch the installer window again.
revanced_ui

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -24,6 +24,7 @@
         "WIP": "Work in progress...",
         "noInstallations": "No patched applications installed",
         "installed": "Installed",
+        "installUpdate": "Continue to install the update?",
         "updateDialogTitle": "Update Manager",
         "updateChangelogTitle": "Changelog",
         "notificationTitle": "Update downloaded",
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -29,6 +29,7 @@
         "notificationTitle": "Update downloaded",
         "notificationText": "Tap to install the update",
         "downloadingMessage": "Downloading update...",
+        "downloadedMessage": "Update downloaded!",
         "installingMessage": "Installing update...",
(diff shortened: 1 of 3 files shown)


ILoveOpenSourceApplications:
If this UI/UX change is the same as the screenshot, then I would like to suggest some changes. "Update downloaded!" text being shown in place of the progress bar, could be placed in the position of "downloading update...". Once they download completes, the text could change into "Download completed"/"Update downloaded" and then the user can update by tapping the update button.

Ushie:
After you resolve the issue raised in the comment above, this can be merged

TheAabedKhan:
New UI after fixes

revanced_updater_new

Ushie:
Maybe the Update button should be renamed to Install instead?

Ushie:
I think the progress bar doesn't need to be there anymore, what about text instead? "Do you want to install this update?" or something along the lines of that

ILoveOpenSourceApplications:
> Maybe the Update button should be renamed to Install instead?

I second this. Why press update twice, when downloading and installing.

> I think the progress bar doesn't need to be there anymore, what about text instead? "Do you want to install this update?" or something along the lines of that

Would make more sense with this as well.

TheAabedKhan:
> "Do you want to install this update?" or something along the lines of that

The package installer will automatically open after the update is downloaded so a user can't see what's written on the dialog box. If the user mistakenly cancels the installer, then the user can see the text that says, "Continue to install the update?" which makes more sense.

Ushie:
That text sounds good

Ushie:
Have you been able to make time to update the text?

TheAabedKhan:
> Have you been able to make time to update the text?

What text to what?

Ushie:
Oh my mistake, completely missed that commit

PR #941, opened 2023-06-12
feat(patches-selector): Improve disabled card UI

ponces:
This includes hiding the "Warning" label on the disabled card as interacting with it only makes sense on an enabled card.

Before vs After:
Screenshot_20230612-221638

### DIFF
--- a/lib/ui/widgets/patchesSelectorView/patch_item.dart
+++ b/lib/ui/widgets/patchesSelectorView/patch_item.dart
@@ -43,135 +43,135 @@ class _PatchItemState extends State<PatchItem> {
   Widget build(BuildContext context) {
     return Padding(
       padding: const EdgeInsets.symmetric(vertical: 4.0),
-      child: CustomCard(
-        backgroundColor: widget.isUnsupported &&
+      child: Opacity(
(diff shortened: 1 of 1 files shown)


ponces:
They will try to interact with it and the popup will clearly say it why.
Having an interactable subcomponent inside a non-interactable component does not make sense at all...

oSumAtrIX:
The UI will not motivate the user to interact with it. The user will simply take for granted that it is not available, but not know why.

ponces:
> The UI will not motivate the user to interact with it. The user will take for granted that it is unavailable but not know why. Before, the user was motivated by the interactive chip.

That's not true. The user is always tempted to click it anyway just to see what happens.
In fact, I only knew about the toast because I tried to interact with the greyed out card.
Yet again:
> Having an interactable subcomponent inside a non-interactable component does not make sense at all...

Aunali321 (approved):
Reasoning makes sense. Looks good to me.

EvadeMaster (approved):
:tada: Looks good to me!

ponces:
Is anything missing?

PR #943, opened 2023-06-14
feat(app-selector): Improve app item UI to avoid overflow issues

TheAabedKhan:
Before vs After

revanced_app_selector_ui

### DIFF
--- a/lib/ui/widgets/appSelectorView/installed_app_item.dart
+++ b/lib/ui/widgets/appSelectorView/installed_app_item.dart
@@ -1,5 +1,4 @@
 import 'dart:typed_data';
-
 import 'package:flutter/material.dart';
 import 'package:flutter_i18n/flutter_i18n.dart';
 import 'package:revanced_manager/ui/widgets/shared/custom_card.dart';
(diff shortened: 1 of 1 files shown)


lib/ui/widgets/appSelectorView/installed_app_item.dart
```diff
@@ -52,26 +51,25 @@ class _InstalledAppItemState extends State<InstalledAppItem> {
...
-                        ),
-                      ),
-                      const SizedBox(width: 4),
-                      I18nText(
-                        widget.installedVersion.isEmpty
-                            ? 'All versions'
-                            : 'v${widget.installedVersion}',
-                      ),
-                    ],
+                  Text(
```
oSumAtrIX, line 54: Why was the Row removed? Semantically each line of text should be a row item

TheAabedKhan: That's because the installed version is shown right above the suggested version and no Row is needed.

oSumAtrIX: What if the device is widescreen? Won't the items stop to be wrapped and showed next to each other? Using Row and Columns usually also are also done for semantical reasons. If you know that you want to show a line of text below another, you use a row of those items

TheAabedKhan: I could have used Wrap as well, but I guess the Installed version and the Suggested version should be next to each other.

PR #944, opened 2023-06-14
fix(app-selector): Remove direct use of strings 

TheAabedKhan:


lib/ui/widgets/appSelectorView/not_installed_app_item.dart
```diff
@@ -54,12 +55,28 @@ class _NotInstalledAppItem extends State<NotInstalledAppItem> {
...
                     children: [
-                      Text(
-                        'Suggested: ${widget.suggestedVersion.isEmpty ? 'All versions' : 'v${widget.suggestedVersion}'}',
+                      I18nText(
+                        FlutterI18n.translate(
+                          context,
+                          'suggested',
+                          translationParams: {
+                            'version': widget.suggestedVersion.isEmpty
+                                ? 'All versions'
```
Ushie, line 76: Did you happen to miss using the key here instead?

Ushie (changes requested):

lib/ui/widgets/appSelectorView/not_installed_app_item.dart
TheAabedKhan: Which key is missing?

Ushie: You could create an i18n key for the string "All versions" here

TheAabedKhan: Can you check the new commits?

assets/i18n/en_US.json
```diff
@@ -5,7 +5,7 @@
     "enabledLabel": "Enabled",
     "disabledLabel": "Disabled",
     "installed":"Installed: {version}",
-    "suggested":"Suggested: {version}",
+    "suggested":"Suggested: ",
```
Ushie, line 8: Did you remove `{version}` by mistake? or intentionally

Ushie (changes requested):

assets/i18n/en_US.json
TheAabedKhan: Intentionally, as it is no longer required anywhere.

Ushie (changes requested):

assets/i18n/en_US.json
```diff
@@ -58,9 +58,11 @@
         "widgetTitleSelected": "Selected application",
         "widgetSubtitle": "No application selected",
         "noAppsLabel": "No applications found",
+        "notInstalled":"App not installed",
         "currentVersion": "Current",
         "suggestedVersion": "Suggested",
-        "anyVersion": "any"
+        "anyVersion": "any",
+        "allVersions": "All versions"
```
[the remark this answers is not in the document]
TheAabedKhan: The `anyVersion` is used in the Patcher page after you select any application for patching. I also thought if I could replace it with `allVersions`.  Let me know what you think.

PR #1000, opened 2023-07-08
Adding necessary perms

HinaHyugaHime:


TheAabedKhan:
No need to add those permissions in the manifest manually, as the permissions are automatically added by `permission_handler` package.

oSumAtrIX:
What are these permissions each used for?

Hinatahyugahime:
> What are these permissions each used for anyways?

They are used to allow installing packages and access to files so dont have to manually allow it for after install app

TheAabedKhan:
> > What are these permissions each used for anyways?
> 
> They are used to allow installing packages and access to files so dont have to manually allow it for after install app

Users are asked for permissions no matter what the manifest says.

Hinatahyugahime:
> What are these permissions each used for anyways?

I added the two perms it asks for before patching and after patching to the manifest to access files (to get the apk) and install packages, (so you dont have to tick that manually)

TheAabedKhan:
> > What are these permissions each used for anyways?
> 
> I added the two perms it asks for before patching and after patching to the manifest to access files (to get the apk) and install packages, (so you dont have to tick that manually)

image
Maybe you are new to android development

Hinatahyugahime:
> > > What are these permissions each used for anyways?
> > 
> > 
> > I added the two perms it asks for before patching and after patching to the manifest to access files (to get the apk) and install packages, (so you dont have to tick that manually)
> 
> image Maybe you are new to android development

I mean, you do need those two permissions for the app to function correctly, being the purpose of the app is to patch a apk and before it only had internet, not to access files and install packages, so wouldnt it be necessary?

TheAabedKhan:
> I mean, you do need those two permissions for the app to function correctly, being the purpose of the app is to patch a apk and before it only had internet, not to access files and install packages, so wouldnt it be necessary?

The app asks for all the permissions it needs whenever it needs. You can see there is a `permission_handler` package in the `pubspec.yaml` file.

Hinatahyugahime:
> > I mean, you do need those two permissions for the app to function correctly, being the purpose of the app is to patch a apk and before it only had internet, not to access files and install packages, so wouldnt it be necessary?
> 
> The app asks for all the permissions it needs whenever it needs. You can see there is a `permission_handler` package in the `pubspec.yaml` file.

as you quoted though, the permissions to operate properly should be in the manifest file, regardless of if it is handled by the handler, because it flags them as required.

TheAabedKhan:
> as you quoted though, the permissions to operate properly should be in the manifest file, regardless of if it is handled by the handler, because it flags them as required.

Adding the permissions to manifest will make the app ask for permissions after launch when the permissions may not be required. However, if the permissions are handled by the `permission_handler` package, then the app can ask for the permissions only when it is required.

Hinatahyugahime:
> > as you quoted though, the permissions to operate properly should be in the manifest file, regardless of if it is handled by the handler, because it flags them as required.
> 
> Adding the permissions to manifest will make the app ask for permissions after launch when the permissions may not be required. However, if the permissions are handled by the `permission_handler` package, then the app can ask for the permissions only when it is required.

to use the app to begin with, you need to have the access files, and by adding both you are shortening the need to do both seperately vs doing them both at once and then can get on with your day patching whatever needs to be patched, I personally find it very annoying being asked before finding apk and after its patched and potentially losing the patched apk because if I misclick I have to patch again cause it thinks I had already installed when I only enabled the ability to install.

taku-nm:
>I have to patch again cause it thinks I had already installed when I only enabled the ability to install.

While this is somewhat of an UX issue, it would be worse to ask for permissions for which it is not apparent to the user, why the app would need such.
Without the direct justification of clicking the install button and the app then asking for permission to install, you will have many users that don't allow it in the beginning and then have to figure out how to change it or (more likely) will head to support and ask why it said install failed.
With it asking for permission when it needs to, it defines that clear justification for people unaware of how the app is supposed to work.

Also remember to always make PRs to merge to :dev ; not :main

LordeSenka:
> > I have to patch again cause it thinks I had already installed when I only enabled the ability to install.
> 
> While this is somewhat of an UX issue, it would be worse to ask for permissions for which it is not apparent to the user, why the app would need such. Without the direct justification of clicking the install button and the app then asking for permission to install, you will have many users that don't allow it in the beginning and then have to figure out how to change it or (more likely) will head to support and ask why it said install failed. With it asking for permission when it needs to, it defines that clear justification for people unaware of how the app is supposed to work.
> 
> Also remember to always make PRs to merge to :dev ; not :main

fixed

LordeSenka:
so should I just remove the install one? even though that one is a necessary in my opinion.

taku-nm:
> so should I just remove the install one? even though that one is a necessary in my opinion.

I just use the install permission as an example but the same applies for storage access. We already have people sometimes coming into support, asking why the app needs access to storage, even tho its already fairly obviously linked.
Asking for these permissions before the user has the possibility to understand why the app needs a certain permission its just going to create more confusion.

LordeSenka:
> > so should I just remove the install one? even though that one is a necessary in my opinion.
> 
> I just use the install permission as an example but the same applies for storage access. We already have people sometimes coming into support, asking why the app needs access to storage, even tho its already fairly obviously linked. Asking for these permissions before the user has the possibility to understand why the app needs a certain permission its just going to create more confusion.

honestly, if they are stupid enough to question that, they probably would be asking multiple questions to begin with, but for people who actually know what they are doing, its better, just idiots who dont know how to use the app and ask 20 questions would be lost

taku-nm:
> > > so should I just remove the install one? even though that one is a necessary in my opinion.
> > 
> > 
> > I just use the install permission as an example but the same applies for storage access. We already have people sometimes coming into support, asking why the app needs access to storage, even tho its already fairly obviously linked. Asking for these permissions before the user has the possibility to understand why the app needs a certain permission its just going to create more confusion.
> 
> honestly, if they are stupid enough to question that, they probably would be asking multiple questions to begin with, but for people who actually know what they are doing, its better, just idiots who dont know how to use the app and ask 20 questions would be lost

It would only matter for those that have never used the app before = new users.
new users are infinitely more likely to not be aware of how the app functions and why it would need permissions to access files and install apps.
People that have used the manager before already have the permissions set.

PR #1003, opened 2023-07-08
fix: app crash after custom source

### LINKED ISSUE #982: bug: ReVanced Manager hangs with black screen since v1.3.7 when using a custom source
### Type

Error while running the manager

### Bug description

Phone: Samsung Glaxy S22 Ultra (Android 13, latest patch May 2023)

When using a custom source, the latest working version of ReVanced Manager is 1.3.6. It asks for all files permision on first run and then starts normally. After updating to 1.3.7 or 1.3.8 (while having a custom source in the settings) the screen stays black when running the app. I tried clean install, phone reboot and setting all prermissions prior to running the app, nothing works.

Side note: Downgrading to 1.3.6 doesn't work while a newer version is installed (invalid package), so I have to manually uninstall 1.3.8, only then I can install 1.3.6 which is the latest working version with a custom source.

Currently, the only "fix" is to downgrade to 1.3.6 as described, go to the settings and remove the custom source (set to default), then update to 1.3.8 because apparently the manager settings persist even after an uninstall. This is a separate issue which makes troubleshooting very difficult if the app breaks because of the settings and there is no obvious way to reset them.

### Steps to reproduce

Install ReVanced Manager 1.3.7 or 1.3.8, set a custom source and run.

### Android version

Android 13 (May 2023)

### Manager version

1.3.7 and 1.3.8

### Target package name

ReVanced Manager

### Target package version.

1.3.7 and 1.3.8

### Installation type

Non-root

### Patches selected.

-

### Device logs (exported using Manager settings).

```shell
-
```

### Installer logs (exported using Installer menu option) [unneeded if issue is not during patching].

_No response_

### Screenshots or videos

_No response_

### Solution

_No response_

### Additional context

_No response_

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager and not related to patching errors or patches

TheAabedKhan:
This PR fixes #982 by using separate function to check for manager update so that the custom source don't use the same function because the custom source may not support the same logic as the default one.

### DIFF
--- a/lib/services/github_api.dart
+++ b/lib/services/github_api.dart
@@ -1,6 +1,5 @@
 import 'dart:convert';
 import 'dart:io';
-
 import 'package:collection/collection.dart';
 import 'package:dio/dio.dart';
 import 'package:dio_cache_interceptor/dio_cache_interceptor.dart';
@@ -9,7 +8,6 @@ import 'package:flutter_cache_manager/flutter_cache_manager.dart';
 import 'package:injectable/injectable.dart';
 import 'package:revanced_manager/models/patch.dart';
 import 'package:revanced_manager/services/manager_api.dart';
-
 
 @lazySingleton
 class GithubAPI {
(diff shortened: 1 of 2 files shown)


Aunali321 (approved):

kenethc:
If using inotia00 source, inotia00 has released his own fork of ReVanced Manager called RVX Manager v1.3.8.

I used the RVX Manager and it worked perfectly. Was able to patch YouTube 18.25.40

PR #1005, opened 2023-07-09
refactor: Acknowledgements in bug issue form

KobeW50:
Clarified acknowledgements in bug issue form

Prompted by [this discussion](https://github.com/revanced/revanced-patcher/issues/205#issuecomment-1627727173)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-issue.yml
+++ b/.github/ISSUE_TEMPLATE/bug-issue.yml
@@ -118,5 +118,5 @@ body:
           required: true
         - label: The issue is related solely to the ReVanced Manager and not related to patching errors not caused by issues of the ReVanced Manager.
           required: true
-          - label: The issue is related solely to the ReVanced Manager and not related to issues with patches.
+        - label: The issue is related solely to the ReVanced Manager and not related to issues with patches.
           required: true
--- a/.github/ISSUE_TEMPLATE/bug-issue.yml
+++ b/.github/ISSUE_TEMPLATE/bug-issue.yml
@@ -116,5 +116,7 @@ body:
           required: true
         - label: I filled out all of the requested information in this issue properly.
           required: true
-        - label: The issue is related solely to the ReVanced Manager and not related to patching errors or patches
+        - label: The issue is related solely to the ReVanced Manager and not related to patching errors not caused by issues of the ReVanced Manager.
+          required: true
+          - label: The issue is related solely to the ReVanced Manager and not related to issues with patches.
           required: true
--- a/.github/ISSUE_TEMPLATE/bug-issue.yml
+++ b/.github/ISSUE_TEMPLATE/bug-issue.yml
@@ -29,7 +29,7 @@ body:
   - type: textarea
     attributes:
       label: Steps to reproduce
-      description: Add the steps to reproduce this bug including your environment.
+      description: Add the steps to reproduce this bug, including your environment.
       placeholder: Step 1. Download some files. Step 2. ...
     validations:
       required: true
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug-issue.yml
```diff
@@ -116,5 +116,7 @@ body:
           required: true
         - label: I filled out all of the requested information in this issue properly.
           required: true
-        - label: The issue is related solely to the ReVanced Manager and not related to patching errors or patches
+        - label: The issue is related solely to the ReVanced Manager and not related to patching errors not caused by issues of the ReVanced Manager.
```
Ushie, line 119: This implies the user is aware of whether or not errors are caused by ReVanced Manager or ReVanced Patcher

KobeW50: I don't think there could be a sure way for the user to know what their issue is related to unless they understand the context and meaning of the patching error.

But before the user was also expected to know what their issue related to. But before it didn't say this explicitly, which led me, and likely other users, whose issues were indeed related to issues with the Manager to report in the patcher and patches repo instead.

oSumAtrIX: There's not much we can do about this. The end user is forced to differentiate between those errors. The suggested sentence is also very confusing, I don't understand what it means

KobeW50: Is this better?

"The issue is not related to patching errors not caused by a ReVanced Manager issue.

The issue is related solely to the ReVanced Manager and not related to issues with patches."

I can't think of a way to avoid having a double negative in the first sentence.

oSumAtrIX: > The issue is related solely to the ReVanced Manager and not related to issues with patches.

How about:


The issue is solely related to ReVanced Manager and not patching errors or patches

KobeW50: That's what it was lol. But the issue is that it implies that all patching errors do not belong in this repo, as discussed.

oSumAtrIX: Correct, all patching errors do not belong in this repo. Only errors caused by ReVanced Manager.

KobeW50: But it doesn't suggest that there is a difference between a "patching error" and an error (while patching) caused by the manager.

Without explicitly saying that a differentiation can be made the user will never know. It's best to inform the user, even at the cost of a slightly difficult-to-read sentence, rather than scare them away due to a lack of understanding.

oSumAtrIX: > But it doesn't suggest that there is a difference between a "patching error" and an error (while patching) caused by the manager.

I can not see this in your suggestion either though.

KobeW50: Sorry. It isn't explicit, but it is heavily implied.

oSumAtrIX: I have a more challenging time understanding the suggestion. The implication is present in the current text as well.

KobeW50: It is not. 

"solely to the ReVanced Manager and not related to patching errors or patches." implies that the issue must be something not related AT ALL to patching errors.

Ushie: There's no right way to word this, there's really absolutely no way for the user to figure out at which part of the chain the error occured

KobeW50: Depending on context (the error message, patches applied, application) and knowledge of the user, they may be able to discern what is happening. For example, if the same error is given for all applications the issue is likely with the Manager.

oSumAtrIX: > It says nothing about an exeption for patching errors that are caused by the manager.

It does, implicitly by saying "solely to ReVanced Manager", that implicitly includes errors caused by ReVanced Manager.

oSumAtrIX:
Closing as no real solution can be found atm.

.github/ISSUE_TEMPLATE/bug-issue.yml
KobeW50: Last attempt

It is difficult to deduct that.

What about, "The issue is related solely to the ReVanced Manager and not related to other patching errors or patches."

oSumAtrIX: There is no difference or any additional benefit to the current sentence.

oSumAtrIX (approved):

PR #1018, opened 2023-07-13
fix: Prevent `Unsupported operation` exception

ponces:
This should fix:
```
Unhandled Exception:
Unsupported operation: Cannot add to a fixed-length list
```


EvadeMaster (approved):
⬆️ Looks fine to me!

PR #1052, opened 2023-07-26
fix: `original.apk` not found despite existing

TheAabedKhan:
The manager has no permissions to access the contents of `/data/adb` folder. So, it is not detecting the `revanced` folder and its contents. When the `original.apk` file actually exists when checked on the Flutter side, the path is invoked to the Android side which is not accessible. To solve the issue, we can directly use the `apkFilePath` as the `originalFilePath` without using the root method.

### DIFF
--- a/lib/services/patcher_api.dart
+++ b/lib/services/patcher_api.dart
@@ -144,20 +144,6 @@ class PatcherAPI {
     );
   }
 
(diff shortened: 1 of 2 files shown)


Aunali321 (approved):
Make sure to test before merge

oSumAtrIX (approved):
lgtm

Issue #886, opened 2023-05-23
Move Export button next to the Install button


KobeW50:
### Type

Cosmetic

### Issue

The export button is kinda hidden, and some people won't realize that it is an option.

### Feature

Place the Export/Save APK button beside the Install button at the patch end-screen.

### Motivation

Improved UX

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

tejasbadone:
Hey, I would like to work on this, can you please assign it to me?

Ushie:
How would you achieve this? there are already 2 buttons (install and install as root) which is ideally the maximum amount of buttons

oSumAtrIX:
@Ushie A popup can show the installation variants. The popup can be shown, when clicking on"Install". This would free the existing second button

tejasbadone:
will add a new button in the row inside the Visibility widget and will use sizedbox for the space. and will remove "Export APK" from the CustomPopupMenu. the rooted devices will see three buttons and non-root will see two


Visibility(
                              visible: !model.isInstalled,
                              child: const SizedBox(
                                width: 16,
                              ),
                            ),

Visibility(
                              visible: !model.isInstalled,
                              child: CustomMaterialButton(
                                label: I18nText('installerView.exportApkMenuOption'),
                                isExpanded: true,
                                onPressed: () => model.exportResult(),
                              ),
                            ),

Ushie:
> the rooted devices will see three buttons and non-root will see two

I'm not sure if we're meant to go with as many as 3 buttons in a single row, let's see what Palm says

PalmDevs:
> How would you achieve this? there are already 2 buttons (install and install as root) which is ideally the maximum amount of buttons

@Ushie The new Installer screen design in the Figma somewhat already fixes that. Check it out.

PalmDevs:
> > @Ushie A popup can show the installation variants. The popup can be shown, when clicking on"Install". This would free the existing second button
> 
> @PalmDevs Can you review this design proposal? 

Looks good to me. However, I'm adding on to the proposal that we do not need to show the dialog to non-root users because they can't select any other installation types anyway. I'm thinking of a dialog with list items that have radio buttons.

oSumAtrIX:
[...]

PR #1062, opened 2023-07-29
feat(installer): redesign utility options

validcube:
## ✨ Redesign utility options

This replace the 3 dots menu with buttons, improve visibility

> Save APK
> Share Log

This also removed the share option from the installer

image


Ushie:
What'd you do to share?

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -264,17 +254,14 @@ class InstallerViewModel extends BaseViewModel {
...
     switch (value) {
       case 0:
-        shareResult();
+        exportResult();
         break;
       case 1:
         exportResult();
         break;
-      case 2:
-        shareLog();
```
validcube, line 276: bruh

Ushie:
Use contained tonal icon buttons https://m3.material.io/components/icon-buttons/overview

Ushie:
Also @PalmDevs could you take a look at this?

PalmDevs:
I'd personally suggest using the already-existing [Bottom App Bar](https://m3.material.io/components/bottom-app-bar/overview) component.

Ushie:
I'd completely forgotten about that, looks perfect

ILoveOpenSourceApplications:
#886 can be closed if this is implemented right?

oSumAtrIX (approved):

Ushie:
> #886 can be closed if this is implemented right?

Yes

Ushie (changes requested):
Use [Bottom App Bar](https://m3.material.io/components/bottom-app-bar/overview) for the patch result options

Ushie:
@PalmDevs Need a bit of help, what should the copy logs button do?:
- Copy logs as text
- Copy file uri with the logs as the content
- Open share menu (current behavior)


image

Ushie:
Also, should the Install FAB contain a label? or should it be Icon only

PalmDevs:
> @PalmDevs Need a bit of help, what should the copy logs button do?:
> - Copy logs as text
> - Copy file uri with the logs as the content
> - Open share menu (current behavior)
> 
> [Screenshot]

I think the best choice would be to open a file picker to let the user save the logs. The user (if they want to share) can access their file manager and share it from there.

PalmDevs:
The icons should also be changed. The icon for exporting the APK should be "Save" (floppy disk icon), and the icon for saving log should be "Post add" icon. References are found on Google's Icons website.

Ushie:
image

ILoveOpenSourceApplications:
> image

This looks good but it would look even better if the style is similar to that of how bottom app bar is in Gmail app. When scrolling down the container disappears and when scrolled up the container reappears. Also the installation button can be similar to the compose button in Gmail app. When still will show the icon and "Install" text next to it and scrolling down will shrink it to just the icon.

Demo:

https://github.com/ReVanced/revanced-manager/assets/117499019/402975df-154a-4066-81c6-6cd09f0d704f

Ushie:
The FAB in BottomAppBar can't be shown separetely, its part of the BottomAppBar, if the BottomAppBar gets hidden then the FAB does too

ILoveOpenSourceApplications:
Under scrolling in guidelines,  my opinion is represented and can be seen in action. Please have a look at it. This FAB implementation can also benefit throughout the UI. When clicking the "Storage" button, "Patch" button and now "Install" button.

Ushie:
I looked into hiding on scroll, and I think the implementation is rather annoying just for something I personally think has no use in this context

Maybe if @validcube wants to add hide on scroll, they can but I personally wont (I dont know flutter enough to do it in a way I think is good code)

As for using Tonal IconButtons, I tried both standard and tonal versions and I prefer tonal, I don't see specs that speak against using it

PR #1063, opened 2023-07-29
feat: allow control over patches update

TheAabedKhan:
This PR aims to allow users to choose whether the ReVanced Patches are updated to the latest one automatically or manually at anytime they want. Right now, users are automatically served with latest patches. This results in users not noticing a new patch or new patchable app easily. When this PR gets merged, they will be able to read the changelogs when updating manually and make the decision to update accordingly.
patch_update_final

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -172,7 +172,7 @@
         "sourcesResetDialogTitle": "Reset",
         "sourcesResetDialogText": "Are you sure you want to reset custom sources to their default values?",
         "apiURLResetDialogText": "Are you sure you want to reset API URL to its default value?",
-        "sourcesUpdateNote": "Note: Patches will be updated to the latest version automatically.",
+        "sourcesUpdateNote": "Note: ReVanced Patches will be updated to the latest version automatically.",
 
         "apiURLLabel": "API URL",
         "apiURLHint": "Configure your custom API URL",
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -1,7 +1,7 @@
 {
     "okButton": "OK",
     "cancelButton": "Cancel",
-    "quitButton": "Quit App",
+    "quitButton": "Quit",
     "updateButton": "Update",
     "enabledLabel": "Enabled",
     "disabledLabel": "Disabled",
@@ -36,8 +36,8 @@
         "updateChangelogTitle": "Changelog",
 
         "patchesConsentDialogText": "ReVanced Patches need to be downloaded to patch apps.",
-        "patchesConsentDialogText2": "Downloading them from the API {url} may reveal your IP address to the server.",
-        "patchesConsentDialogText3": "Auto update the patches",
+        "patchesConsentDialogText2": "This will reveal your IP address to {url}.",
+        "patchesConsentDialogText3": "Auto update",
         "patchesConsentDialogText3Sub": "You can still change this in the settings later",
 
         "notificationTitle": "Update downloaded",
@@ -194,7 +194,7 @@
         "logsLabel": "Logs",
         "logsHint": "Share Manager's logs",
 
-        "autoUpdatePatchesHint": "Automatically update ReVanced Patches to the latest stable version",
+        "autoUpdatePatchesHint": "Automatically update ReVanced Patches to the latest version",
         "experimentalUniversalPatchesLabel": "Experimental universal patches support",
         "experimentalUniversalPatchesHint": "Display all applications to use with universal patches, loading list of apps may be slower",
         "experimentalPatchesLabel": "Experimental patches support",
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -31,6 +31,7 @@
         "installUpdate": "Continue to install the update?",
 
         "updateDialogTitle": "Update Manager",
+        "updatePatchesDialogTitle": "Update Patches",
         "updateChangelogTitle": "Changelog",
 
         "notificationTitle": "Update downloaded",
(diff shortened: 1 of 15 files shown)


assets/i18n/en_US.json
```diff
@@ -31,6 +31,7 @@
         "installUpdate": "Continue to install the update?",
 
         "updateDialogTitle": "Update Manager",
+        "updatePatchesDialogTitle": "Update Patches",
```
oSumAtrIX, line 34:
```suggestion
        "updatePatchesDialogTitle": "Update ReVanced Patches",
```

assets/i18n/en_US.json
```diff
@@ -171,6 +172,7 @@
         "sourcesResetDialogTitle": "Reset",
         "sourcesResetDialogText": "Are you sure you want to reset custom sources to their default values?",
         "apiURLResetDialogText": "Are you sure you want to reset API URL to its default value?",
+        "sourcesUpdateNote": "Note: Patches will be updated to the latest version automatically.",
```
oSumAtrIX, line 175:
```suggestion
        "sourcesUpdateNote": "Note: ReVanced Patches will be updated to the latest version automatically.",
```

oSumAtrIX (changes requested):
Name might be too long

oSumAtrIX:
Is there a popup asking the user to update the patches and ReVanced Manager automatically or manually at least once?
Is there also an option to change the setting in the settings? 
Does the popup let the user know about being able to change the setting later in the settings?
In case ReVanced Manager is launched the first time, how does download behave?

TheAabedKhan:
> asking the user to update the patches and ReVanced Manager automatically or manually 

Updating both patches and manager automatically or updating patches automatically when ReVanced Manager is updated?

> Is there also an option to change the setting in the settings?
Does the popup let the user know about being able to change the setting later in the settings?

I will add them accordingly.

>In case ReVanced Manager is launched the first time, how does download behave?

Latest patches from default repo.

assets/i18n/en_US.json
TheAabedKhan: Adding `ReVanced` before Patches will make the UI inconsistent with line 33.

oSumAtrIX: "ReVanced Patches" is the "name". You can either refer to the patches as "the patches" or "ReVanced Patches" but not "the Patches" or "Patches".

oSumAtrIX:
> Updating both patches and manager automatically or updating patches automatically when ReVanced Manager is updated?

Patches should update independently with ReVanced Manager, both manually and automatically.

TheAabedKhan:
>Patches should update independently with ReVanced Manager, both manually and automatically.

How about adding a toggle in settings page for auto updating the patches. If the toggle is unchanged by the user, then whenever new update for patches shows up, we can ask them to whether to update automatically from next time.

>Automatically without consent by the user? 

This is exactly what happens in the current version. Manager uses the default repo to get the patches info. It downloads the patches at the time of patching only.

oSumAtrIX:
> How about adding a toggle in the settings page for auto-updating the patches? If the toggle is unchanged by the user, then whenever a new update for patches shows up, we can ask them whether to update automatically next time.

That sounds good. Slight changes I recommend would be the following:

1. If the user has never seen the auto-update popup yet, show it. 
2. The popup could look like this (not all details required): 
   image

> This is exactly what happens in the current version. The manager uses the default repo to get the patch's info. It downloads the patches at the time of patching only.

Since this PR is labeled "allow control over patches update," it would be beneficial to show an initial popup: "ReVanced Manager needs ReVanced Patches to work, download?" when ReVanced Manager has never downloaded patches. The popup can look similar to the one in the image above with these options to remember or update manually, for example. Since it is the initial download and ReVanced Manager can not work, if the patches have not been downloaded at least once, the popup is not dismissable.

assets/i18n/en_US.json
```diff
@@ -31,8 +32,14 @@
         "installUpdate": "Continue to install the update?",
 
         "updateDialogTitle": "Update Manager",
+        "updatePatchesDialogTitle": "Update ReVanced Patches",
         "updateChangelogTitle": "Changelog",
 
+        "patchesConsentDialogText": "ReVanced Patches need to be downloaded to patch apps.",
+        "patchesConsentDialogText2": "Downloading them from the API {url} may reveal your IP address to the server.",
```
oSumAtrIX, line 39:
```suggestion
        "patchesConsentDialogText2": "This will reveal your IP address to {url}.",
```

assets/i18n/en_US.json
```diff
@@ -1,6 +1,7 @@
 {
     "okButton": "OK",
     "cancelButton": "Cancel",
+    "quitButton": "Quit App",
```
oSumAtrIX, line 4:
```suggestion
    "quitButton": "Quit",
```

oSumAtrIX (approved):
When changing sources, the warning dialog may be show again, because the IP will be revealed to it.

assets/i18n/en_US.json
```diff
@@ -171,6 +178,7 @@
         "sourcesResetDialogTitle": "Reset",
         "sourcesResetDialogText": "Are you sure you want to reset custom sources to their default values?",
         "apiURLResetDialogText": "Are you sure you want to reset API URL to its default value?",
+        "sourcesUpdateNote": "Note: ReVanced Patches will be updated to the latest version automatically.",
```
TheAabedKhan, line 181: >When changing sources, the warning dialog may be show again, because the IP will be revealed to it.

@oSumAtrIX How about merging the warning text here?

oSumAtrIX: Can you clarify what you mean by that?

TheAabedKhan: Instead of another dialog
image

oSumAtrIX: Works

oSumAtrIX:
@Aunali321 Please review this.

Aunali321 (approved):
Thanks!

PR #1068, opened 2023-08-01
fix: keystore password dialog showing up before importing

### LINKED ISSUE #1033: bug: keystore password pop up dialog
### Type

Error while running the manager

### Bug description

I am really confused by the keystore pop up timing. Should it pop up when the user clicks the import option within settings or after the user selects the keystore file and then the keystore pops up for confirmation. Because right now it's popping up in both the situations.

### Steps to reproduce

1) Open the manager and go into the settings
2) Tap on import keystore
3) Notice that the keystore password pop up comes first after which the user is directed to the storage for selecting the keystore file.
4) By simply going back or after selecting a keystore file, returning to the manager, the user is once again welcomed by the keystore password popup.

### Android version

Android 13

### Manager version

1.4.1

### Target package name

ReVanced Manager 

### Target package version.

Any

### Installation type

Non-root

### Patches selected.

Nil

### Device logs (exported using Manager settings).

```shell
Nil
```

### Installer logs (exported using Installer menu option) [unneeded if the issue is not during patching].

_No response_

### Screenshots or video

https://github.com/ReVanced/revanced-manager/assets/117499019/7f858d71-8c02-418c-969e-53b994a00842

### Solution

_No response_

### Additional context

_No response_

### Acknowledgments

- [X] I have searched the existing issues; this is new and no duplicate or related to another open issue.
- [X] I properly filled out all of the requested information in this issue.
- [X] The issue is solely related to ReVanced Manager and not caused by patches.

TheAabedKhan:
This PR fixes #1033

### DIFF
--- a/lib/ui/widgets/settingsView/settings_export_section.dart
+++ b/lib/ui/widgets/settingsView/settings_export_section.dart
@@ -73,10 +73,12 @@ class SExportSection extends StatelessWidget {
             ),
           ),
           subtitle: I18nText('settingsView.importKeystoreHint'),
-          onTap: () {
-            _settingsViewModel.importKeystore();
+          onTap: () async{
+            await _settingsViewModel.importKeystore();
             final sManageKeystorePassword = SManageKeystorePassword();
-            sManageKeystorePassword.showKeystoreDialog(context);
+            if(context.mounted){
+              sManageKeystorePassword.showKeystoreDialog(context);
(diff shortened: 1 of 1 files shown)


validcube:
I would like if the solution would be after keystore have been selected, the dialog to set keystore password come up. Overrall seems fine to me

PR #1077, opened 2023-08-03
chore: merge `dev` to `main`

Ushie:


pubspec.yaml
ILoveOpenSourceApplications, line 1: Shouldn't it be 
```java
version: 1.5.1+100500001
```  
?

ILoveOpenSourceApplications:
I didn't mean to be a reviewer, just a simple doubt considering the previous versions. Also after this commit was pushed, the previous release has also disappeared.

pubspec.yaml
validcube: It should :skull:

validcube: > Shouldn't it be 
> ```java
> version: 1.5.1+100500001
> ```  
> ?

@Ushie

validcube (changes requested):
See comment

pubspec.yaml
Ushie: The versioncode doesn't matter, it was the way it was for the Semantic Releases CI, which we no longer use, and as it doesn't matter there's no need to put more effort than just bumping it by 1

Ushie:
> I didn't mean to be a reviewer, just a simple doubt considering the previous versions. Also after this commit was pushed, the previous release has also disappeared.

Both releases contained upstream bugs, the correct way would've been to downgrade ReVanced Patcher to 11.0.4 and make a new release though, but it's fine

ILoveOpenSourceApplications:
Why downgrade the ReVanced Patcher version though?

Ushie:
The ReVanced Patcher is where the issues where introduced in

PR #1082, opened 2023-08-04
Added more to Readme.md for easy accessibility of info. From docs

Unity-Blaster:
### **Reason:**
There were some Feature Requests & Bug Reports regarding having no knowledge on how to do some things related to the Manager, and that they couldn't find the files under /docs. 

### **What I did:**
Added a section at the bottom of the main README.md containing links to those files.

### **Changing These Contents:**
The wording can be changed, as it is suitable...


oSumAtrIX (commented):
This would require us to maintain two places of documentation, one here and the other in the ReVanced Manager repository. Instead, the documentation should merely linked, though it is planned to move this to ReVanced Website.

PR #1087, opened 2023-08-04
fix(patch-item): hide universal patches if not enabled

### LINKED ISSUE #1080: bug: universal patches are always shown
### Bug description

Universal patches are shown in the patches selector screen regardless of whether or not "Experimental universal patches support" is enabled in ReVanced Manager settings

### Solution

Hide universal patches in patches selector if the setting is not toggled on

### Acknowledgments

- [X] I have searched the existing issues; this is new and no duplicate or related to another open issue.
- [X] I properly filled out all of the requested information in this issue.
- [X] The issue is solely related to ReVanced Manager and not caused by patches.

TheAabedKhan:
Fixes #1080

### DIFF
--- a/lib/services/patcher_api.dart
+++ b/lib/services/patcher_api.dart
@@ -103,16 +103,20 @@ class PatcherAPI {
   }
 
   List<Patch> getFilteredPatches(String packageName) {
-    if (!filteredPatches.keys.contains(packageName)) {
-      final List<Patch> patches = _patches
-          .where(
-            (patch) =>
-                patch.compatiblePackages.isEmpty ||
-                !patch.name.contains('settings') &&
-                    patch.compatiblePackages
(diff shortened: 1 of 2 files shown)


Ushie:
Also, what do you think about separating them into different sections? So scrolling to the bottom would display the universal patches section

TheAabedKhan:
>Also, what do you think about separating them into different sections? So scrolling to the bottom would display the universal patches section

It can help achieve better UX when selecting patches. Will create a new PR for this.

validcube (changes requested):
This changes looks good, but the patch count in app selector will not display correctly and this can cause confusion.

The patch count will still display (example) 60 patches instead of 52 patches regardless if the settings enable or not.

validcube (approved):
💫 Changes approved

PR #1092, opened 2023-08-05
fix(patches-selector): separate all universal patches to the bottom

TheAabedKhan:
https://github.com/ReVanced/revanced-manager/assets/39409020/1ccda832-910a-4016-a6b6-09b461c38374


validcube (approved):
💫 Changes tested and approved.

The text looks small to me, I could be wrong, @PalmDevs what do you think?

Issue #999, opened 2023-07-08
feat: clarify patches card


taku-nm:
### Type

Cosmetic

### Issue

As seen [here](https://discord.com/channels/952946952348270622/1122167988276187228/1127032172491063366) ; 

Users that patched a while ago can forget, that the list of selected patches is clickable. This is a problem when the old list of patches got automatically loaded. 

### Feature

copied from #967

an edit button, preferably on the top right of the app and patch selection card. If it were at the bottom of the list, I can already see how people would miss it by not scrolling down;

### Motivation

same as issue #967
except that playback issues are less of a problem for now;

### Additional context

This came to be after a _lengthy_ [discussion](https://discord.com/channels/952946952348270622/952946952348270626/1127022080840445953). 

This link is just a refence for osu and me

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

ILoveOpenSourceApplications:
This issue was resolved with #1099 and can be closed right?

taku-nm:
> This issue was resolved with #1099 and can be closed right?

No the issue just went stale

oSumAtrIX:
ReVanced Manager will soon be released as [a full rewrite](https://github.com/ReVanced/revanced-manager/tree/compose-dev). Over time, many feature requests have been made for the current version of ReVanced Manager, but to accelerate the development of the new version of ReVanced Manager, we have to focus on it, which I am closing this issue according to the [contribution guidelines](https://github.com/ReVanced/revanced-manager/blob/main/CONTRIBUTING.md#-submitting-a-feature-request):

> Requests can be accepted or rejected at the discretion of maintainers of ReVanced Patches

This feature request has been closed for one or more of the following reasons:

- Stale for too long/ Low priority/ Maintainers un-interested/ Too complex to realize
- Feature planned/ already implemented for the new version of ReVanced Manager

Regardless, anyone who comes across this feature request can PR it to the current version of ReVanced Manager until the new version is ready. Thanks for your feature request!

Issue #808, opened 2023-04-21
Add a **New** text on latest patches. 


fard33n:
### Type

Cosmetic

### Issue

Since a lot of patches are being added, it has become a but cluttered especially for the ReVanced YouTube patches. It is very difficult to keep track of the patches added in the manager. 

### Feature

The *NEW* highlight over the latest patches added with be very helpful for users are repatching the app for updates. Will help to create a distinction between older patches that may or may not have been already selected and the lastest added ones so that the user can easily go through them. 

### Motivation

This feature should be considered to improve the user experience which selecting patches whose number is increasing day by day. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

oSumAtrIX:
This feature should work as follows:

- If a new patch is recognized by ReVanced Manager, it must label it as new and highlight it by pinning it at the top of the patches list. Depending on if the patch is included by default or not, select it, or don't select it by default
- If the new patch is included by default, the patch should be included by default by ReVanced Manager provided that the user has never made their own custom selection even once and left the selection as it was originally
- If a patch has been deleted, ReVanced Manager must issue a notice.
- If a new patch is available or a patch has been deleted, the patcher screen of ReVanced Manager should let the user know
- Once the user patches their app at least once after the addition or removal of patches, all current "New" labels or notices must be removed until the next time, when new patches are added or removed

fard33n:
Yes that is a better solution.

ILoveOpenSourceApplications:
Completed with the merge of #1099.

PR #1099, opened 2023-08-06
feat(patches-selector): show `New` tag for new patches

TheAabedKhan:
image

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -133,7 +133,7 @@
         "unsupportedDialogText": "Selecting this patch may result in patching errors.\n\nApp version: {packageVersion}\nSupported versions:\n{supportedVersions}",
         "unsupportedPatchVersion": "Patch is not supported for this app version. Enable the experimental toggle in settings to proceed.",
 
-        "newPatchDialogText": "This is a new patch.",
+        "newPatchDialogText": "This is a new patch that has been added since the last time you have patched this app.",
         "newPatch": "New patch"
     },
     "installerView": {
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -10,6 +10,7 @@
     "yesButton": "Yes",
     "noButton": "No",
     "warning": "Warning",
+    "new": "New",
     "navigationView": {
         "dashboardTab": "Dashboard",
         "patcherTab": "Patcher",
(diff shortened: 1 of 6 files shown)


assets/i18n/en_US.json
```diff
@@ -130,7 +131,10 @@
     },
     "patchItem": {
         "unsupportedDialogText": "Selecting this patch may result in patching errors.\n\nApp version: {packageVersion}\nSupported versions:\n{supportedVersions}",
-        "unsupportedPatchVersion": "Patch is not supported for this app version. Enable the experimental toggle in settings to proceed."
+        "unsupportedPatchVersion": "Patch is not supported for this app version. Enable the experimental toggle in settings to proceed.",
+
+        "newPatchDialogText": "This is a new patch.",
```
oSumAtrIX, line 136:
```suggestion
        "newPatchDialogText": "This is a new patch that has been added since the last time you visited the patch selection screen.",
```

oSumAtrIX (approved):
In the "main" patcher screen, do you hint that new patches are available? What about patches that have been removed?

assets/i18n/en_US.json
TheAabedKhan: How about "...since the last time you patched the app" ?

TheAabedKhan: Because that's the case.

oSumAtrIX:
Please check if this PR is conform with [this](https://github.com/ReVanced/revanced-manager/issues/808#issuecomment-1660581278) comment.

TheAabedKhan:
>In the "main" patcher screen, do you hint that new patches are available?

I think we can do this while working on #999.

>What about patches that have been removed?

We can't know whether a particular patch is actually removed or just name changed. So, how about explaining the users, when they click on the `New` tag, the patch may be new or modified?

oSumAtrIX:
I think if you say "new or modified" it may confuse the user in what this means.

TheAabedKhan:
>I think if you say "new or modified" it may confuse the user in what this means.

So, better not say anything?

oSumAtrIX:
If you don't say anything in the popup, the user will maybe not understand what "new" means. It is better to explain what it means.

TheAabedKhan:
>If you don't say anything in the popup, the user will maybe not understand what "new" means. It is better to explain what it means.

I mean, not saying anything related to removal of patches.

oSumAtrIX:
If a patch is missing the user will wonder when patching the next time, why the patch didn't work. It is better to clarify in this case.

TheAabedKhan:
>If a patch is missing the user will wonder when patching the next time, why the patch didn't work. It is better to clarify in this case.

Then, for all those patches that are not there after patches update, we can show a Removed/Modified dialog listing the name of those patches or we can show them in the patches selector screen with a `Removed` label

oSumAtrIX:
Sure, this can be added in another PR if you like. The remaining points from the comment linked should be ticked for this PR.

assets/i18n/en_US.json
oSumAtrIX:
```suggestion
        "newPatchDialogText": "This patch has been added during the time you have last patched.",
```

validcube (approved):
🎉 Code looks fine to me!

FraneCaleta (approved):

PR #1102, opened 2023-08-07
fix: pair integrations and patches updates

TheAabedKhan:
This PR fixes the issue related to the integrations file being updated to the latest version automatically, causing issues in the patching process when the ReVanced Patches being used are not the latest.


validcube (approved):
🎉 LGTM

Issue #483, opened 2022-11-05
Stuck at "creating working directory"


Bluscream:
### Type

Error while running the manager

### Bug description

![Screenshot_2022-11-05-13-09-34-800](https://user-images.githubusercontent.com/3318223/200119179-4df7313f-33c3-4d1f-a5bd-daa6cf2b5d35.jpeg)

### Steps to reproduce

Open vanced manager
Select youtube and any patch
Wait

### Android version

Idk?

### Manager version

Version 0.0.39 (39)

### Target package name

youtube stock

### Target package version.

17.43.36

### Installation type

Root

### Patches selected.

First one

### Device logs (exported using Manager settings).

```shell
No idea
```

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [x] I have written a short but informative title.
- [x] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

vibbow:
I have the same issue

Samsung One UI 4.1 (Android 12)
Revanced manager 0.0.40

target package:
youtube stock 17.43.36

Holguitar:
Same here on different devices (Moto G5+ A8 up to Xiaomi Mi9T A11) rooted and non rooted.

HalfArrow:
Same issue, Android 13 One UI 5.

Mygod:
Clearing app data fixed it for me.

trmdi:
> Clearing app data fixed it for me.

But it clears the Cert key too, that sucks.

oSumAtrIX:
Did you exit the app while it was patching?

trmdi:
> Did you exit the app while it was patching?

Can't reproduce in this way.

oSumAtrIX:
Not sure what you are saying. Please explain.

trmdi:
> Not sure what you are saying. Please explain.

I mean I tried to exit the app while patching as you said to reproduce this bug but I can't.

oSumAtrIX:
I asked wether you exited the app; I didn't say you should.

trmdi:
> I asked wether you exited the app

No, I didn't.

oSumAtrIX:
Please attach a complete recording of the following steps to understand your issue better:

1. Open and close ReVanced Manager
2. Review the settings of ReVanced Manager
3. Patch the app

terrytw:
[...]

PR #1114, opened 2023-08-09
feat: patch apps without internet

TheAabedKhan:
This PR allows the manager to make use of the cache files (patches and integrations) while offline

When the files are not there and also no internet, then the users just see, "no applications" text in the app selector page like they used to 

https://github.com/ReVanced/revanced-manager/assets/39409020/1aa7dcd4-d4e6-41d9-9051-8389a95123b1

### DIFF
--- a/lib/services/github_api.dart
+++ b/lib/services/github_api.dart
@@ -241,7 +241,10 @@ class GithubAPI {
   }
 
   Future<List<Patch>> getPatches(
-      String repoName, String version, String url) async {
+    String repoName,
+    String version,
+    String url,
+  ) async {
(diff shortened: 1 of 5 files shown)


Ushie:
Maybe this fixes #483 ?

TheAabedKhan:
> Maybe this fixes https://github.com/ReVanced/revanced-manager/issues/483 ?

Yes. That happens due to no active internet connection and no response from GitHub server.

Aunali321 (approved):

oSumAtrIX (approved):

validcube (approved):
💫 Tested and approved!

Issue #990, opened 2023-07-02
feat: be able to copy the recommended version


wiiznokes:
### Type

Functionality

### Issue

because it takes a long time to have to remember the right version and write it in a search engine.

### Feature

be able to copy the recommended version with its name, example: `twitter apk version v9.71.0-release0`

### Motivation

because it simple to implement

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

AdiAr11:
Hi, I would like to work on this issue. Can someone guide me where can i find the recommended version? where would the user have to click to be able to copy the recommended version with its name? Thank you.

wiiznokes:
I feel like the best way of doing this is to add a screen when we click on a app which is not installed.
The screen will have a button 
- to comme back to the list
- copy the recommend version
- maybe open it on a navigator idk
- the floating button storage

wiiznokes:
The version seems to be fetched via a PatcherApi. See the viewmodel:
https://github.com/ReVanced/revanced-manager/blob/main/lib/ui/views/app_selector/app_selector_viewmodel.dart

oSumAtrIX:
I think instead of copying; it could launch the browser. In this case, it should be accompanied by a  "link" icon to signify an outgoing link.

oSumAtrIX:
Next to the package name or version a "link" icon can be added to signify an outgoing link when the label is clicked on.

AdiAr11:
@oSumAtrIX , sounds good. will try to implement this. could you help me finding the file where i will have to make this changes if possible? I'll also try to find it as well. Thank you!

wiiznokes:
If we add a new screen, the list will be more "visible".
And it will be easier to add functionality like predefined website like appmirror etc.
But I am not a maintainer.

AdiAr11:
@oSumAtrIX No worries, I will find it. Thank you

wiiznokes:
I just saw it has been implemented, nice works!

Indranil012:
Should be closed

PR #1151, opened 2023-08-26
feat: ability to search query for suggested version

AdiAr11:
Implemented the feature requested in issue #990


AdiAr11:
https://github.com/ReVanced/revanced-manager/assets/54585121/e573ec81-d308-4ca2-8543-13fc6e27099d

AdiAr11:
@oSumAtrIX @validcube Do you want the link on uninstalled apps as well? I have added the link icon on the selected app as well. The icon will be shown when there is a specific suggested version available. Please let me know if more changes are requried.

oSumAtrIX:
The link icon can be on every card. What I suggest is that you visually link the link button with the suggested version, so that it is clear what the link is. Bring them nearer together, next to each other. Space is a problem so that has to be figured out. You could make the link button smaller, bring it next to the suggested version, and make the suggested version label inclusive the icon clickable

AdiAr11:
You mean make the suggested version text clickable, right? Yes, that sounds like a good idea. And will make the icon logo smaller. 
One doubt though, should i add this to every card? Even for which the suggested version is all version? Will add it for uninstalled apps as well.

oSumAtrIX:
Here is a naive illustration of what I mean:

image

oSumAtrIX:
> One doubt though, should i add this to every card? Even for which the suggested version is all version? Will add it for uninstalled apps as well.

Yes. The reason for that is you may want to download an app to patch you do not have installed already.

AdiAr11:
> > One doubt though, should i add this to every card? Even for which the suggested version is all version? Will add it for uninstalled apps as well.
> 
> Yes. The reason for that is you may want to download an app to patch you do not have installed already.

Understood. Will do it!

AdiAr11:
https://github.com/ReVanced/revanced-manager/assets/54585121/3e58ebb8-e0c2-44a4-9b49-9c1f15d7caa7

@oSumAtrIX please have a look at this. If any more changes are required, please let me know.

oSumAtrIX:
Keep in mind, my illustration was naive as mentioned previously, it shouldn't look like this, a more sophisticated design should be used.

AdiAr11:
> Keep in mind, my illustration was naive as mentioned previously, it shouldn't look like this, a more sophisticated design should be used.

Could you please provide some images or something on how the design should look like then?

PalmDevs:
At this point, it may as well be better to go with chips. Like the same style of chips that appear when the version is incompatible (the [ ⚠️ Warning ] chip).

oSumAtrIX:
@AdiAr11 Unfortunately I do not have anything in mind right now

AdiAr11:
> @AdiAr11 Unfortunately I do not have anything in mind right now

Yes same with me, xd. Hows the one i created looking though? How about instead i underline it?

oSumAtrIX:
I'd like @PalmDevs to look into this, as they might have better ideas than me.

validcube (approved):
🎉 Looks fine to me!

PalmDevs:
> I'd like @PalmDevs to look into this, as they might have better ideas than me. 

As mentioned before, I've suggested using chips (the same ones used in the Warning chip in the patch selector area).

PalmDevs:
I did not mean to lock 😅

lib/ui/views/patcher/patcher_viewmodel.dart
```diff
@@ -161,6 +162,40 @@ class PatcherViewModel extends BaseViewModel {
...
+    final String apk = FlutterI18n.translate(
+      context,
+      'appSelectorCard.apk',
+    );
+    final String version = FlutterI18n.translate(
+      context,
+      'appSelectorCard.version',
+    );
+    final String suggestedVersion =
+        _patcherAPI.getSuggestedVersion(selectedApp!.packageName);
```
Ushie, line 186: What if the suggested version is Any/All? that doesn't make sense in a web search

lib/ui/widgets/appSelectorView/installed_app_item.dart
```diff
@@ -72,16 +74,39 @@ class _InstalledAppItemState extends State<InstalledAppItem> {
...
+                                      ? FlutterI18n.translate(
+                                          context,
+                                          'appSelectorCard.allVersions',
+                                        )
+                                      : 'v${widget.suggestedVersion}',
+                                },
+                              ),
+                              // if(widget.suggestedVersion.isNotEmpty)
+                              const SizedBox(width: 4),
+                              // if(widget.suggestedVersion.isNotEmpty)
```
Ushie, line 102: This should be deleted if it's not of use

lib/ui/widgets/appSelectorView/installed_app_item.dart
```diff
@@ -72,16 +74,39 @@ class _InstalledAppItemState extends State<InstalledAppItem> {
...
+                                translationParams: {
+                                  'version': widget.suggestedVersion.isEmpty
+                                      ? FlutterI18n.translate(
+                                          context,
+                                          'appSelectorCard.allVersions',
+                                        )
+                                      : 'v${widget.suggestedVersion}',
+                                },
+                              ),
+                              // if(widget.suggestedVersion.isNotEmpty)
```
Ushie, line 100: This should be deleted if it's not of use

lib/ui/widgets/appSelectorView/not_installed_app_item.dart
```diff
@@ -66,16 +68,38 @@ class _NotInstalledAppItem extends State<NotInstalledAppItem> {
...
+                                  'version': widget.suggestedVersion.isEmpty
+                                      ? FlutterI18n.translate(
+                                          context,
+                                          'appSelectorCard.allVersions',
+                                        )
+                                      : 'v${widget.suggestedVersion}',
+                                },
+                              ),
+                              const SizedBox(width: 4),
+                              // if(widget.suggestedVersion.isNotEmpty)
```
Ushie, line 95: This should be deleted if it's not of use

Ushie (changes requested):
What about instead of adding this to the card itself, it get's moved to the "Please select APK from storage" dialog?? that makes the most sense in my opinion

image

Considering this dialog only comes up when selecting from installed isnt possible, that means it's the best location to put this into, the current positioning may not be understood as the user is unaware of why they would need to use it

Ushie:
@oSumAtrIX @PalmDevs can I have your opinions [on the above?](https://github.com/ReVanced/revanced-manager/pull/1151#pullrequestreview-1608388405)

PalmDevs:
> What about instead of adding this to the card itself, it get's moved to the "Please select APK from storage" dialog?? that makes the most sense in my opinion
> 
> image
> 
> Considering this dialog only comes up when selecting from installed isnt possible, that means it's the best location to put this into, the current positioning may not be understood as the user is unaware of why they would need to use it
> 

What would that look like?

Ushie:
I was hoping you'd have one, well for now the current design should work

lib/ui/widgets/appSelectorView/installed_app_item.dart
```diff
@@ -72,16 +74,39 @@ class _InstalledAppItemState extends State<InstalledAppItem> {
...
+                              const SizedBox(width: 4),
+                              // if(widget.suggestedVersion.isNotEmpty)
+                              const Icon(
+                                Icons.link,
```
Ushie, line 104: Instead, use [Icons.launch](https://fonts.google.com/icons?icon.set=Material+Icons&selected=Material+Icons+Outlined:launch:&icon.query=launch)

AdiAr11:
> > One doubt though, should i add this to every card? Even for which the suggested version is all version? Will add it for uninstalled apps as well.
> 
> Yes. The reason for that is you may want to download an app to patch you do not have installed already.
Hi @Ushie, this is the reason why i added launcher to every card. 
And I have made the requested changes. Please let me know if any further changes are required.

Ushie:
What about this? https://github.com/ReVanced/revanced-manager/pull/1151#discussion_r1314016179

AdiAr11:
@Ushie, Okay, should i remove this from apps already installed and for which the recommended version is Any or All, right?

Ushie:
Maybe you can just send the same query without `version: suggestedversion`, it looks fine to me then

AdiAr11:
@Ushie, Done with the changes!

lib/ui/views/app_selector/app_selector_viewmodel.dart
```diff
@@ -69,6 +70,36 @@ class AppSelectorViewModel extends BaseViewModel {
...
+    );
+    final String suggestedVersion = getSuggestedVersion(packageName);
+    String stringUrl;
+    if(isInstalled){
+      stringUrl = 'https://www.google.com/$search?q=$appName'
+          '+$apk';
+    }else{
+      stringUrl = 'https://www.google.com/$search?q=$appName'
+          '+$apk+$version+v+$suggestedVersion';
+    }
```
Ushie, line 95: This doesn't seem right to me, why would having the app installed effect the need to include the suggested version?

lib/ui/views/app_selector/app_selector_viewmodel.dart
```diff
@@ -69,6 +70,36 @@ class AppSelectorViewModel extends BaseViewModel {
     return true;
   }
 
+  Future<void> searchSuggestedVersionOnWeb(BuildContext context,
+      {required String packageName, required String appName, bool isInstalled = false,}) async {
```
Ushie, line 74: appName doesn't exist for apps that aren't installed, what's the usage of it here? it's also less reliant, package name tends to give better results

lib/ui/views/app_selector/app_selector_viewmodel.dart
```diff
@@ -69,6 +70,36 @@ class AppSelectorViewModel extends BaseViewModel {
...
+      context,
+      'appSelectorCard.version',
+    );
+    final String suggestedVersion = getSuggestedVersion(packageName);
+    String stringUrl;
+    if(isInstalled){
+      stringUrl = 'https://www.google.com/$search?q=$appName'
+          '+$apk';
+    }else{
+      stringUrl = 'https://www.google.com/$search?q=$appName'
```
Ushie, line 93: App name doesn't exist if the app isn't installed

lib/ui/views/app_selector/app_selector_viewmodel.dart
```diff
@@ -69,6 +70,36 @@ class AppSelectorViewModel extends BaseViewModel {
...
+    );
+    final String version = FlutterI18n.translate(
+      context,
+      'appSelectorCard.version',
+    );
+    final String suggestedVersion = getSuggestedVersion(packageName);
+    String stringUrl;
+    if(isInstalled){
+      stringUrl = 'https://www.google.com/$search?q=$appName'
+          '+$apk';
```
Ushie, line 91: Why is this in a new line?

lib/ui/views/app_selector/app_selector_viewmodel.dart
```diff
@@ -69,6 +70,36 @@ class AppSelectorViewModel extends BaseViewModel {
...
+      'appSelectorCard.search',
+    );
+    final String apk = FlutterI18n.translate(
+      context,
+      'appSelectorCard.apk',
+    );
+    final String version = FlutterI18n.translate(
+      context,
+      'appSelectorCard.version',
+    );
```
Ushie, line 86: You shouldnt translate these strings, for example translating "search" would likely break the search result

Ushie (changes requested):

lib/ui/views/app_selector/app_selector_viewmodel.dart
AdiAr11: @Ushie, Honestly, I can't remember why i added this. Will remove it.

lib/ui/views/app_selector/app_selector_viewmodel.dart
AdiAr11: will use packageName instead.

AdiAr11:
@Ushie I have made further changes. Thank you

lib/ui/views/app_selector/app_selector_viewmodel.dart
```diff
@@ -69,6 +70,21 @@ class AppSelectorViewModel extends BaseViewModel {
     return true;
   }
 
+  Future<void> searchSuggestedVersionOnWeb(
+      {required String packageName,}) async {
+    final String suggestedVersion = getSuggestedVersion(packageName);
+    String stringUrl = 'https://www.google.com/search?q=$packageName'
+        '+apk';
```
oSumAtrIX, line 77: Do not hardcode Google. If possible "launch" a browser search using the default browser.

oSumAtrIX:
@AdiAr11 Is this PR still active?

AdiAr11:
@oSumAtrIX, Sorry, I didn't see your review. Will make the necessary changes.

lib/ui/views/app_selector/app_selector_viewmodel.dart
oSumAtrIX: A quick Google search yielded this: https://stackoverflow.com/questions/2201917/how-can-i-open-a-url-in-androids-web-browser-from-my-application

oSumAtrIX: I am also failing to find solutions to this, this sounds like a really stupid limitation of Android

Ushie: https://developer.android.com/guide/components/intents-common#SearchWeb

AdiAr11: @oSumAtrIX @Ushie This might require native platform implementation, seperate for iOS and Android.

PalmDevs:
Can we see screenshots of this?

AdiAr11:
Hi @PalmDevs . I have attached screen recording of the feature.

AdiAr11:
Almost done with the required changes.

AdiAr11:
@oSumAtrIX @Ushie Done!! Please review.

PalmDevs:
> Hi @PalmDevs . I have attached screen recording of the feature.

I don't see it, where did you attach the recording?

AdiAr11:
> https://github.com/ReVanced/revanced-manager/assets/54585121/3e58ebb8-e0c2-44a4-9b49-9c1f15d7caa7
> 
> @oSumAtrIX please have a look at this. If any more changes are required, please let me know.

@PalmDevs Here you go🙂

AdiAr11:
The current implementation is almost the same, except i am not hard coding google.com anymore. The search query will be passed to the default browser and perform search.

Ushie (approved):

Ushie:
image
image

Please review the UI @PalmDevs

PalmDevs:
@Ushie What does the button/chip do?

Ushie:
it opens a browser window with a search query for the app's apk (with it's suggested version if available)

PalmDevs:
> it opens a browser window with a search query for the app's apk (with it's suggested version if available) 

Current icon looks like it's gonna copy the version. We should switch to something like a globe icon. The size of the "button" is also really small which isn't good for accessibility.

PalmDevs:
I was gonna say chips but it'll look insanely weird. Do we need this in the first place?

Ushie:
It's a QoL improvement until downloading is possible

oSumAtrIX:
Since the label opens an external link, the "External link"/ Popout icon should be used.
Regarding the design, check how using a blue tinted label works instead of a highlighted label background. Also, an icon without text could also work with a tooltip.

PalmDevs:
> Regarding the design, check how using a blue tinted label works instead of a highlighted label background.

Great idea. We can't really hardcode colors though, so how about `primary` colored **bolded** text?

> Also, an icon without text could also work with a tooltip but accessibility should be checked. It would not be clear what it does, but after the user tries it, it would be.

Accessibility properties can be added. I think just a small icon on the right of the text should be enough to indicate it's an action.

validcube:
Is this PR ready to be merge?

AdiAr11:
Hi, what all things should i change then?

Ushie:
TODO:
- [ ] Consistent design across the cards (the suggested version in the patcher screen should be the same design as the one in the app selector screen)
- [ ] Change the link icon to popup icon
- [ ] Find the appropriate colors for the text and background of the suggested version

AdiAr11:
@Ushie, any suggestions for colors?

Ushie:
@oSumAtrIX @PalmDevs Why don't we make it open the browser on card click instead?

Ushie:
> @Ushie, any suggestions for colors?

unfortunately no, i think the background color may be the best there could be rn but unsure about text

PalmDevs:
> @PalmDevs Why don't we make it open the browser on card click instead?

The whole card? Would that not conflict with already-installed applications when you try to select them?

oSumAtrIX:
> @oSumAtrIX @PalmDevs Why don't we make it open the browser on card click instead?

There is no visual hint for the user that it's clickable

> unfortunately no, i think the background color may be the best there could be rn but unsure about text

Simply a more tinted color than the foreground color would work.

Ushie:
I'm closing this PR, as development is focused on the rewrite of the app in the compose-dev branch

Sorry for the time spent, and sorry for the long review times

This PR can be re-opened if you choose to finish it before the release of the rewrite

AdiAr11:
Hi, yes, no worries. Although, just to confirm, what should be changed here? Should i just change and try a few different background colors for the link part?
Also, May i ask what changes are being done in compose-dev branch and if there is something i can do to help?
Thank you.

oSumAtrIX:
Hey, I am considering reopening this PR because it is nearly finished. It would be a pitty to discard the existing work, if that is fine, otherwise close again.

Ushie:
@oSumAtrIX @PalmDevs @AdiAr11 

https://github.com/ReVanced/revanced-manager/assets/79272171/75ce85b2-3f9a-47ee-9823-b2d36d4961d8

AdiAr11:
> @oSumAtrIX @PalmDevs @AdiAr11
> 
>  2023-11-21.02-18-06.mp4

Hi @Ushie are you talking about the double v in the video? In a few apps, v is coming from getSuggestedVersion only.

Ushie:
I did see the double v yeah, I'll fix it in a commit after this PR, is there anything else you'd like to note?

oSumAtrIX:
The double v is caused by one v being hardcoded into the version name of the app.  The only way to properly fix this is to remove our prefix

AdiAr11:
Or if you want, I can first check if version already contains 'v', then i won't add it!

oSumAtrIX:
The version name is what comes after the v. If you remove it, then the version name is incorrect because the apps full version name includes the v. The issue originates from prefixing the version name with a v whereas because it's a name not a code, you are supposed to prefix the version code. This is out of scope for this PR though, so don't worry about it

AdiAr11:
Okay, Thank you !!

PR #1161, opened 2023-08-27
feat: Split "Enable Universal Patches" setting

BenjaminHalko:
The setting "Enable Universal Patches" used to enable universal patches & load unsupported apps, now it is split into 2 settings. This allows universal patches to be used without the need to also load a longer list of apps.


Ushie:
Instead, I propose speeding up the loading of the apps screen

We could do it this way:

- Load the apps that have patches from the patch bundle instead of installed apps
- Now load the installed apps and for the patch count parameter you can count the amount of packages without explicit package compatibility and assign it to all of them

BenjaminHalko:
@Ushie I attempted to improve the loading speed here #1166

I suspect that it is the `DeviceApps.getInstalledApplications` and `DeviceApps.getApp` functions that are slowing it down, not getting the patch count for each app.

I did a test where it would get all installed apps but not add it to the list, thus skipping over getting the patch count. The overall type it takes to load stayed the same.

PR #1166, opened 2023-08-28
fix: Improve app list loading speed

BenjaminHalko:
An attempt to improve the app page loading speed when universal patches have been enabled.

I suspect that the `DeviceApps.getInstalledApplications` and `DeviceApps.getApp` functions are slowing it down, though I have applied some other optimizations to improve the speed slightly.

### DIFF
--- a/lib/services/patcher_api.dart
+++ b/lib/services/patcher_api.dart
@@ -25,6 +25,8 @@ class PatcherAPI {
   late Directory _tmpDir;
   late File _keyStoreFile;
   List<Patch> _patches = [];
+  List<Patch> _universalPatches = [];
+  List<String> _compatiblePackages = [];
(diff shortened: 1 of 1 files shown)


BenjaminHalko:
After further inspection, it seems like when universal patches have been enabled, the patcher loads the list of apps twice. It first gets the list of all apps, then gets each app individually.

This has now been edited to get the list of apps only once, using `DeviceApps.getInstalledApplications`. Seems like it takes roughly half the time to get all apps now.

NelsonSolis90 (approved):

validcube (approved):
🥞 Approved, It does feel faster to me (? LET FIND OUT IN PRODUCTION :trollface: that was a joke, I did review it)

PR #1177, opened 2023-08-28
chore: merge `dev` to `main`

Ushie:


validcube:
Before merging to `main` and release new version

I would like to twerk the patch selector UI by increasing the text size of "universal patches" and add an divider.

PR #1223, opened 2023-09-01
fix: providers.gradleProperty

BenjaminHalko:
Because of gradle/gradle#23572, `providers.gradleProperty` can not see other `gradle.properties` files. This should temporarily fix that until the Gradle issue is resolved.

This now allows the use of `~/.gradle/gradle.properties` to store GitHub tokens instead of `~/gradle.properties`.


Axelen123 (commented):
Does this still work if you set the properties globally in your home directory instead of in the module?

BenjaminHalko:
Currently no, but I can easily change it to look in both.

BenjaminHalko:
@Axelen123 Done

Axelen123 (approved):

PR #1227, opened 2023-09-02
docs: update readme badges

BenjaminHalko:
- The 2 commit badges have been updated to point to `compose-dev` instead of `main`.
- The `LICENSE` badge now points to the revanced-manager license file on `main` instead of attempting to point to the `compose-fix` license which does not exist


validcube (approved):
Whoops

PR #1228, opened 2023-09-02
ci: build pull requests

BenjaminHalko:
This should hopefully allow `compose-dev` pr's to build the manager. (Though I haven't been able to test it much)

This is a combination of [flutter pr build](https://github.com/ReVanced/revanced-manager/blob/main/.github/workflows/pr-build.yml) and [compose release](https://github.com/ReVanced/revanced-manager/blob/compose-dev/.github/workflows/release.yml).

This workflow is not for building & uploading to GitHub releases, it is meant for building/verifying compose pull requests.


validcube (approved):
I have no objections, would anyone from team like to comment on this?

oSumAtrIX:
This duplicates an existing workflow in the wrong branch. The PR target should be the compose branch and it should replace an existing release workflow

Ushie:
The workflow will not run if it doesn't reach the main (dev), it can't simply stay in the compose-dev branch

BenjaminHalko:
This workflow is meant to mimic the [pr-build](https://github.com/ReVanced/revanced-manager/blob/dev/.github/workflows/pr-build.yml) workflow for Flutter and should be merged with main so that pull requests for compose can verify that it compiles / can be easily downloadable for testing purposes.

#1228 is the release workflow for compose that will stay on `compose-dev` until `compose-dev` is merged with main.

oSumAtrIX:
The workflow will also run if not on the main or dev branch, so it should be introduced with the compose branch if it is meant to create compose releases.

BenjaminHalko:
@oSumAtrIX If I were to switch the branch to `compose-dev` would I need to create a separate pull request? As this on targets `dev` and switching the branch inside the pr seems to create conflicts (unless I am just doing it wrong).

oSumAtrIX:
The conflicts happen because your source branch is based on the current dev branch. If you simply cherry pick your commits on top of the compose branch and force push it to the PRs source branch, this should solve the merge conflicts when you change the target to the compose branch

oSumAtrIX:
Currently, there is a workflow present in the compose-dev branch, but it does not run @PalmDevs may know more about this.

Ushie:
> Apparently, the workflow should also run if not on the main or dev branch, so it should be introduced with the compose branch if it is meant to create compose releases.

It simply uploads the build as an artifact in Actions, it doesn't create a release

oSumAtrIX:
What do you mean by "it"? The workflow introduced by this PR or the existing one in the compose dev branch?

Ushie:
The one introduced by this PR

oSumAtrIX:
Sounds good, changing the target branch should be enough then.

BenjaminHalko:
@oSumAtrIX done

Ushie:
I don't get it, do you not understand the intention behind this PR or am I the one misunderstanding?

We want to merge this into the default branch, which is currently the dev branch

Otherwise this workflow won't run, and merging it into compose-dev won't make the workflow run, This workflow exists so PRs targetting compose-dev get built and published as artifacts in the workflow, merging this to compose-dev won't allow us to have PR builds until the merge of compose-dev into the default branch

BenjaminHalko:
@Ushie I did some tests and for pull request workflows, it seems like the workflow has to be in the branch getting merged.

Example:
- branch a has the workflow pr-build.yml. 
- branch b has the workflow pr-build-b.yml
- if the pr targets branch a, pr-build.yml will trigger
- if the pr targets branch b, pr-build-b.yml will trigger
- if the pr targets branch c (which has no pull request workflows), nothing will happen.

Ushie:
It seems like what I said wasn't true, I'm not sure why I was strongly under the assumption that workflows are only recognised in the default branch, my comments can be ignored

BenjaminHalko:
@Ushie No worries!
@oSumAtrIX Is this ready to merge now?

Ushie:
Not yet, for now it should be scoped to specific branches, "compose*" can be used to specify that

Ushie:
Otherwise it'll run on branches unrelated to compose, which will waste action time and create failed workflows for unrelated PRs

BenjaminHalko:
@Ushie Because only branches that have this workflow trigger this workflow, only compose branches will use it.

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,48 @@
+name: PR Build Compose
+
+on:
+  pull_request:
+    paths-ignore:
+      - "**.md"
+      - ".idea/**"
```
oSumAtrIX, line 7: Odd ignore rules; these should never be committed.

BenjaminHalko:
So nothing needs changing

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,48 @@
+name: PR Build Compose
```
oSumAtrIX, line 1:
```suggestion
name: Build pull request
```

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,48 @@
...
+          # https://github.com/cycjimmy/semantic-release-action#private-packages
+          persist-credentials: false
+          fetch-depth: 0
+
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: gradle
```
oSumAtrIX, line 27: I don't think this is necessary, the runner already comes with JDK 17.

oSumAtrIX (changes requested):

Ushie:
> @Ushie Because only branches that have this workflow trigger this workflow, only compose branches will use it.

I see, apologies again

.github/workflows/pr-build.yml
oSumAtrIX: Also maybe adjust the filename too.

BenjaminHalko: Are you sure you don't want "Compose" in the title? I am trying to make the title unique from [the flutter pr build workflow](https://github.com/ReVanced/revanced-manager/blob/dev/.github/workflows/pr-build.yml)

BenjaminHalko: Plus the "PR Build" part is to also follow the same format as the flutter workflow

oSumAtrIX: Yes, the branch is after all a feature branch, and supposed to eventually reach main. The workflow itself acts without regard to any other for that reason. Besides, "Build pull request" is a better sentencea

.github/workflows/pr-build.yml
BenjaminHalko: They were rules from the old compose repo, seems like since .md and and .idea don't actually change the final build so there is no need to create a test.

Unless you are saying they should have never been commited even in the old repo, in which case I ask why?

.github/workflows/pr-build.yml
BenjaminHalko: Sounds good

.github/workflows/pr-build.yml
oSumAtrIX: This would assume these are the only paths that should be ignored. You can use a whitelist to only build PRs if pushed to selected paths.

oSumAtrIX: And yes, they should've also not been in the "old" repo.

BenjaminHalko: What would the whitelisted paths be?

oSumAtrIX: I'd say anything worthwhile to trigger a PR release. Source, for example? With a whitelist you'd also have to consider build files probably, which is why simply triggering on any kind of push is fine I suppose.

BenjaminHalko: I am assuming:
```
.github/workflows/pr-build.yml
app/**
assets/**
lib/**
package-lock.json
package.json
pubspec.yaml
```

`fastlane` might also need to be in there as well.

BenjaminHalko: I am assuming:
```
.github/workflows/pr-build.yml
app/**
gradle/**
build.graddle.kts
gradle.properties
settings.gradle.kts
```

.github/workflows/pr-build.yml
BenjaminHalko: The build is failing and saying that it requires Java 17

oSumAtrIX: Odd, the docs specify the runner comes with JDK 17: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#java

.github/workflows/pr-build.yml
oSumAtrIX: Probably, yeah

BenjaminHalko:
@oSumAtrIX Should the workflow be running `assembleRelease` or `assembleDebug`.

- On one hand `assembleDebug` makes sense as it is being used to debug the pr.
- On the other, `assembleRelease` makes sense as it will be the only way of getting compose builds without having to build it yourself, thus others can test the builds and submit issues.

BenjaminHalko:
It seems like the [pr workflow](https://github.com/ReVanced/revanced-manager/actions/runs/6067187077/job/16458559821?pr=1228) is already doing it's job. Showing that the minifying task is failing.

oSumAtrIX:
Good question, if PRs are supposed to be debugged, a debug release should be done, otherwise a regular release.

BenjaminHalko:
@oSumAtrIX I agree, but now see that this workflow failed on the minifying step, it makes me wondering if we should do both. As the minifying step would not have been run in debug mode.

What if there are other bugs in building that only appear in release builds.

BenjaminHalko:
Running both shouldn't be too harmful right? They could even run async.

oSumAtrIX:
I am missing the necessity of running release builds solely motivated by this. I'd say releases should be tested at a different stage such as dev.

BenjaminHalko:
But if the pr changes code that only affects release builds then only compiling debug builds will allow that bug to slip by.

.github/workflows/pr-build.yml
BenjaminHalko: As you can see from [this run](https://github.com/ReVanced/revanced-manager/actions/runs/6067130741/job/16458436540) it failed to find JDK 17.

BenjaminHalko:
Seems like the [workflow](https://github.com/ReVanced/revanced-manager/actions/runs/6067620495/job/16459526132) can't find the signing key.

oSumAtrIX:
PRs originating outside of this repository are not granted access to read it. You should open PR branches in this repository.

BenjaminHalko:
@oSumAtrIX Well considering that there will be some pr's from people outside the organization, should we remove the key signing part?

BenjaminHalko:
Is it possible to create "debug keys" that are for signing pr builds that can be given access?

BenjaminHalko:
Also why aren't they given access, if the pr is created by someone outside the organization doesn't the workflow have to be approved before it is run anyways?

oSumAtrIX:
I've encountered this issue too. There are ways to leak repository secrets according to GitHub docs, but that should be done with utmost care. The simplest solution is to not sign the releases, but then I think you'd have issues installing them.

BenjaminHalko:
Debug keys seem to be a good option. #1230 brings up a good point

oSumAtrIX:
I am still not entirely convinced about creating releases and debug builds. I'd say just debug is fine, after all it's just PRs, we can think about release builds in dev.

BenjaminHalko:
Sounds good. We still need to figure out an issue with the keys.

oSumAtrIX:
The default debug keystore can be used: https://developer.android.com/studio/publish/app-signing#debug-mode

BenjaminHalko:
Will this key be the same every workflow?

BenjaminHalko:
Wait it seems like it will be different. Is this the same as just not running the sign apk step.

BenjaminHalko:
If that is the case, will it run into the same issues #1231 tried to fix?

oSumAtrIX:
The debug keystore should be included in the Android SDK included with the runner.

BenjaminHalko:
@oSumAtrIX Do you know where it is located? `~/.android/debug.keystore`?

BenjaminHalko:
Is this different than not running the apk sign step?

BenjaminHalko:
@oSumAtrIX Should the Flutter pr workflow also use this key then instead of the secret key?

oSumAtrIX:
Debug builds can use that, yes.

BenjaminHalko:
@oSumAtrIX Seems like there is still a small bug in the key sign step of the workflow.

BenjaminHalko:
@oSumAtrIX Where did you see that the Android SDK in GitHub Actions comes with a debug keystore

Ushie:
iirc, it's automatically used when using ```.\gradlew assembleDebug```, it should be fine to remove the signing section

BenjaminHalko:
@Ushie But then we run into the same problems that #1230 addressed

BenjaminHalko:
@oSumAtrIX Due to the [performance impact](https://developer.android.com/jetpack/compose/performance#build-release) that compose apps have in debug mode, I wonder if it is actually worth building both a debug and release build for PRs.

oSumAtrIX:
The builds wouldn't act as production ready builds. They are intended for testing. Release builds should be fine considering, if you plan to debug you'd simply checkout the PR branch anyways

BenjaminHalko:
@oSumAtrIX Ok sounds good. But I will need a way of signing the builds.

BenjaminHalko:
@oSumAtrIX Should Flutter PRs also build release then?

BenjaminHalko:
The compose repo had test keys, maybe we could use those to sign pr builds.

validcube (approved):
🪐 The CI looks good to me!

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Sign APK
+        id: sign_apk
+        uses: ilharp/sign-android-release@v1
+        with:
+          releaseDir: build/app/outputs/apk/release
+          signingKey: ${{ secrets.SIGNING_KEYSTORE }}
+          keyStorePassword: ${{ secrets.SIGNING_KEYSTORE_PASSWORD }}
+          keyAlias: ${{ secrets.SIGNING_KEY_ALIAS }}
+          keyPassword: ${{ secrets.SIGNING_KEY_PASSWORD }}
```
validcube, line 45: Need to be remove, PR can't access them.

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,51 @@
...
+
+      - name: Upload build
+        uses: actions/upload-artifact@v3
+        with:
+          name: revanced-manager
+          path: ${{ steps.sign_apk.outputs.signedFile }}
```
validcube, line 51: Need to be remove, PR can't access them.

validcube (changes requested):

.github/workflows/pr-build.yml
BenjaminHalko: It needs to be signed with a key as release builds aren't signed by default. I suggested maybe debug keys.

validcube: > It needs to be signed with a key as release builds aren't signed by default. I suggested maybe debug keys.

I guess we could sign the pr build with debug key.

BenjaminHalko: I do not know who has access to add keys, but I do not.

BenjaminHalko: > I guess we could sign the pr build with debug key, if that's the only option.

I guess technically the other option is that the unsigned build is uploaded and testers will need to sign it themselves before testing.

BenjaminHalko:
@oSumAtrIX Are release builds for PRs not possible as they can't be signed with secret keys?

.github/workflows/pr-build.yml
BenjaminHalko: @validcube We could also only compile debug builds, but debug builds for compose have performance issues, reducing the usefulness of testing it.

oSumAtrIX:
I am not sure. You should be able to sign any APK.

BenjaminHalko:
@oSumAtrIX Where are these keys located?

BenjaminHalko:
It finally works!

validcube (approved):
🥞 CI looks good to me!

BenjaminHalko:
While it dies work, it would be nice if I did not have to load the Android SDK though.

BenjaminHalko:
In case anyone is curious, I also tried:

```
uses: ilharp/sign-android-release@v1
  with:
    releaseDir: app/build/outputs/apk/release
    signingKey: $(openssl base64 < debugkey.jks | tr -d '\n')
    keyAlias: androiddebugkey
    keyStorePassword: android
    keyPassword: android
```
[failed](https://github.com/ReVanced/revanced-manager/actions/runs/6176720802/job/16766287972#step:6:552) with `java.io.IOException: DerInputStream.getLength(): lengthTag=23, too big.`

```
uses: r0adkll/sign-android-release@v1
  with:
    releaseDirectory: app/build/outputs/apk/release
    signingKeyBase64: $(openssl base64 < debugkey.jks | tr -d '\n')
    alias: androiddebugkey
    keyStorePassword: android
    keyPassword: android
```
[failed](https://github.com/ReVanced/revanced-manager/actions/runs/6176068539/job/16764245452#step:6:547) with `java.io.IOException: Tag number over 30 is not supported`

BenjaminHalko:
Now it will sign the app using arguments passed to Gradle, instead of apksigner. This eliminates the need to download the Android SDK

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Generate keystore
+        run: keytool -genkey -v -keystore debugkey.jks -alias androiddebugkey -storepass android -keypass android -keyalg RSA -keysize 2048 -validity 10000 -dname "O=ReVanced"
+
+      - name: Build with Gradle
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew assembleRelease --no-daemon -Pandroid.injected.signing.store.file=$GITHUB_WORKSPACE/debugkey.jks -Pandroid.injected.signing.store.password=android -Pandroid.injected.signing.key.alias=androiddebugkey -Pandroid.injected.signing.key.password=android
+
+      - name: Rename file
+        run: mv app/build/outputs/apk/release/app-release.apk app-debug.apk
```
Ushie, line 41: Given that the build type is release, it doesn't make sense to rename the file to app-debug which suggests that it's a debug build

You can instead rename it to something like revanced-manager-${commit-hash}.apk, what do you think?

BenjaminHalko: Sounds good

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Generate keystore
+        run: keytool -genkey -v -keystore debugkey.jks -alias androiddebugkey -storepass android -keypass android -keyalg RSA -keysize 2048 -validity 10000 -dname "O=ReVanced"
+
+      - name: Build with Gradle
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew assembleRelease --no-daemon -Pandroid.injected.signing.store.file=$GITHUB_WORKSPACE/debugkey.jks -Pandroid.injected.signing.store.password=android -Pandroid.injected.signing.key.alias=androiddebugkey -Pandroid.injected.signing.key.password=android
+
+      - name: Rename file
```
Ushie, line 40:
```suggestion
      - name: Add version to APK
```

BenjaminHalko: It is not the version though, so maybe `Add hash to APK`?

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,47 @@
...
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Make sure the release step uses its own credentials:
+          # https://github.com/cycjimmy/semantic-release-action#private-packages
+          persist-credentials: false
+          fetch-depth: 0
```
Ushie, line 24: What are these for?

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Set up JDK
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
```
Ushie, line 30: Are you sure JDK needs to be set up?

.github/workflows/pr-build.yml
BenjaminHalko: I do not know but it is on the other workflow.
[pr-build](https://github.com/ReVanced/revanced-manager/blob/main/.github/workflows/pr-build.yml#L16)

I think it might be safe to remove.

.github/workflows/pr-build.yml
BenjaminHalko: Yes... but I can double check.

.github/workflows/pr-build.yml
Ushie: Sure

.github/workflows/pr-build.yml
BenjaminHalko: https://github.com/ReVanced/revanced-manager/actions/runs/6202388269/job/16841036450#step:4:35

I guess it needs to be setup.

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,45 @@
+name: Build pull request
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/pr-build.yml"
+      - "app/**"
+      - "gradle/**"
+      - "build.graddle.kts"
```
Ushie, line 9:
```suggestion
      - "build.gradle.kts"
```

Ushie:
Maybe [cache](https://github.com/ReVanced/revanced-integrations/blob/main/.github/workflows/release.yml#L26) can be added here too?

BenjaminHalko:
sure thing, huh it seems that that workflow also disables credentials... maybe it is important...

Ushie:
The credentials thing seems to be semantic releases action related, we don't use that

BenjaminHalko:
Are the cache properties correct? I think we can omit node related things like `package-lock.json` and `node_modules`?

```
       path: |
            ${{ runner.home }}/.gradle/caches
            ${{ runner.home }}/.gradle/wrapper
            .gradle
            node_modules
          key: ${{ runner.os }}-gradle-npm-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties', 'package-lock.json') }}
```

Ushie:
Yeah, should remove them, we only need gradle related things

BenjaminHalko:
Like this?
```
path: |
            ${{ runner.home }}/.gradle/caches
            ${{ runner.home }}/.gradle/wrapper
            .gradle
          key: ${{ runner.os }}-gradle-npm-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
```

BenjaminHalko:
wait gradle-wrapper.properties doesn't apply either

BenjaminHalko:
should it also be hashing other gradle files like `settings.gradle.kts` and `build.gradle.kts`?

BenjaminHalko:
Like this:
```
path: |
            ${{ runner.home }}/.gradle/caches
            ${{ runner.home }}/.gradle/wrapper
            .gradle
          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '*.gradle*') }}
```

Ushie:
gradle-wrapper.properties should be kept, I'm not sure if other files are necessary

BenjaminHalko:
sure, I just found that we do have a gradle-wrapper file. My bad.

BenjaminHalko:
@Ushie I think the cache is not working properly, this workflow should have used the cache, should it be loading the app/.gradle folder as well?

BenjaminHalko:
@oSumAtrIX Is this ready for merge now?

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,51 @@
...
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/pr-build.yml"
+      - "app/**"
+      - "gradle/**"
+      - "build.gradle.kts"
+      - "gradle.properties"
+      - "settings.gradle.kts"
```
oSumAtrIX, line 11:
```suggestion
      - "*.properties"
      - ".kts"
```

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: ./gradlew assembleRelease --no-daemon -Pandroid.injected.signing.store.file=$GITHUB_WORKSPACE/debugkey.jks -Pandroid.injected.signing.store.password=android -Pandroid.injected.signing.key.alias=androiddebugkey -Pandroid.injected.signing.key.password=android
+
+      - name: Add hash to APK
+        run: mv app/build/outputs/apk/release/app-release.apk revanced-manager-${{ env.COMMIT_HASH }}.apk
+
+      - name: Upload build
+        uses: actions/upload-artifact@v3
+        with:
+          name: revanced-manager
+          path: revanced-manager-${{ env.COMMIT_HASH }}.apk
```
oSumAtrIX, line 51:
```suggestion
      - name: Set up cache
        uses: actions/cache@v3
        with:
          path: |
            ${{ runner.home }}/.gradle/caches
            ${{ runner.home }}/.gradle/wrapper
            .gradle
          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}

      - name: Set up Java
        run: echo "JAVA_HOME=$JAVA_HOME_17_X64" >> $GITHUB_ENV

      - name: Generate keystore
        run: keytool -genkey -v -keystore debugkey.jks -alias androiddebugkey -storepass android -keypass android -keyalg RSA -keysize 2048 -validity 10000 -dname "O=ReVanced"

      - name: Build with Gradle
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: ./gradlew assembleRelease --no-daemon -Pandroid.injected.signing.store.file=$GITHUB_WORKSPACE/debugkey.jks -Pandroid.injected.signing.store.password=android -Pandroid.injected.signing.key.alias=androiddebugkey -Pandroid.injected.signing.key.password=android

      - name: Set env
        run: echo "COMMIT_HASH=$(git rev-parse --short HEAD)" >> $GITHUB_ENV

      - name: Add hash to APK
        run: mv app/build/outputs/apk/release/app-release.apk revanced-manager-${{ env.COMMIT_HASH }}.apk

      - name: Upload build
        uses: actions/upload-artifact@v3
        with:
          name: revanced-manager
          path: revanced-manager-${{ env.COMMIT_HASH }}.apk
```

oSumAtrIX: Moved set env step below

.github/workflows/pr-build.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Generate keystore
+        run: keytool -genkey -v -keystore debugkey.jks -alias androiddebugkey -storepass android -keypass android -keyalg RSA -keysize 2048 -validity 10000 -dname "O=ReVanced"
+
+      - name: Build with Gradle
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: ./gradlew assembleRelease --no-daemon -Pandroid.injected.signing.store.file=$GITHUB_WORKSPACE/debugkey.jks -Pandroid.injected.signing.store.password=android -Pandroid.injected.signing.key.alias=androiddebugkey -Pandroid.injected.signing.key.password=android
+
+      - name: Add hash to APK
+        run: mv app/build/outputs/apk/release/app-release.apk revanced-manager-${{ env.COMMIT_HASH }}.apk
```
oSumAtrIX, line 45: These steps should ideally be included in the build system. Depending on the type of build, it could append the hash to the file name. 

Here is an example: https://github.com/iBotPeaches/Apktool/blob/master/build.gradle.kts#L50

Let me know what you think about this.

oSumAtrIX (changes requested):

.github/workflows/pr-build.yml
BenjaminHalko: Sounds good. It would clean things up a bit.

BenjaminHalko: By the way, what do you think about having pr builds having minify disabled.

BenjaminHalko: Just to clarify, would I be adding a new assemble task to gradle such as `assembleReleaseTesting`?

oSumAtrIX: I am not sure. I think it is possible to make a task depend on another. So you can make the assembleRelease task depend on the renaming task maybe?

BenjaminHalko: Oh I get now. Cool.

BenjaminHalko: Should it only rename when given a keystore? Currently it outputs as `app-release-unsigned` with no key and `app-release` with a key?

BenjaminHalko: So when given no key it should output the same thing but with `-unsigned` at the end?

Ushie: yep!

oSumAtrIX: I think it doesn't matter too much. What I sent is just for reference. You can append `-unsigned` at the end. I think @validcube already renames the file to begin with if its a debug build.

BenjaminHalko: Should it just append the git hash at the end?

BenjaminHalko: > I think that is possible. A release task can sign a release apk, but the hash should not be added. The release task only runs when a release is made, which happens in main. The release file name is appended with the tag assigned to HEAD.

When would a hash be added to the file name then? If the release task does not add it.

BenjaminHalko: > When you locally debug, you should probably rely on a debug release task so that you can actually debug

This is true for most cases, except when testing performance

BenjaminHalko: > You could configure only minify if for example proguard was in the gradle start properties

```
buildTypes {
        release {
            if ("proguard" in gradle.startParameter.taskNames) {
                isMinifyEnabled = true
                isShrinkResources = true
                proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
            }
        }
    }
```

oSumAtrIX: In this case two tasks can be made or the task can be made configurable to build releases or debug releases.

oSumAtrIX: > When would a hash be added to the file name then? If the release task does not add it.

A separate task made for PR builds I suppose.

BenjaminHalko: > A separate task made for PR builds I suppose.

With having a configurable assembleRelease task, it could normally always append the hash, but when adding the extra param, it could minify AND rename with tag instead of hash.

What to call that param though, `assembleRelease debug` doesn't seem right...

oSumAtrIX: How about making `assembleRelease` work as it does right now, and when using the parameter `debug`, append the has, sign with a debug key as well as don't minify?

oSumAtrIX: I am not sure if I understood correctly. The task `assembleRelease` should work as it works right now, andwhen using the parameter `debug`, the hash can be appended, sig with a debug key as well as don't minify

BenjaminHalko: Yesn

BenjaminHalko: I was meaning if we had a task called `installRelease` to install the release build to a device, it should run the `assembleRelease` task with the `debug` param since otherwise it would be trying to install an unsigned apk.

BenjaminHalko: @oSumAtrIX Ok the changes have been done:

- `assembleRelease` will output `revanced-manager-release-unsigned.apk`
- `assembleRelease -Pdebug` will output `revanced-manager-{git hash}-release.apk` (this build will also have minify disabled)
- `installRelease` will fail as you cannot install an unsigned apk
- `installRelease -Pdebug` will install the signed release build
- `assembleDebug` will output `revanced-manager-debug,apk`

Also let me know what you think about the prefix change from `app` to `revanced-manager`.

oSumAtrIX: @Ushie Is the `assembleRelease` fine? How would the tag be appended when pushing a tag on main?

BenjaminHalko: It could either be renamed manually or I could add a param like `-PaddTag` to rename it

BenjaminHalko: it would probably be used as `-Ptag=${GITHUB_REF#refs/*/}` if you want to use a param

Ushie: > @Ushie Is the `assembleRelease` fine? How would the tag be appended when pushing a tag on main?

```yaml
- name: Set env
  run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
```

This is what's currently done in https://github.com/ReVanced/revanced-manager/blob/main/.github/workflows/release-build.yml

BenjaminHalko: Updated params:

- `assembleRelease` will output `app-release-unsigned.apk`
- `assembleRelease -Psign` will output `app-release.apk`
- `assembleRelease -PnoProguard` will disable proguard
- `installRelease` will fail as you cannot install an unsigned apk
- `installRelease -Psign` will install the signed release build
- `installRelease -PnoProguard -Psign` will install the signed release build without proguard

oSumAtrIX (approved):

BenjaminHalko:
Manager now has a different appID and name so that you can install both a pr build/debug build and the release version (whenever that gets released) at the same time.

BenjaminHalko:
@oSumAtrIX Is this ready to merge?

oSumAtrIX:
You got the greens from me ✅
image

BenjaminHalko:
@Ushie Anything else to add?

app/build.gradle.kts
```diff
@@ -21,15 +21,20 @@ android {
...
-            isMinifyEnabled = true
-            isShrinkResources = true
-            proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
+            if (!project.hasProperty("noProguard")) {
+                isMinifyEnabled = true
+                isShrinkResources = true
+                proguardFiles(getDefaultProguardFile("proguard-android-optimize.txt"), "proguard-rules.pro")
+            }
+
+            if (project.hasProperty("sign")) {
```
Ushie, line 35: signAsDebug would make more sense

Ushie (changes requested):

BenjaminHalko:
@Ushie done

Ushie (approved):

PR #1229, opened 2023-09-03
ci(pre-release-builds): add workflow for pre releases

BenjaminHalko:
- It would be nice to have compose builds uploaded under the `pre-release` tag instead of just from GitHub actions. This would make the process of downloading a compose build easier.
- The app itself could also be updated to download only `pre-release` tags as it currently will only download the flutter manager.
- The workflow in this pr uses a currently non-existent `compose-main` branch to upload releases.

**Disclaimers:**
- I have no idea if the workflow works, I just merged bits of the flutter release workflow and the compose build workflow
- I do not know how/if/when pre-releases for compose will work/are planned, the actual system for them will probably be very different than the one proposed here.

### DIFF
--- a/.github/workflows/pre-release-build.yml
+++ b/.github/workflows/pre-release-build.yml
@@ -1,52 +1,49 @@
-name: "Release Build"
+name: "Pre Release Build"
 
 on:
   push:
     tags:
       - "v*"
     branches:
-      - main
-
+      - compose-main
+      
 jobs:
-  release:
+  build:
+    name: Build
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
       - name: Set env
         run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
-      - name: Set up JDK 11
+
+      - name: Set up JDK 17
         uses: actions/setup-java@v3
         with:
-          java-version: "11"
-          distribution: "zulu"
-      - uses: subosito/flutter-action@v2
-        with:
-          channel: "stable"
-      - name: Set up Flutter
-        run: flutter pub get
-      - name: Generate files with Builder
-        run: flutter packages pub run build_runner build --delete-conflicting-outputs
-      - name: Build with Flutter
+          java-version: '17'
+          distribution: 'temurin'
+          cache: gradle
+
+      - name: Build with Gradle
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          SIGNING_KEY_ALIAS: ${{ secrets.SIGNING_KEY_ALIAS }}
-          SIGNING_KEY_PASSWORD: ${{ secrets.SIGNING_KEY_PASSWORD }}
-          SIGNING_STORE_PASSWORD: ${{ secrets.SIGNING_KEYSTORE_PASSWORD }}
-        run: flutter build apk
+        run: ./gradlew assembleRelease --no-daemon
+
       - name: Sign APK
         id: sign_apk
         uses: ilharp/sign-android-release@v1
         with:
-          releaseDir: build/app/outputs/apk/release
+          releaseDir: ./app/build/outputs/apk/release/
           signingKey: ${{ secrets.SIGNING_KEYSTORE }}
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
Any reason why the existing build workflow can be used?

BenjaminHalko:
@oSumAtrIX The existing workflow for flutter or compose?

oSumAtrIX:
I am failing to see a difference in the workflow for both. They should be the same shouldn't they?

BenjaminHalko:
This one builds compose and uploads to github releases, the other uploads to github actions as a debug build for pull requests.

BenjaminHalko:
Like the difference between the pr-build and release-build workflows

oSumAtrIX:
The repository should have two workflows. One for PRs (if necessary) and one for creating releases. The current branch in which the workflow was triggered should determine what kind of release is created. This means the release workflow for Compose and Flutter should be identical. When triggered from `dev`, generate pre-releases when triggered from `main` generate releases, assuming the building process is identical of course.

BenjaminHalko:
The build process is different though right? Plus compose does not use the dev branch (unless I am mistaken).

oSumAtrIX:
From what I am reading, both simply require the assembleRelease task to be run. The workflow should, therefor be identical. Compose is currently being worked on in a feature branch. If you want to, you can also let the workflow run on feature branches, though the PR build workflow already covers feature branches and the release workflows should only run on the dev and main branches.

BenjaminHalko:
Doesn't pr build run flutter build apk?

oSumAtrIX:
Right. Does running the gradle task not work, though? Also, is it intended that this PR target branch is `dev`? The workflow is intended for Compose, therefore should be introduced with the merge of the Compose dev branch

BenjaminHalko:
Oh the other compose work flow is for prs and uploads a debug build to actions. This is for a branch (right now it compose-main), and uploads to github releases.

PalmDevs:
Currently, there are no plans to actually officially release Compose Manager. It's not that stable to actually do a release yet. Although when we're ready to do (for example) a beta release, this would be discussed inside another issue (which is the `ci: init` one).

BenjaminHalko:
Workflows only work on main (right???) (thus I am push to dev in order to be merged with main). But it targets `compose-main` (it uses on push with branch `compose-main`)

PalmDevs:
We would also ideally have an automatic version bumping workflow which uses CalVer, so mistakes in bumping will no longer be made.

oSumAtrIX:
The workflow should not be pushed to main nor dev and instead to the compose branch and introduced with the compose and introduced with the merge, but then again, we only need one build workflow.

BenjaminHalko:
@oSumAtrIX Do workflows work even if the .yml is not on the default branch?

@PalmDevs Sounds good

PalmDevs:
> **oSumAtrIX**, Do workflows work even if the .yml is not on the default branch? (This is separate from what the workflow targets)

As long as they can be triggered somehow, yes.

oSumAtrIX:
No, workflows have to be in the default branch to run.

BenjaminHalko:
@PalmDevs My bad, thanks for clarification

BenjaminHalko:
So which branch should I switch this pr to. `compose-dev`?

BenjaminHalko:
@oSumAtrIX Thanks for clarification

BenjaminHalko:
So this should get merged with dev then main right?? Otherwise it will not run....

oSumAtrIX:
As previously explained, it should not run until the compose is merged. Until then, the only workflow that should run for feature branches is the PR build workflow.

BenjaminHalko:
Oh I see what you mean. This pr is for potentially creating pre releases BEFORE compose is merged with main so that it is easier to download compose builds. Not as a release workflow for after compose gets merged with main.

BenjaminHalko:
If pre releases are 100% not planned for compose then I can close this pr.

BenjaminHalko:
@oSumAtrIX Sounds good!

oSumAtrIX:
You can simply reopen the PR to the compose branch target. Simply make sure there is only one release workflow in that branch

PR #1243, opened 2023-09-05
feat: root installation

CnC-Robert:


app/src/main/java/app/revanced/manager/domain/installer/RootInstaller.kt
```diff
@@ -0,0 +1,132 @@
...
+    companion object {
+        const val modulesPath = "/data/adb/modules"
+
+        /*val isDeviceRooted =
+            try {
+                Runtime.getRuntime().exec("su --version")
+                true
+            } catch (_: IOException) {
+                false
+            }*/
```
Axelen123, line 128: Why is this left here?

app/src/main/java/app/revanced/manager/ui/component/SegmentedButton.kt
```diff
@@ -28,41 +30,54 @@ fun RowScope.SegmentedButton(
...
                 )
+                .weight(1f)
+                .padding(vertical = 20.dp)
+        ) {
+            when (icon) {
+                is ImageVector -> {
+                    Icon(
+                        imageVector = icon,
+                        contentDescription = iconDescription,
+                        //tint = MaterialTheme.colorScheme.primary,
```
Axelen123, line 61: Why was this changed? Why is the comment left here?

app/src/main/java/app/revanced/manager/ui/screen/InstallerScreen.kt
```diff
@@ -121,6 +130,50 @@ fun InstallerScreen(
...
+        confirmButton = {
+            Button(onClick = {
+                onConfirm(selectedInstallType)
+                onDismiss()
+            }
+            ) {
+                Text("Install")
+            }
+        },
+        title = { Text("Select installation type") },
```
Axelen123, line 156: Make the text translatable

app/src/main/java/app/revanced/manager/ui/viewmodel/AppSelectorViewModel.kt
```diff
@@ -11,22 +11,19 @@ import java.nio.file.Files
...
     private val app: Application,
-    pm: PM
+    private val pm: PM
 ) : ViewModel() {
-    private val packageManager = app.packageManager
-
     val appList = pm.appList
 
-    fun loadLabel(app: PackageInfo?) = (app?.applicationInfo?.loadLabel(packageManager) ?: "Not installed").toString()
+    fun loadLabel(app: PackageInfo?) = with(pm) { app?.label() ?: "Not installed" }
```
Axelen123, line 18: The fallback message should be translatable

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -192,20 +213,59 @@ class InstallerViewModel(input: Destination.Installer) : ViewModel(), KoinCompon
...
     }
 
-    fun installOrOpen() = viewModelScope.launch {
-        installedPackageName?.let {
+    fun install(installType: InstallType) = viewModelScope.launch {
+        /*installedPackageName?.let {
             pm.launch(it)
             return@launch
-        }
+        }*/
```
Axelen123, line 220: Remove the comment

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/domain/installer/RootInstaller.kt
CnC-Robert: I forgot to remove it

app/src/main/java/app/revanced/manager/ui/component/SegmentedButton.kt
CnC-Robert: The content color should be gray when it is disabled. Im providing the color with CompositionLocalProvider now

Axelen123: Fair enough. The comment should be removed though

app/src/main/java/app/revanced/manager/domain/installer/RootInstaller.kt
```diff
@@ -0,0 +1,134 @@
...
+
+                                while (inputStream.read(buffer).also { bytesRead = it } != -1) {
+                                    val content = String(buffer, 0, bytesRead)
+                                        .replace("__PKG_NAME__", packageName)
+                                        .replace("__VERSION__", version)
+                                        .replace("__LABEL__", label)
+                                        .toByteArray()
+
+                                    outputStream.write(content)
+                                }
```
Axelen123, line 92: This will cause problems in the future if the file is changed in such a way that `__PKG_NAME__` ends up at a 1024-byte boundary, because it will get cut off and not get replaced by the expected value.

Can be solved by reading the entire file instead of streaming (or by using something thats made for streams). Memory needs to be traded for correctness in this situation

Axelen123 (changes requested):

Axelen123 (approved):

PR #1247, opened 2023-09-07
chore: update dependencies

PatrykMis:
Replaces https://github.com/ReVanced/revanced-manager-compose/pull/90

Not tested yet, will do it tomorrow.


gradle/libs.versions.toml
```diff
@@ -1,31 +1,31 @@
 [versions]
-ktx = "1.10.1"
-viewmodel-lifecycle = "2.6.1"
+ktx = "1.12.0"
```
validcube, line 2: Bumping ktx version to 1.12.0 will fail because it was compiled against 34, our application was compiled against 33.

validcube (changes requested):

validcube (changes requested):
Upon compiling I see error related to `androidx.emoji2:emoji2-views-helper:1.4.0` compiled for Android 14, not Android 13.

PatrykMis:
Ok, fixed and now tested.

app/build.gradle.kts
```diff
@@ -4,7 +4,7 @@ plugins {
     alias(libs.plugins.devtools)
     alias(libs.plugins.about.libraries)
     id("kotlin-parcelize")
-    kotlin("plugin.serialization") version "1.9.0"
+    kotlin("plugin.serialization") version "1.9.10"
```
Axelen123, line 7: The previous `kotlinx.serialization` bump caused a crash at runtime. Can you get past the auto updates dialog with both options selected without crashing?

PatrykMis: Yes that works, at least for me on Pixel 4 device with Android 13.

validcube (approved):
🌟 +1 Approved, Tested on Debug variant of ReVanced Manager Compose

Axelen123 (approved):

Axelen123:
Apologies for the late review

PR #1271, opened 2023-09-17
fix(ui): make entire patches view button selectable

BenjaminHalko:
The button to view the available patches inside a bundle has been update to allow the entire row to be clickable similar to the rest of the interactable elements.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/component/bundle/BaseBundleDialog.kt
+++ b/app/src/main/java/app/revanced/manager/ui/component/bundle/BaseBundleDialog.kt
@@ -159,15 +159,12 @@ fun BaseBundleDialog(
         )
     }
 
-
     BundleListItem(
         headlineText = stringResource(R.string.patches),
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/manager/ui/component/bundle/BaseBundleDialog.kt
```diff
@@ -159,22 +159,23 @@ fun BaseBundleDialog(
...
-                            stringResource(R.string.patches)
-                        )
-                    }
-                }
+
+    BundleListItem(
+        headlineText = stringResource(R.string.patches),
+        supportingText = if (patchCount == 0) stringResource(R.string.no_patches)
+        else stringResource(R.string.patches_available, patchCount),
+        modifier = Modifier.clickable {
```
Axelen123, line 167: Use the `enabled` parameter of `clickable` to disable clicking if `patchCount == 0`

Axelen123 (changes requested):

Axelen123 (approved):

PR #1274, opened 2023-09-17
feat: warn on changing default patches

BenjaminHalko:
Now the manager will warn if you are changing default patches.


BenjaminHalko:
Oh wait....... I just saw a pr just like this was opened just recently...

whoops

PR #1277, opened 2023-09-18
refactor: update progress onBackClick function

BenjaminHalko:
The `onBackClick` argument was not specified for `UpdateProgressScreen`.


Axelen123:
Not a bugfix, but it makes sense to do it to be consistent with the other function calls in the file

Issue #1260, opened 2023-09-10
Follow the system theme


xmha97:
### Feature description

Now I can switch between light theme and dark theme.
But I can't follow the system theme.
Please add this feature.
Thanks 🙏

### Motivation

N/A

### Acknowledgements

- [X] This request is not a duplicate of an existing issue.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The issue is solely related to the ReVanced Manager

KobeW50:
There's already MaterialYou on supported devices

xmha97:
I need this option:

![Screenshot_2023-09-12-21-41-41-177_app.revanced.android.youtube.jpg](https://github.com/ReVanced/revanced-manager/assets/23485114/e289dbb2-8fd5-483d-9344-edab966a754f)

KobeW50:
> I need this option:
> 
> ![Screenshot_2023-09-12-21-41-41-177_app.revanced.android.youtube.jpg](https://github.com/ReVanced/revanced-manager/assets/23485114/e289dbb2-8fd5-483d-9344-edab966a754f)
> 
> 

Ah, sorry, I misunderstood what following the system theme meant.

PR #1279, opened 2023-09-20
feat: follow the system theme

BenjaminHalko:
fixes: #1260

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -178,11 +178,11 @@
         "exportSectionTitle": "Import & export",
         "logsSectionTitle": "Logs",
 
-        "themeModeLabel": "App Theme",
+        "themeModeLabel": "App theme",
         "themeModeHint": "Select the theme to use",
-        "systemThemeLabel": "System default",
-        "lightThemeLabel": "Light mode",
-        "darkThemeLabel": "Dark mode",
+        "systemThemeLabel": "System",
+        "lightThemeLabel": "Light",
+        "darkThemeLabel": "Dark",
 
         "dynamicThemeLabel": "Material You",
         "dynamicThemeHint": "Enjoy an experience closer to your device",
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -178,8 +178,11 @@
         "exportSectionTitle": "Import & export",
         "logsSectionTitle": "Logs",
 
+        "themeModeLabel": "App Theme",
+        "themeModeHint": "Select the theme to use",
+        "systemThemeLabel": "System default",
+        "lightThemeLabel": "Light mode",
         "darkThemeLabel": "Dark mode",
-        "darkThemeHint": "Welcome to the dark side",
 
         "dynamicThemeLabel": "Material You",
         "dynamicThemeHint": "Enjoy an experience closer to your device",
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -179,7 +179,6 @@
         "logsSectionTitle": "Logs",
 
         "themeModeLabel": "App theme",
-        "themeModeHint": "Select the theme to use",
         "systemThemeLabel": "System",
         "lightThemeLabel": "Light",
         "darkThemeLabel": "Dark",
--- a/lib/services/manager_api.dart
+++ b/lib/services/manager_api.dart
@@ -197,12 +197,12 @@ class ManagerAPI {
     await _prefs.setBool('useDynamicTheme', value);
   }
 
-  bool getUseDarkTheme() {
-    return _prefs.getBool('useDarkTheme') ?? false;
+  int getThemeMode() {
+    return _prefs.getInt('themeMode') ?? 2;
   }
 
-  Future<void> setUseDarkTheme(bool value) async {
-    await _prefs.setBool('useDarkTheme', value);
+  Future<void> setThemeMode(int value) async {
+    await _prefs.setInt('themeMode', value);
   }
 
   bool areUniversalPatchesEnabled() {
--- a/lib/ui/theme/dynamic_theme_builder.dart
+++ b/lib/ui/theme/dynamic_theme_builder.dart
@@ -1,12 +1,16 @@
+import 'dart:ui';
 import 'package:dynamic_color/dynamic_color.dart';
 import 'package:dynamic_themes/dynamic_themes.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:google_fonts/google_fonts.dart';
+import 'package:revanced_manager/app/app.locator.dart';
 import 'package:revanced_manager/app/app.router.dart';
+import 'package:revanced_manager/services/manager_api.dart';
 import 'package:revanced_manager/theme.dart';
 import 'package:stacked_services/stacked_services.dart';
 
-class DynamicThemeBuilder extends StatelessWidget {
+class DynamicThemeBuilder extends StatefulWidget {
   const DynamicThemeBuilder({
     Key? key,
     required this.title,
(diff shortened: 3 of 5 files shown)


BenjaminHalko:
https://github.com/ReVanced/revanced-manager/assets/73490201/b81bb9aa-ebb4-46bd-81b1-7e84275f7a1f

oSumAtrIX:
Wondering if this is following correct M3 specs.

PS: Banger wallpaper

Ushie:
Instead of a DropDown button, you can use a FilledTonalButton and a popup, it looks much better

BenjaminHalko:
@Ushie I am assuming a popup of radio buttons?

Ushie:
yep!

BenjaminHalko:
@Ushie Something like this:
Screenshot_20230920-111849_ReVanced Manager Debug

Ushie:
It needs to have an apply and cancel button, and you could add an icon to the popup title maybe

This is how the old revanced manager codebase had it 
SmartSelect_20230920_212207_Hyperion

BenjaminHalko:
@Ushie Does it really need a cancel button though? Who is going to want to cancel their theme change...

Ushie:
I believe according to m3 guidelines the only time you provide the user with a single action is when it's an acknowledgement popup, not an action

Ushie:
It's a material design icon, I believe that specific one is Icon.palette, you can use it like other icons are used across the app

PalmDevs:
Yep, exactly as what Ushie said. Make sure the buttons are all text buttons though. No need for a filled button.

Ushie:
@PalmDevs Should this dialog include a title icon? no other dialog in the app does, it may be better to not include one just for the sake of consistency at the moment

PalmDevs:
> @PalmDevs Should this dialog include a title icon? no other dialog in the app does, it may be better to not include one just for the sake of consistency at the moment 

Technically, all dialogs should have an icon; but if no dialogs have it, don't include it for this one specific dialog.

Ushie:
Generally the current popups in the app aren't consistent in any way so we could have one PR refactoring all the popups, I could also probably pair that with a settings refactor because settings are currently in a horrible state

assets/i18n/en_US.json
PalmDevs, line 1: All dialogs in M3 use `Sentence case`, so change this to `App theme`.

assets/i18n/en_US.json
```diff
@@ -178,8 +178,11 @@
         "exportSectionTitle": "Import & export",
         "logsSectionTitle": "Logs",
 
+        "themeModeLabel": "App Theme",
+        "themeModeHint": "Select the theme to use",
```
PalmDevs, line 182: This is not required. Don't include any supporting text at all.

assets/i18n/en_US.json
```diff
@@ -178,8 +178,11 @@
         "exportSectionTitle": "Import & export",
         "logsSectionTitle": "Logs",
 
+        "themeModeLabel": "App Theme",
```
PalmDevs, line 181: All dialog headlines in M3 use `Sentence case`, so change this to `App theme`.

PalmDevs (changes requested):
The design should ideally look exactly like what Ushie's sent in the screenshot. Except only use text buttons, filled button isn't necessary. Also needs a few string changes.

BenjaminHalko:
Why not filled buttons, the other dialogs have them.

PalmDevs:
> Why not filled buttons, the other dialogs have them.

We don't need to point the user to click on one specific option as clicking on both does no harm.

BenjaminHalko:
how's this:
Screenshot_20230920-114831_ReVanced Manager Debug

BenjaminHalko:
also, pressing back on the navbar acts like the user has pressed ok, because chances are they would rather keep the changed theme than revert back.

BenjaminHalko:
Honestly I think that it should have a subtitle as everything else has one, or is it just me....

BenjaminHalko:
Also, I think the ok button should be filled for consistency but again, it might be just me...

Ushie:
> also, pressing back on the navbar acts like the user has pressed ok, because chances are they would rather keep the changed theme than revert back.

that's unexpected behavior, we shouldn't stray away from what's the norm

the only time changes are applied is with the confirm button, anything else is a dismissal

BenjaminHalko:
@Ushie fair, though I think when `DynamicTheme` is changed it is saved automatically. (separate from the individual material you and light/dark setting) so I will have to find out how to stop it from auto saving...

Ushie:
I don't think I understand

BenjaminHalko:
I disabled the line that saves the theme:
```
Future<void> setThemeMode(BuildContext context, int value) async {
    //await _managerAPI.setThemeMode(value);
    final bool isDynamicTheme = DynamicTheme.of(context)!.themeId.isEven;
    await DynamicTheme.of(context)!.setTheme(value * 2 + (isDynamicTheme ? 0 : 1));
    final bool isLight = value != 2 && (value == 1 || DynamicTheme.of(context)!.theme.brightness == Brightness.light);
    SystemChrome.setSystemUIOverlayStyle(
      SystemUiOverlayStyle(
        systemNavigationBarIconBrightness:
        isLight ? Brightness.dark : Brightness.light,
      ),
    );
    notifyListeners();
  }
  ```
  And you can see that the button displaying the current theme doesn't change from System, yet it still remembers the theme.
https://github.com/ReVanced/revanced-manager/assets/73490201/1dab0d6a-bdf0-42a6-b3e6-701b123b99af

BenjaminHalko:
@Ushie If I tell it to revert the theme on back button press, will it still matter that technically if you exit out of the app in the theme dialog it will retain your new theme despite not saving?

Ushie:
Are you doing that to be able to live-preview? you should make it so that it applies it on confirm, no need to apply it when a radio button is clicked

BenjaminHalko:
Ok then, sure thing!

BenjaminHalko:
@Ushie done.

I still think the OK button should be filled.

Ushie:
Sure, you can do that to make it consistent with other popups for the time being

Ushie:
Could you remove the weird padding around the radio button list? I hate how android does that, you can remove it the same way I removed it in the [installer type dialog](https://github.com/ReVanced/revanced-manager/pull/1062/files#diff-6ef035c5b3abb583a047adaa0263fce17ed343f3a871d60aab2beac59f076ef1R171)

BenjaminHalko:
@Ushie is this better?
Screenshot_20230920-144510_ReVanced Manager Debug

Ushie:
Looks better to me, @PalmDevs what do you think?

PalmDevs:
> @Ushie is this better?
> Screenshot_20230920-144510_ReVanced Manager Debug
> 

Can you decrease the gap between the radio button and the text to be equal to the padding of the dialog?

BenjaminHalko:
@PalmDevs do you know how to decrease that padding?

BenjaminHalko:
I searched it up and apparently there is no way to decrease that padding....

BenjaminHalko:
We can add padding to the others so it balances out.....

Ushie:
I think it'll be fine like this, the install type dialog also looks like this

BenjaminHalko:
I do fine it odd that the space on the left of the radio button is slightly less that to the right between the text...

BenjaminHalko:
This looks better to me personally:
Screenshot_20230920-163303_ReVanced Manager Debug

Ushie:
If you go with that instead, you should also update the padding on the installer one

BenjaminHalko:
I agree, if we do that should I open a new PR or just include it in this

Ushie:
Hm, I just remembered that installer dialog includes a title icon actually

Ushie:
Try removing the padding and adding the title icon we mentioned earlier in this thread, it might look better then

BenjaminHalko:
`Icons.palette` gives this icon which is different than the one shown in the picture you sent, but it still works.

Screenshot_20230920-163946_ReVanced Manager Debug

Ushie:
Ah no, I meant remove the extra padding you added in your last preview

BenjaminHalko:
Screenshot_20230920-164453_ReVanced Manager Debug

Though I personally like this one better as the padding between the edges and text are the same:
Screenshot_20230920-164658_ReVanced Manager Debug

BenjaminHalko:
There is also this problem we have to take into account:
Screenshot_20230920-164846_ReVanced Manager Debug

Ushie:
Alright, you can update the installer one aswell and push your changes

BenjaminHalko:
Sounds good

Ushie:
> There is also this problem we have to take into account: Screenshot_20230920-164846_ReVanced Manager Debug

Hm, I think radiolists in dialogs are meant to be scrollable, let me show you an example

I'm not sure how to do that in Flutter though

BenjaminHalko:
I think using `SingleChildScrollView`

Ushie:
unnamed (4)

BenjaminHalko:
fixed:
Screenshot_20230920-165221_ReVanced Manager Debug

Ushie:
Looks good to me, since that change should also be done in installer type's dialog you should push the changes relevant to theme picker only and a followup commit in the dev branch for installer type

Ushie (approved):

Issue #1249, opened 2023-09-07
fix: Updates popup shows incorrect names


oSumAtrIX:
This should say "ReVanced Manager" and "ReVanced Patches":

![image](https://github.com/ReVanced/revanced-manager/assets/13122796/2d2d76cf-c14e-45b5-8767-961042b8d278)

Axelen123:
Can be done by editing `app/src/main/res/values/strings.xml`

Ushie:
Shouldnt the patches one say Patch bundles instead of ReVanced Patches?

PR #1283, opened 2023-09-20
fix: Updates popup shows incorrect names

BenjaminHalko:
fixes: #1249

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -26,8 +26,8 @@
     
     <string name="auto_updates_dialog_title">Select updates to receive</string>
     <string name="auto_updates_dialog_description">Periodically connect to update providers to check for updates.</string>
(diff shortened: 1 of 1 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -26,8 +26,8 @@
     
     <string name="auto_updates_dialog_title">Select updates to receive</string>
     <string name="auto_updates_dialog_description">Periodically connect to update providers to check for updates.</string>
-    <string name="auto_updates_dialog_manager">Manager updates</string>
-    <string name="auto_updates_dialog_patches">Patches</string>
+    <string name="auto_updates_dialog_manager">ReVanced Manager</string>
+    <string name="auto_updates_dialog_patches">Patch bundles</string>
```
Axelen123, line 30: The option only affects the default bundle, so `Patch Bundle` or `ReVanced Patches` make more sense

oSumAtrIX (approved):

Issue #1290, opened 2023-09-21
bug: Dark mode is broken


kyoyacchi:
### Bug description

I was using the app in dark mode until I updated the app to v1.10.1, but it no longer works. It consistently shows the light mode, whether I enable or disable dark mode or light mode. My phone model is Tecno Spark 7T (Android 11)

![Screenshot_20230921-112605](https://github.com/ReVanced/revanced-manager/assets/63583961/84baef59-68d7-4082-9169-6d62b6e755b0)
![Screenshot_20230921-112609](https://github.com/ReVanced/revanced-manager/assets/63583961/9d1515b9-f7d5-4e20-bdac-5278e59878ba)
![Screenshot_20230921-112611](https://github.com/ReVanced/revanced-manager/assets/63583961/ff1e32eb-6289-49c0-8e06-424b8a4156be)

Video(if needed):

https://github.com/ReVanced/revanced-manager/assets/63583961/2ab6f3fd-fa6e-4be2-9de3-fcc818df0c16

### Version of ReVanced Manager and version & name of application you tried to patch

ReVanced Manager v1.10.1
I didn't try any app to patch...

### Installation type

Non-root

### Device logs

```shell
--------- beginning of main
09-21 11:24:49.488 17108 17108 V PhoneWindow: DecorView setVisiblity: visibility = 0, Parent = android.view.ViewRootImpl@7f5eb08, this = DecorView@181b2a1[MainActivity]
09-21 11:24:49.508 17108 17108 V PhoneWindow: DecorView setVisiblity: visibility = 0, Parent = android.view.ViewRootImpl@7f5eb08, this = DecorView@181b2a1[MainActivity]
09-21 11:24:57.390 17108 17108 V PhoneWindow: DecorView setVisiblity: visibility = 4, Parent = android.view.ViewRootImpl@7f5eb08, this = DecorView@181b2a1[MainActivity]
09-21 11:24:57.429 17108 17185 I GED     : ged_boost_gpu_freq, level 100, eOrigin 2, final_idx 2, oppidx_max 2, oppidx_min 0
```

### Acknowledgements

- [X] This request is not a duplicate of an existing issue.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The issue is solely related to the ReVanced Manager

BenjaminHalko:
This seems to be caused because for some reason manager has enabled Material You on a device that does not support it.

luis-fss:
[...]

PR #1293, opened 2023-09-21
fix: force disable material you on Android 11 and below

BenjaminHalko:
fixes: #1290

### DIFF
--- a/lib/ui/views/navigation/navigation_viewmodel.dart
+++ b/lib/ui/views/navigation/navigation_viewmodel.dart
@@ -1,4 +1,5 @@
 // ignore_for_file: use_build_context_synchronously
+import 'package:device_info_plus/device_info_plus.dart';
 import 'package:dynamic_themes/dynamic_themes.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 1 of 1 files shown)


BenjaminHalko:
I tested this by allowing the toggle of material you on older versions, then turning on material you, quitting, then the new code correctly disables material you.

But I was not able to replicate the bug where Material You is enabled by default on old devices EXCEPT when using the release apk from GitHub.

kartik-commits:
tested it, seems working, hope someone merges

lib/ui/views/navigation/navigation_viewmodel.dart
```diff
@@ -32,7 +33,20 @@ class NavigationViewModel extends IndexTrackingViewModel {
 
     if (prefs.getInt('themeMode') == null) {
       await prefs.setInt('themeMode', 0);
+      await DynamicTheme.of(context)!.setTheme(0);
```
PalmDevs, line 36: Line 40 and 46 uses `DynamicTheme.of(context)` too, so declare it as a variable

PalmDevs (changes requested):
Needs a change

lib/ui/views/navigation/navigation_viewmodel.dart
```diff
@@ -32,7 +33,20 @@ class NavigationViewModel extends IndexTrackingViewModel {
...
     }
+
+    // Force disable Material You on Android 11 and below
+    if (DynamicTheme.of(context)!.themeId.isOdd) {
+      const int ANDROID_12_SDK_VERSION = 31;
+      final AndroidDeviceInfo info = await DeviceInfoPlugin().androidInfo;
+      if (info.version.sdkInt < ANDROID_12_SDK_VERSION) {
```
validcube, line 44: I would personally remove the variable `ANDROID_12_SDK_VERSION` because it's being used once

```suggestion
      final AndroidDeviceInfo info = await DeviceInfoPlugin().androidInfo;
      if (info.version.sdkInt < 31) {
```

validcube (approved):
🥞 This will do, I might re-visit it later!

lib/ui/views/navigation/navigation_viewmodel.dart
BenjaminHalko: It is to make the code more clear as 12 makes more sense than 31

BenjaminHalko:
In order to check I recommend enabling the material you slider, then change it and see if it resets on app load.

PR #1294, opened 2023-09-21
feat: add social links

BenjaminHalko:
Added social links with icons from [DevSrSouza/compose-icons](https://github.com/DevSrSouza/compose-icons)

Screenshot_20230921-155615_ReVanced Manager Debug


app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -48,14 +55,32 @@ fun AboutSettingsScreen(
...
+        Pair(FontAwesomeIcons.Brands.Reddit) {
+            context.openUrl("https://reddit.com/r/app_revanced")
+        },
+        Pair(FontAwesomeIcons.Brands.Twitter) {
+            context.openUrl("https://twitter.com/@revancedapp")
+        },
+        Pair(FontAwesomeIcons.Brands.Youtube) {
+            context.openUrl("https://youtube.com/@revancedapp")
+        },
+    )
```
oSumAtrIX, line 82: Try using [/socials](https://api.revanced.app/docs/swagger#/socials).

BenjaminHalko: Sure thing, the only reason I didn't do it initially was because the Flutter manager did not use it. I thought maybe you wanted to save API calls or something.

oSumAtrIX: Ideally the data layer is dynamically managed by the API so that when changing the API, the data also updates accordingly, which is why the links should be consumed from the API.

Axelen123: That presents a problem because now you have to ping the API

oSumAtrIX: How is that a problem?

BenjaminHalko: So should I or shouldn't I use the API

oSumAtrIX: I believe you should, but lets wait for @Axelen123

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -48,14 +55,32 @@ fun AboutSettingsScreen(
...
+            context.openUrl("https://revanced.app/discord")
+        },
+        Pair(FontAwesomeIcons.Brands.Telegram) {
+            context.openUrl("https://t.me/app_revanced")
+        },
+        Pair(FontAwesomeIcons.Brands.Reddit) {
+            context.openUrl("https://reddit.com/r/app_revanced")
+        },
+        Pair(FontAwesomeIcons.Brands.Twitter) {
+            context.openUrl("https://twitter.com/@revancedapp")
```
validcube, line 77:
```suggestion
            context.openUrl("https://x.com/@revancedapp")
```

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
Axelen123: > How is that a problem and how is that worse than hardcoding?

You brought up pinging as a problem when dealing with auto updates for manager and patches.
It seems like we may have to hardcode anyway because we have to get icons for the links.

oSumAtrIX: > You brought up pinging as a problem when dealing with auto updates for manager and patches

No, I was saying pinging without the user knowing. The popup at the start of the app already asks for pinging.

> It seems like we may have to hardcode anyway because we have to get icons for the links.

You can include a map of names to icons.

Axelen123: > No, I was saying pinging without the user knowing. The popup at the start of the app already asks for pinging.

That popup only mentions auto updates. It does not mention socials

> You can include a map of names to icons and a generic social icon

Yes, and it means fetching from the API doesn't benefit us much because we have to hardcode anyway.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -48,14 +55,32 @@ fun AboutSettingsScreen(
...
+        Pair(FontAwesomeIcons.Brands.Discord) {
+            context.openUrl("https://revanced.app/discord")
+        },
+        Pair(FontAwesomeIcons.Brands.Telegram) {
+            context.openUrl("https://t.me/app_revanced")
+        },
+        Pair(FontAwesomeIcons.Brands.Reddit) {
+            context.openUrl("https://reddit.com/r/app_revanced")
+        },
+        Pair(FontAwesomeIcons.Brands.Twitter) {
```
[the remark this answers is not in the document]
BenjaminHalko: Unfortunately the font awesome library has only been updated to v5.15, Twitter X seems to be added in v6. I found a different font awesome library for compose but it too is only at v5.15.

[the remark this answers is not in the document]
Ushie: We don't have a need for FontAwesome, we can import any SVG as a resource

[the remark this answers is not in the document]
oSumAtrIX: That would be inconsistent. You should use a library, or none, not a library and svgs at the same time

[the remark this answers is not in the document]
Ushie: All social icons can be imported as a resource, there's no need to use an additional icon library aside from Material Icons in my opinion

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
Axelen123: > Then simply change it, not sure how this is an obstracle.

Now we need extra logic for when the user rejects it. This seems like unnecessary complexity to me.

> It does. If I change the API url, the social urls change. While a database of commons social media icons exist, which can be used, the generic social media icon would act in places where the existing set of icons are not enough, just like on revanced.app.

If a project wants to do this then they should maintain a version of revanced manager with altered branding.
The revanced site or linktree seem like better options for handling social links, because they don't require code or network requests. Socials don't change often either, which means there is little to no drawback to hardcoding them (especially if you only need to deal with one link).

oSumAtrIX: > Now we need extra logic for when the user rejects it. This seems like unnecessary complexity to me.

No, the user can not reject it, just like in the current ReVanced Manager. The ping is required if you want to use the app, but the user is given the chance to deny it. On a related note, what would be good is that in the same popup, you can configure the API in case you want to change it without pinging it first.

There is no unnecessary complexity, firstival because it is simply not unnecessary as explained in my previous comment, and second, the complexity boils down to fetching the socials as intended from the API while using a map for the social names and icons, just like it is done for the support page on ReVanced.app.

> If a project wants to do this then they should maintain a version of revanced manager with altered branding

ReVanced Manager offers an API URL as a dynamic data layer. The API supplies socials. If I change the API in the app, it should dynamically adjust to it and also change the socials, no fork should be necessary just to display other socials when the API does that for you already. After all, the endpoint is present just for that purpose.

> The revanced site or linktree seem like better options for handling social links, because they don't require code or network requests

I am not sure what you mean. The website does make requests to the API.

> Socials don't change often either, which means there is little to no drawback to hardcoding them (especially if you only need to deal with one link).

They change every time the API changes. It would be wrong design to hardcode them. 

You merely need to request the socials and display them. If you deny requests to the API, no social links will be present just like no default patches will be present, thus the application exists or you can given the choice to visit the settings to change the API for example in the popup. I don't see why it is such a problem to consume the API.

Axelen123: > No, the user can not reject it, just like in the current ReVanced Manager. The ping is required if you want to use the app, but the user is given the chance to deny it.

The ping was not necessary before and this is not a good reason to change that.

> There is no unnecessary complexity, firstival because it is simply not unnecessary as explained in my previous comment, and second, the complexity boils down to fetching the socials as intended from the API while using a map for the social names and icons, just like it is done for the support page on ReVanced.app.

It is unnecessary because there is a better solution available as I explained above. Fetching is more complex than hardcoding. You have to deal with network failures and the data is now asynchronous.

> ReVanced Manager offers an API URL as a dynamic data layer. The API supplies socials. If I change the API in the app, it should dynamically adjust to it and also change the socials, no fork should be necessary just to display other socials when the API does that for you already. After all, the endpoint is present just for that purpose.

Any successor (or similar) has to maintain their own fork regardless. Different project means different branding. There is no getting around the fact that such a project would eventually need their own fork of patcher as well. Classloader can only get you so far. Forking is an important part of open source software development and any project which would benefit from the dynamic adjustment would also benefit from a fork.

> > The revanced site or linktree seem like better options for handling social links, because they don't require code or network requests
> 
> I am not sure what you mean. The website does make requests to the API.

Just a page filled with social links. Lookup what linktree is.
 
> > Socials don't change often either, which means there is little to no drawback to hardcoding them (especially if you only need to deal with one link).
> 
> They change every time the API changes. It would be wrong design to hardcode them.

No, consuming socials from a network service is unconventional. Everyone hardcodes socials because that makes them easier to work with. Examples: Vanced Manager, Vendetta Manager, Magisk, Aurora Store.

oSumAtrIX: > The ping was not necessary before and this is not a good reason to change that.

The ping is necessary if you want to use "online features" such as updating patches or the app itself therefore already required to ping regardless. It is an extremely good reason to change that due to if you change the API the socials will not change when the API is supplying them and is supposed to be consumed, as it is being done by the website. If you do not do that, hardcode the socials, they will __not__ change if you change the API URL. This is fatal as the socials do not match the API you consume. Your single point is that it pings the API which it inherently does regardless, whereas the fatal drawbacks include what I just mentioned and have mentioned in previous comments.

> It is unnecessary because there is a better solution available as I explained above.

No, hardcoding is not a better solution and fails to solve the issue of changing the API, a dynamic aspect of the app, but not the socials when the API supplies the socials. That being said it is not a solution.

> Fetching is more complex than hardcoding. You have to deal with network failures and the data is now asynchronous.

A network layer for the API client is required and present regardless of the socials. The complexity is overexaggerated as A. The code is present regardless and B. It is merely a network request. Dealing with the API being offline you can simply utilize the caching layer of the network layer to display the socials when offline, this is once again triviality.

> Any successor (or similar) has to maintain their own fork regardless. 

No, I can host my own API while others use upstream of ReVanced Manager to consume my API data.

> Different project means different branding.

The application itself can have a socials section for the application itself if that is what you think is necessary, but the API socials should be displayed too. 

> There is no getting around the fact that such a project would eventually need their own fork of patcher as well.

For the same reason as above, this is also not correct. The API has nothing to do with the patcher.

> Forking is an important part of open source software development and any project which would benefit from the dynamic adjustment would also benefit from a fork.

The API is supposed to prevent the need of forking. The dynamic aspect is lost if you need to fork the app and require users to change the front end just to consume a different backend. For the same reasons as above, a fork is once again unnecessary, just to change the socials.

> No, consuming socials from a network service is unconventional. Everyone hardcodes socials because that makes them easier to work with. Examples: Vanced Manager, Vendetta Manager, Magisk, Aurora Store.

ReVanced Manager is inherently following a different approach and thus unconventional by definition for that reason. Unlike the apps above, it should act as a mere frontend, and not incorporate any data. The data should be consumed from a separate layer, in this case, the API acts as the data source. The application can be branded with where it comes from (simple example: "by ReVanced"), but if I change the API source, the data changes too, and with the data changing, the socials have to change too, as ReVanced Manager is merely the frontend that is not supposed to hardcode any data.

On a related note, your approach shows a problem in a very current event. Twitter rebranded to X and we changed YouTube channels. Your approach requires a frontend to update just to update what is supposed to be the data layer unrelated to the frontend. You would have to push a new ReVanced Manager update solely because we have to change the socials. My approach on the other hand completely fixes this issue and updates the socials dynamically and thus strongly suggests my approach in combination with the reasons provided above.

BenjaminHalko: @oSumAtrIX How should we decide which urls get to be buttons with text and which should just be icons. Maybe a flag gets added to the api like "important"?

It would also be nice if a contact link was added to the API.

oSumAtrIX: > which urls get to be buttons with text and which should just be icons

All buttons should have icons. Unknown socials get a generic icon. You can use a set of icons for the top social media sites such as "tiktok", "x", "twitter", "youtube", and so on.

oSumAtrIX: > It would also be nice if a contact link was added to the API

Here you go: https://api.revanced.app/v2/info

BenjaminHalko: What I meant by "important" was buttons with an icon and text and buttons with just icons. Unless they all should be one or the other.

oSumAtrIX: I'd say you can use social icon buttons for socials. No text needed, but if the social media name is unknown a generic icon would be shown, which is why I'd say its better to show text as well.

BenjaminHalko: One thing that might be of concern is space if there are only 2 socials per row.

BenjaminHalko: Maybe placing them in a drop down like in Flutter is a way to fix that issue?

Axelen123: > The ping is necessary if you want to use "online features" such as updating patches or the app itself therefore already required to ping regardless. It is an extremely good reason to change that due to if you change the API the socials will not change when the API is supplying them and is supposed to be consumed, as it is being done by the website. If you do not do that, hardcode the socials, they will **not** change if you change the API URL. This is fatal as the socials do not match the API you consume. Your single point is that it pings the API which it inherently does regardless, whereas the fatal drawbacks include what I just mentioned and have mentioned in previous comments.

Do users really expect opening the settings to cause network pings?
Do users expect the Github button to take them to the github page belonging to the "ReVanced Upgraded" project (the API they configured) when the page still has the regular ReVanced logo?
I am not convinced that displaying socials is Manager's job.

> > It is unnecessary because there is a better solution available as I explained above.
> 
> No, hardcoding is not a better solution and fails to solve the issue of changing the API, a dynamic aspect of the app, but not the socials when the API supplies the socials. That being said it is not a solution.
> 

Hardcoding a single link to a page filled with socials requires less code and does not need any network requests.
The web page can be altered if socials do change.
 
> > Fetching is more complex than hardcoding. You have to deal with network failures and the data is now asynchronous.
> A network layer for the API client is required and present regardless of the socials. The complexity is overexaggerated as A. The code is present regardless and B. It is merely a network request. Dealing with the API being offline you can simply utilize the caching layer of the network layer to display the socials when offline, this is once again triviality.

I already said this: Network code can fail and is asynchronous. Dealing with that kind of data requires more UI code compared to constants that are immediately available.

> > Different project means different branding.
> 
> The application itself can have a socials section for the application itself if that is what you think is necessary, but the API socials should be displayed too.

Why is displaying the socials of the API ReVanced Manager's job? Can you elaborate on why it is necessary?

> > Any successor (or similar) has to maintain their own fork regardless.
> 
> No, I can host my own API while others use upstream of ReVanced Manager to consume my API data.
>  ...
> > There is no getting around the fact that such a project would eventually need their own fork of patcher as well.
>
> For the same reason as above, this is also not correct. The API has nothing to do with the patcher.
> ...
> The API is supposed to prevent the need of forking. The dynamic aspect is lost if you need to fork the app and require users to change the front end just to consume a different backend. For the same reasons as above, a fork is once again unnecessary, just to change the socials.
> ...
> ReVanced Manager is inherently following a different approach and thus unconventional by definition for that reason. Unlike the apps above, it should act as a mere frontend, and not incorporate any data. The data should be consumed from a separate layer, in this case, the API acts as the data source. The application can be branded with where it comes from (simple example: "by ReVanced"), but if I change the API source, the data changes too, and with the data changing, the socials have to change too, as ReVanced Manager is merely the frontend that is not supposed to hardcode any data.

The dynamic aspect is not useful for socials. If you want to make successor project then you will eventually end up needing to change patcher and/or manager at some point, thereby requiring a manager fork. This is why the data layer is not useful for socials because you can simply change them if you have a fork.

> On a related note, your approach shows a problem in a very current event. Twitter rebranded to X and we changed YouTube channels. Your approach requires a frontend to update just to update what is supposed to be the data layer unrelated to the frontend. You would have to push a new ReVanced Manager update solely because we have to change the socials. My approach on the other hand completely fixes this issue and updates the socials dynamically and thus strongly suggests my approach in combination with the reasons provided above.

Twitter.com links still work and you already said a map would have to be created for icons. You can't change the twitter icon to X without a manager update anyway. 
A link website can be updated with new icons without having to update manager. This approach solves both the twitter rebrand problem and the youtube channel problem.

oSumAtrIX: > Do users really expect opening the settings to cause network pings?

The user was informed about the usage of the API at the beginning. When it happens is not of concern, neither to the user nor the app. You can preload, cache or request just in time, this is not questioned.

> Do users expect the Github button to take them to the github page belonging to the "ReVanced Upgraded" project

It should be clarified that the socials belong to the API. A separate hardcoded link for the app itself can be included that would require a fork as intended to credit the app itself. 

> Hardcoding a single link to a page filled with socials

Which page are you referring to? This does not exist and neither does the API supply one.
Additionally, the code is already present with this PR and spares the user from having to exit the app in order to visit the website and find the social, too many hoops.

> I already said this: Network code can fail and is asynchronous. Dealing with that kind of data requires more UI code compared to constants that are immediately available. 

I don't know what you mean what you mean by UI code. The code is present in this PR and loading a JSON from the API is anything but an obstacle. Again, triviality:

- > A network layer for the API client is required and present regardless of the socials. The complexity is overexaggerated as A. The code is present regardless and B. It is merely a network request. Dealing with the API being offline you can simply utilize the caching layer of the network layer to display the socials when offline, this is once again triviality.

> Why is displaying the socials of the API ReVanced Manager's job? Can you elaborate on why it is necessary?

The API delivers the socials connected to the API. As previously explained, ReVanced Manager acts as a frontend for ReVanced. But you are free to configure the data source. For the frontend, credits can be hardcoded as the frontend you use does not change, regardless of the data source. But for the API the links should be consumed from the API, as you can change it, so the links should also change. 

> The dynamic aspect is not useful for socials. 

This is not the problem. If you change the API, all the data that is supplied has to change too, otherwise, the data you hardcode does not match that of the API. It is quite useful as you can share your socials this way, and if necessary update them.

>  The dynamic aspect is not useful for socials. If you want to make successor project then you will eventually end up needing to change patcher and/or manager at some point, thereby requiring a manager fork

No that is still wrong just like in your previous comment. The API acts as the data source for ReVanced Manager or other clients. I will quote what I already said before:

- > No, I can host my own API while others use upstream of ReVanced Manager to consume my API data.
- > For the same reason as above, this is also not correct. The API has nothing to do with the patcher.

> Twitter.com links still work and you already said a map would have to be created for icons.

Thats good that it works, yet it still acts as an example of how hardcoding what is supposed to be consumed from the API requires changes to the frontend that do not belong to the frontend to begin with.

> You can't change the twitter icon to X without a manager update anyway.

Neither are you forced to (with your approach you 100% are). I will repeat what I said again:

- >  Your approach requires a frontend to update just to update what is supposed to be the data layer unrelated to the frontend
- > but if the social media name is unknown a generic icon would be shown

> A link website can be updated with new icons without having to update manager.

The link hardcoded in the app can not. The icons in the web "frontend" have to be updated by pushing to the frontend, the same way it would require to ReVanced Manager. As I have already solved the problem above. Frontends can provide a subset of popular social media icons, the same way font-awesome does in a broader sense. This set will act as a map for the icons in reliance to high confidence to display an icon for the social link. If it can not find one, in the instance of X for example, it would default to a generic icon, which by choice can be updated (just like it would have to be in the webfront) as explained here:

- > but if the social media name is unknown a generic icon would be shown

Axelen123: The website does not have a socials page yet, but it can be created. The website is also better equipped to deal with socials than the API because it can update both icons as well as links simultaneously and can easily be redeployed. It also makes sense to do so because the websites purpose is to provide information and link to other websites. The socials page on the website can also be updated without having to update manager.

If you *frequently* open manager because you want the social links then you are opening the wrong app because you should be opening a web browser instead. Web browsers also allow you to bookmark frequently accessed pages. Bookmarks and launcher shortcuts can be more easily accessed than managers about page, which requires you to go through the settings page to find it.

It shouldn't be our responsibility to display the social links of other projects. Those projects can create their own website or use linktree for socials instead. It seems weird to display unrelated links next to our logo.

The URL for `revanced.app/socials` can be hardcoded and is unlikely to change. It can be redirected in the event that the link changes. Again, you can update the page without updating manager or anything else.

oSumAtrIX: So how do you deal with the issue that when you change the API, it still links to our socials?

BenjaminHalko: Maybe if it detects your using a different api it can show the socials for revanced and the other api. Because people might want to contact/donate/follow the people who made the app AND the people who are providing other parts.

oSumAtrIX: > it can show the socials for revanced and the other api

Yes, this is what I suggested in my previous comments. The application itself is branded as changing the source does not change the app itself as well as the socials from the API are shown:

> It should be clarified that the socials belong to the API. A separate hardcoded link for the app itself can be included that would require a fork as intended to credit the app itself.

Axelen123: > So how do you deal with the issue that when you change the API, it still links to our socials?

Its intentional. It makes sense to display our own socials because its our app.

oSumAtrIX: > Its intentional. It makes sense to display our own socials because its our app.

Everything else changes though with the API, how do you forward the user to the API's links?

Axelen123: > Everything else changes though with the API, how do you forward the user to the API's links?

The API should only be needed for updates and bundles. The use of the API is not necessary here and the website is better suited for socials than the API as I explained.

oSumAtrIX: > The API should only be needed for updates and bundles.

It supplies the essence of the app, again, if I change the API, the app updates according to it and consumes different patches, how do you forward the user to the source of them?

Axelen123: You can and should be able to use the app without the API. You can use local or remote bundles to patch apps and its possible to update externally. If the user found the API url then they know where to find the source, so it is not necessary to forward the user to the source.

BenjaminHalko: @oSumAtrIX can you add a "branding" entry (link to svg) to the info part of the API so that the about section can load different logos

oSumAtrIX: @Axelen123 

> You can and should be able to use the app without the API.

That does not solve the problem mentioned in my comment, __if__ you use an API:

> It supplies the essence of the app, again, if I change the API, the app updates according to it and consumes different patches or if the patches do not work or something else, the user will seek the socials, how do you forward the user to the source of them?

> If the user found the API url then they know where to find the source, so it is not necessary to forward the user to the source.

So again, just like mentioned before, having to exit the app, and find the correct place and socials from there. Now what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

oSumAtrIX: @BenjaminHalko 

> can you add a "branding" entry (link to svg) to the info part of the API so that the about section can load different logos

That's indeed possible. The /info endpoint can include a branding section. Could you open an issue for this in the API repo?

Axelen123: There is no reason to fetch things over the network when you don't have to.
Just make a bookmark if you need to access a link frequently.
How other projects present their socials is not and should not be our concern.

oSumAtrIX: > There is no reason to fetch things over the network when you don't have to. Just make a bookmark if you need to access a link frequently.

Making a bookmark implies fetching things over the network. 

> How other projects present their socials is not and should not be our concern.

We should give the user the option to see information about the API they use, their origin, their links, and more. After all, the API will 100% act as the data source, unrelated to revanced.app. Again the app branding is not questioned, what is are the socials that are hardcoded and won't change. 

The issue remains unanswered:

 - > So again, just like mentioned before, having to exit the app, and find the correct place and socials from there. Now what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

Axelen123: > > There is no reason to fetch things over the network when you don't have to. Just make a bookmark if you need to access a link frequently.
> 
> Making a bookmark implies fetching things over the network.

That does not happen inside revanced manager. This means we dont have to ask the user about it in the popup or add a setting for it, which is good because now the user is not forced to make as many decisions when they install the app.

> > How other projects present their socials is not and should not be our concern.
> 
> We should give the user the option to see information about the API they use, their origin, their links, and more. After all, the API will 100% act as the data source, unrelated to revanced.app. Again the app branding is not questioned, what is are the socials that are hardcoded and won't change.
> 
> The issue remains unanswered:
> 
>     * > So again, just like mentioned before, having to exit the app, and find the correct place and socials from there. Now what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

I've said these things before and I will say them again:
We are not hardcoding every social link. Only one link to a socials page. The link itself will not change.

How other projects present their socials is not managers responsibility. Its an app for patching.

Adding network dependencies where they are not required is bad practice and should be avoided. We don't need the API to give us socials and the website is better equipped to deal with them.

oSumAtrIX: > That does not happen inside revanced manager.

But it does happen while your argument was that it should not happen at all.

> This means we dont have to ask the user about it in the popup or add a setting for it

As explained already, a popup to connect to the API has to be provided regardless of the socials. If the user wants to use the API, the connection has been made, thus no more popups are needed to fetch the socials. If the user does not want to use the API, the socials for the API that the user does not use anyways simply won't be displayed.

> We are not hardcoding every social link. Only one link to a socials page. The link itself will not change.

You are talking about the app branding. The branding should link to https://revanced.app and not to a socials page, as the origin of the app, regardless of any API would be revanced.app, whereas I am talking about this PR and the socials.

> How other projects present their socials is not managers responsibility. Its an app for patching.

The API is a feature of ReVanced Manager, and thus if you use it, the app should also display the APIs information too. If you do not, the user will not be able to reach out to the API they use in their app:

- >  So again, just like mentioned before, having to exit the app, and find the correct place and socials from there. Now what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

> Adding network dependencies where they are not required is bad practice and should be avoided. We don't need the API to give us socials and the website is better equipped to deal with them.

The dependency is given regardless of the socials:

>  As explained already, a popup to connect to the API has to be provided regardless of the socials. If the user wants to use the API, the connection has been made, thus no more popups are needed to fetch the socials. If the user does not want to use the API, the socials for the API that the user does not use anyways simply won't be displayed.

Axelen123: > > That does not happen inside revanced manager.
> 
> But it does happen while your argument was that it should not happen at all.
> 

It is just a link. Nothing is fetched until you actually want to view socials and open it in a browser. No unnecessary network requests are made here. Your solution **always** fetches even if the user is not there for the socials.

> > This means we dont have to ask the user about it in the popup or add a setting for it
> 
> As explained already, a popup to connect to the API has to be provided regardless of the socials. If the user wants to use the API, the connection has been made, thus no more popups are needed to fetch the socials. If the user does not want to use the API, the socials for the API that the user does not use anyways simply won't be displayed.

That does solve the problem of being bombarded with decisions. I do not object to this, although fetching socials still seems unnecessary.

> 
> > We are not hardcoding every social link. Only one link to a socials page. The link itself will not change.
> 
> You are talking about the app branding. The branding should link to https://revanced.app and not to a socials page, as the origin of the app, regardless of any API would be revanced.app, whereas I am talking about this PR and the socials.

The socials page does not open when you click on the icon on the page. It will be a button or list item just like the other elements in the current version of the page on `compose-dev`. Buttons and list items have no such constraint here.

> > How other projects present their socials is not managers responsibility. Its an app for patching.
> 
> The API is a feature of ReVanced Manager, and thus if you use it, the app should also display the APIs information too. If you do not, the user will not be able to reach out to the API they use in their app:
> 
> * > So again, just like mentioned before, having to exit the app, and find the correct place and socials from there. Now what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

I already answered this. I am not going to repeat it.

> > Adding network dependencies where they are not required is bad practice and should be avoided. We don't need the API to give us socials and the website is better equipped to deal with them.
> 
> The dependency is given regardless of the socials:
> 
> > As explained already, a popup to connect to the API has to be provided regardless of the socials. If the user wants to use the API, the connection has been made, thus no more popups are needed to fetch the socials. If the user does not want to use the API, the socials for the API that the user does not use anyways simply won't be displayed.

My solution achieves the same goal without making network requests when not required.

oSumAtrIX: > It is just a link. Nothing is fetched until you actually want to view socials and open it in a browser. No unnecessary network requests are made here. Your solution always fetches even if the user is not there for the socials.

Well, now the argument boils down to a network request being problematic, which obviously isn't as it is merely a network request. 

> That does solve the problem of being bombarded with decisions. I do not object to this, although fetching socials still seems unnecessary.

Social fetching being unnecessary is wrong as explained in my previous comments (hardcoding is a problem and API changing will still show the hardcoded links wrongly). In any case, the popup has to be shown so the user is aware of any requests, which would also nicely cover the request for socials.

> The socials page does not open when you click on the icon on the page. It will be a button or list item just like the other elements in the current version of the page on compose-dev. Buttons and list items have no such constraint here.

I didn't mean to click on the icon. I said that the app can be branded with ReVanced as changing the API does not change the app, so the links can be hardcoded. See https://github.com/ReVanced/revanced-manager/pull/1294#discussion_r1340615736

> I already answered this. I am not going to repeat it.

You have said:

- > We are not hardcoding every social link. Only one link to a socials page. The link itself will not change.

This does not solve the problem you are refusing to answer to:

> Having to exit the app, and find the correct place and socials for the API they configured or clicking the hardcoded link button on the app, what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

> My solution achieves the same goal without making network requests when not required.

This is not a solution. What we want are socials in the app as outlined in the description of the PR where you can see the socials. You right out remove this feature/ reject this PR by saying that you will link to the website instead where the socials are visible, thus obviously you do not require a request to fetch them.

In order to achieve what this PR does, which is displaying the socials in the app to prevent the user from having to exit the app unnecessarily to see our socials, your initial solution was to hardcode them, as they are right now in this PR, which comes with the problems mentioned above (changing API or updating links, etc.) while your argument for that now boiled down to "we have to make a network request and if we hardcode them we dont" and labeling making a request as a big problem, when you already ask the user to make network requests anyways. An additional network request for the benefit of being able to synchronize it with the selected API and being able to remotely update them is not a problem unlike you claim for that reason.

Axelen123: > Well, now the argument boils down to a network request being problematic, which obviously isn't as it is merely a network request.

Network requests take time, require more UI code and consume data which isn't always free. Not requiring network requests is therefore a benefit.

> Social fetching being unnecessary is wrong as explained in my previous comments (hardcoding is a problem and API changing will still show the hardcoded links wrongly). In any case, the popup has to be shown so the user is aware of any requests, which would also nicely cover the request for socials.

The socials page solution solves all the problems that fetching aims to solve, but without any unnecessary network requests.
Displaying other people's socials is not within the scope of this app.

The socials page can be updated without updating manager. The URL to the page doesn't change and we only need to display our own socials, which means we can hardcode it.

> > The socials page does not open when you click on the icon on the page. It will be a button or list item just like the other elements in the current version of the page on compose-dev. Buttons and list items have no such constraint here.
> 
> I didn't mean to click on the icon. I said that the app can be branded with ReVanced as changing the API does not change the app, so the links can be hardcoded. See [#1294 (comment)](https://github.com/ReVanced/revanced-manager/pull/1294#discussion_r1340615736)

Changing the API does not change the app, but the social links are right next to our branding so they should logically take you to our links.

> 
> > I already answered this. I am not going to repeat it.
> 
> You have said:
> 
>     * > We are not hardcoding every social link. Only one link to a socials page. The link itself will not change.
> 
> 
> This does not solve the problem you are refusing to answer to:
> 
> > Having to exit the app, and find the correct place and socials for the API they configured or clicking the hardcoded link button on the app, what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

The problem of making other people's socials easily accessible is not within the scope of ReVanced Manager.

> > My solution achieves the same goal without making network requests when not required.
> 
> This is not a solution. What we want are socials in the app as outlined in the description of the PR where you can see the socials. You right out remove this feature/ reject this PR by saying that you will link to the website instead where the socials are visible, thus obviously you do not require a request to fetch them.
>
> In order to achieve what this PR does, which is displaying the socials in the app to prevent the user from having to exit the app unnecessarily to see our socials, your initial solution was to hardcode them, as they are right now in this PR, which comes with the problems mentioned above (changing API or updating links, etc.) while your argument for that now boiled down to "we have to make a network request and if we hardcode them we dont" and labeling making a request as a big problem, when you already ask the user to make network requests anyways. An additional network request for the benefit of being able to synchronize it with the selected API and being able to remotely update them is not a problem unlike you claim for that reason.

This PR initially hardcoded all socials, which I agreed to and is a valid solution that works well if you don't need to dynamically update them. You said we needed to be able to update socials without updating manager, so I proposed a different solution (socials page), which also solves that problem. I never rejected what this PR is trying to do (making *our* socials more accessible).

You said my argument boils down to "we have to make a network request and if we hardcode them we dont", but there is more than that. The website is better equipped to deal with socials because it can be easily redeployed and you don't have to update manager in order to change icons.

Displaying someone else's socials is outside the scope of manager and displaying them next to our branding is wrong.

oSumAtrIX: > Network requests take time, require more UI code and consume data which isn't always free. Not requiring network requests is therefore a benefit.

The benefit does not even nearly outweigh the benefits of fetching the socials from the API. The network request is essentially free considering it's size and does not require more UI code as it is abstracted behind the data layer, and does not require time as it can be done asynchronously.

> The socials page solution solves all the problems that fetching aims to solve

No, because it requires the user to exit the app, and is additionally not synced to the configured API. No solution is provided.

> Displaying other people's socials is not within the scope of this app. The socials page can be updated without updating manager. The URL to the page doesn't change and we only need to display our own socials, which means we can hardcode 

The socials (not the credits to the app) belong to the API if you do not display them, this remains once again unanswered:

> Having to exit the app, and find the correct place and socials for the API they configured or clicking the hardcoded link button on the app, what do you think is more likely? Them doing that or just clicking the appealing buttons in the app?

> Changing the API does not change the app, but the social links are right next to our branding so they should logically take you to our links.

As I have explained before, I am not referring to the credits to the app. Repeating:

- > We should give the user the option to see information about the API they use, their origin, their links, and more. After all, the API will 100% act as the data source, unrelated to revanced.app. Again the app branding is not questioned, what is are the socials that are hardcoded and won't change.

> The problem of making other people's socials easily accessible is not within the scope of ReVanced Manager.

That should be a high priority as the user is offered to easily change the API. If I change the data source to osumatrix.me, I expect the user to come to me for anything related to that data, which makes up pretty much 80% the app offers. Again:

- > We should give the user the option to see information about the API they use, their origin, their links, and more. After all, the API will 100% act as the data source, unrelated to revanced.app. Again the app branding is not questioned, what is are the socials that are hardcoded and won't change.

> This PR initially hardcoded all socials, which I agreed to and is a valid solution that works well if you don't need to dynamically update them. You said we needed to be able to update socials without updating manager, so I proposed a different solution (socials page), which also solves that problem. I never rejected what this PR is trying to do (making our socials more accessible).

This is fine, as I explained earlier, the app can be branded with a hardcoded link, may it be the socials inside the app (I prefer this because you don't have to exit the app) or a link to the website (which is good for updating it without having to update ReVanced Manager, as the website consumes the API).

> You said my argument boils down to "we have to make a network request and if we hardcode them we dont", but there is more than that. The website is better equipped to deal with socials because it can be easily redeployed and you don't have to update manager in order to change icons.

No, that is fine, the app can be branded with whatever you want, I have clearly stated this multiple times and I have stated that this is the correct approach as the app itself does not change regardless of the API thus credits to it can be hardcoded.

> Displaying someone else's socials is outside the scope of manager

Refer to what I said earlier regarding it being outside of the scope:

> That should be a high priority as the user is offered to easily change the API. If I change the data source to osumatrix.me, I expect the user to come to me for anything related to that data, which makes up pretty much 80% the app offers. Again:
>
> - > We should give the user the option to see information about the API they use, their origin, their links, and more. After all, the API will 100% act as the data source, unrelated to revanced.app. Again the app branding is not questioned, what is are the socials that are hardcoded and won't change.

> and displaying them next to our branding is wrong.

How you display them is to be figured out by the ReVanced Manager devs. I did not say to add it next to our branding, what I said is that it has to be easily identifiable as the socials of the API, see https://github.com/ReVanced/revanced-manager/pull/1294#discussion_r1340615736:

- > It should be clarified that the socials belong to the API. A separate hardcoded link for the app itself can be included that would require a fork as intended to credit the app itself.

Axelen123: It seems arguing with you isn't worth it since you for example keep on asking the same "Having to exit the app, and find the correct place" question over and over again even though I provided **multiple** valid answers.

Softcoding data that should be hardcoded (every single one or a socials page) is wrong and I have already given several reasons why, but I can let the PR slide if the following things are implemented so we can all move on to something better:
- Make it clear that the socials belong to the API
- Add a setting to disable it (should also be present in the initial popup like oSumAtrIX said)

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -63,24 +63,23 @@ fun AboutSettingsScreen(
...
-            context.openUrl("https://youtube.com/@revancedapp")
-        },
+    val socialIcons = mapOf(
+        "Contact" to Icons.Outlined.MailOutline,
+        "Discord" to FontAwesomeIcons.Brands.Discord, 
+        "Donate" to Icons.Outlined.FavoriteBorder,
+        "GitHub" to FontAwesomeIcons.Brands.Github,
+        "Reddit" to FontAwesomeIcons.Brands.Reddit,
+        "Telegram" to FontAwesomeIcons.Brands.Telegram,
+        "Twitter" to FontAwesomeIcons.Brands.Twitter,
```
oSumAtrIX, line 73: Also add X

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
oSumAtrIX: >  I provided multiple valid answers.

Your answers are:

- Additional complex network request

  This has been argued against already as it is not complex, since it is merely a network request

- Network requests are expensive

   It is cheap, as it is merely a network request, can be cached and preloaded, so does not cause any delay, also small by nature. If offline, as intended no socials should be shown if not cached.

- Network request is useless

  It is not useless as it allows the user to see the socials directly from the app, no need to exit it to find it on a website, also not useless because it prevents users from not bothering to exit the app and using our existing socials in the app


I have been repeating the same answers every time because you have been repeating the same 3 things too.
You claim " I provided multiple valid answers." to "Having to exit the app and find the correct place", but you want the links to be hardcoded, there is therefor __no__ solution to the user having to exit the app to find the APIs socials this way logically speaking. Therefore none of your "multiple valid" solutions is a solution to begin with.


> Softcoding data that should be hardcoded

The data is not supposed to be hardcoded and saying "should be" is wrong therefor because if the API changes the socials have to change too, not possible if hardcoded, whereas the apps branding itself is supposed to be hardcoded, not the socials.

> Make it clear that the socials belong to the API

One way to do it, would be to add the API socials to the API configuration screen, this is completely distinct from the apps branding. Once that is done, move all socials from the "about" page to that and load the socials dynamically. What should be added in the "about" screen is, what you suggested, a hardcoded link to revanced.app, the official distributor of the app. Our socials will be loaded from our api, the app will always be branded with revanced.app (unless forked) and if someone changes the API url, the socials matching it will be displayed.

> Add a setting to disable it 

I didnt mean this about the socials itself. At the first start of the app, the app should notice the user about connecting to the configured API. In this case, by default revanced.app. The warning will ask the user if they want to connect to it to use features such as auto updates. The user should be given the opportunity to not require any API and completely use the app offline (with the option to change this in the settings at any time). But this is out of scope for this PR ofc.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -48,14 +55,31 @@ fun AboutSettingsScreen(
...
         Triple(Icons.Outlined.MailOutline, stringResource(R.string.contact), third = {
             context.openUrl("mailto:nosupport@revanced.app")
         }),
     )
 
+    val socialIcons = mapOf(
+        "Contact" to Icons.Outlined.MailOutline,
+        "Discord" to FontAwesomeIcons.Brands.Discord, 
+        "Donate" to Icons.Outlined.FavoriteBorder,
+        "GitHub" to FontAwesomeIcons.Brands.Github,
```
oSumAtrIX, line 70: Since GitHub is hardcoded as an outlined button, this would show up twice. If we want to dynamically select the outlined/ filled buttons, we'd need to add them to the API response. Not sure how to solve this in another way

BenjaminHalko: This is what I mentioned before. Add an "important" value to social link

oSumAtrIX: We have something similar for the /donations endpoint that have `preferred` set to true for example.

BenjaminHalko: Yeah I was thinking something like that

oSumAtrIX: Waiting on https://github.com/ReVanced/revanced-api/pull/100

oSumAtrIX: Should be gtg.

BenjaminHalko: Should the "preferred" links use the outlined buttons or the filled ones.

BenjaminHalko:
Screenshot_20231104-092538_ReVanced Manager Debug

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
```diff
@@ -15,7 +16,7 @@ class AboutViewModel(private val reVancedAPI: ReVancedAPI) : ViewModel() {
 
     init {
         viewModelScope.launch {
-            withContext(Dispatchers.IO) { reVancedAPI.getSocials().getOrNull() }?.let(
+            withContext(Dispatchers.IO) { reVancedAPI.getSocials("https://api.revanced.app").getOrNull() }?.let(
```
Ushie, line 19: Why are you hardcoding the API?

Ushie (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
BenjaminHalko: Because this is the social page for the manager itself.

BenjaminHalko: A separate about page can show if it detects that the api has change to something other than revanced but this PR is about adding social icons to the ReVanced about page.

Ushie: I don't believe the solution for that is to hardcode the API URL, worse case scenario can't you get the default API key?

Nevertheless, it's rather odd to do it like this, since all the API-dependent screens can also hold the same argument of "it's for the manager itself", updates are for the manager but changing the API URL still changes where updates are received from

BenjaminHalko: Ok good point.

oSumAtrIX: The idea is that the app is branded regardless of changing the API because the same people still developed the app that the branding is for. Implementation wise, I agree, saving the default API URL somewhere as a constant and reading it here sounds like a good idea.

BenjaminHalko: Which file should hold the default?

Ushie: That is just... kind of against the whole reason we added branding to API for

If the plan is to essentially soft-hardcode, just hardcode altogether, there was no reason to change the ReVanced icon to work through the API especially considering that even if we change the logo remotely, that doesn't mean the app logo updates

Not only is loading the image remotely worse, but it can be inconsistent with the current app icon and also make the screen unnecessarily dependent on the internet

BenjaminHalko: Yeah I agree

oSumAtrIX: > If the plan is to essentially soft-hardcode, just hardcode altogether

In this case, we would not be able to update the links. The point is that the API can ship its own links, but that would be added somewhere else instead of the about page of ReVanced Manager, whereas the about page will have the branding of its origin with the ability to update links if needed.

BenjaminHalko: Personally I think that all the links should be hardcoded, the argument that "We would have to update the app to change links" is kind of pointless when we are still updating the app for other reasons quite often.

BenjaminHalko: If the API changes then the second about page that loads socials from the new API could be used though.

Ushie: Why will there be a second about page?? what did I miss?

oSumAtrIX: In this case, the links from the API can be shown somewhere in the screen for the API.

BenjaminHalko: In the event that someone makes their own API, then users should have option to contact them in case something goes wrong.

BenjaminHalko: Though adding a second about page is not what this PR is about.

BenjaminHalko: | In this case, the links from the API can be shown somewhere in the screen for the API.

Yeah that is a good idea.

BenjaminHalko: @oSumAtrIX So do we return to hardcoding the manager social links?

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
oSumAtrIX: I believe, semantically, the filled ones are considered highlighted/preferred.

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
oSumAtrIX: Consuming the API is smarter as you can update everything remotely. What is the argument against that?

BenjaminHalko: Half a second delay from opening the about page to showing the socials. Plus greater control of how they are laid out if they are hardcoded.

oSumAtrIX: > Half a second delay from opening the about page to showing the socials using the API

You can preload and cache requests.

> Plus greater control of how they are laid out if they are hardcoded.

Can you explain?

app/build.gradle.kts
```diff
@@ -158,4 +158,7 @@ dependencies {
 
     // Markdown
     implementation(libs.markdown.renderer)
+
+    // Icons
+    implementation("br.com.devsrsouza.compose.icons:font-awesome:1.1.0")
```
Axelen123, line 163: Use the gradle version catalog.

Axelen123, line 163: What is the license of the icons *and* the library? non-free icons should be avoided

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -37,42 +36,60 @@ import androidx.compose.ui.res.stringResource
...
     })
 
-    val filledButton = listOf(
-        Triple(Icons.Outlined.FavoriteBorder, stringResource(R.string.donate)) {
-            context.openUrl("https://revanced.app/donate")
-        },
-        Triple(Icons.Outlined.Language, stringResource(R.string.website), third = {
-            context.openUrl("https://revanced.app")
-        }),
+    val socialIcons = mapOf(
```
Axelen123, line 70: This should be a global variable or part of a companion object.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -37,42 +36,60 @@ import androidx.compose.ui.res.stringResource
...
+        Triple(socialIcons[it.name] ?: Icons.Outlined.Language, it.name, third = {
+            context.openUrl(it.url)
+        })
+    }
+
+    val socialButtons = viewModel.socials.filterNot(ReVancedSocial::preferred).map {
+        Pair(socialIcons[it.name] ?: Icons.Outlined.Language) {
+            context.openUrl(it.url)
+        }
+    }
```
Axelen123, line 92: Use `partition` instead

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
Axelen123, line 92: Try to use partition here instead

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
BenjaminHalko: > Can you explain?

For example, with the donation links we used to have a link to https://revanced.app/donate but now there is no link to there on the API so it doesn't show up. I guess you could add it to the socials or donations sections of the API, so I guess I don't really know.

oSumAtrIX: Yeah, it can be added to the API in case we want to, though we have a dedicated endpoint for that in case we want to implement it: https://api.revanced.app/v2/donations

BenjaminHalko: Yeah that is where I was thinking it might get placed.

Axelen123:
@BenjaminHalko what is the status of this?

app/build.gradle.kts
BenjaminHalko: Both the [library](https://github.com/DevSrSouza/compose-icons/blob/master/LICENSE) and [font awesome](https://github.com/DevSrSouza/compose-icons/blob/master/font-awesome/src/commonMain/resources/font-awesome-license.txt) are free to use.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
BenjaminHalko: If I were to make `socialIcons` a global variable, where should I declare it?

BenjaminHalko:
This should finally be done now.
Screenshot_20240111-093813_ReVanced Manager Debug

oSumAtrIX:
The chips have different widths, maybe settle on one.

BenjaminHalko:
Screenshot_20240112-090526_ReVanced Manager Debug

oSumAtrIX:
A future PR may follow up on our past conclusion regarding hardcoding the app's branding here and move the dynamic part implemented here to the API screen. The concluded reason above was that despite changing the API, the app itself is still vendored by us and, therefor, branded with our links and whoever wants to change the branding has to fork the app and vendor it themselves with their branding. In contrast, the API can deliver additional information about the API in the API settings page.

Apart from that the PR lgtm!

Ushie:
> The concluded reason above was that despite changing the API, the app itself is still vendored by us and, therefor, branded with our links and whoever wants to change the branding has to fork the app and vendor it themselves with their branding.

This sounds perfectly reasonable to be, even if the API is changed you still are using **our** app, in which case should always use **our** branding

I personally see no need in adding an About screen that is strictly for the API, that would be rather odd and I'm not particularly sure how we would approach that

How would it look? where would it be placed?

oSumAtrIX:
I have mentioned where this can be added:

> In contrast, the API can deliver additional information about the API in the API settings page.

When you use an API, it is useful to mention their information to signify that you are now consuming from them and where to reach out to for issues and other reasons.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -9,70 +9,98 @@ import androidx.compose.foundation.layout.Column
...
-            context.openUrl("mailto:nosupport@revanced.app")
-        }),
+    val filledButtons = mutableListOf<Triple<ImageVector, String, () -> Unit>>()
+    filledButtons.addAll(
+        preferred.map {
+            Triple(getSocialIcon(it.name), it.name, third = {
+                context.openUrl(it.url)
+            })
+        }
     )
```
Axelen123, line 72: You can get rid of the mutable list and use a `remember` block instead with `preferred` as the key

gradle/libs.versions.toml
```diff
@@ -29,6 +29,7 @@ app-icon-loader-coil = "1.5.0"
 skrapeit = "1.2.2"
 libsu = "5.2.0"
 scrollbars = "1.0.4"
+composeIcons = "1.1.0"
```
Axelen123, line 32: Rename this to `compose-icons`

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
```diff
@@ -0,0 +1,60 @@
...
+            }?.let {
+                socials.addAll(it.socials)
+                contact = it.contact.email
+                donate = it.donations.links.partition(ReVancedDonationLink::preferred).first.firstOrNull()?.url
+            }
+        }
+    }
+
+    companion object {
+        private val _socialIcons = mapOf(
```
Axelen123, line 48: Rename this to `socialIcons`

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
```diff
@@ -0,0 +1,60 @@
...
+    var donate: String? = null
+
+    init {
+        viewModelScope.launch {
+            withContext(Dispatchers.IO) {
+                reVancedAPI.getInfo("https://api.revanced.app").getOrNull()
+            }?.let {
+                socials.addAll(it.socials)
+                contact = it.contact.email
+                donate = it.donations.links.partition(ReVancedDonationLink::preferred).first.firstOrNull()?.url
```
Axelen123, line 42: Use `.find` instead of `partition`

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
```diff
@@ -0,0 +1,60 @@
...
+import compose.icons.fontawesomeicons.brands.Twitter
+import compose.icons.fontawesomeicons.brands.Youtube
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class AboutViewModel(private val reVancedAPI: ReVancedAPI) : ViewModel() {
+    val socials = mutableStateListOf<ReVancedSocial>()
+    var contact: String? = null
+    var donate: String? = null
```
Axelen123, line 33: This is not compose state. The UI might not update. Use `mutableStateOf`.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -9,70 +9,98 @@ import androidx.compose.foundation.layout.Column
...
+                Icons.Outlined.MailOutline,
+                stringResource(R.string.contact),
+                third = {
+                    context.openUrl(viewModel.contact!!)
+                }
+            )
+        )
+    }
+
+    val socialButtons = notPreferred.map {
```
Axelen123, line 98: Use `remember` with `notPreferred` as the key.

oSumAtrIX (approved):

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
```diff
@@ -0,0 +1,62 @@
...
+import compose.icons.fontawesomeicons.brands.Twitter
+import compose.icons.fontawesomeicons.brands.Youtube
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class AboutViewModel(private val reVancedAPI: ReVancedAPI) : ViewModel() {
+    val socials = mutableStateListOf<ReVancedSocial>()
+    val contact = mutableStateOf<String?>(null)
+    val donate = mutableStateOf<String?>(null)
```
Axelen123, line 35: Use the `by` keyword to avoid having to use `.value`

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -9,69 +9,98 @@ import androidx.compose.foundation.layout.Column
...
 
-    val filledButton = listOf(
-        Triple(Icons.Outlined.FavoriteBorder, stringResource(R.string.donate)) {
-            context.openUrl("https://revanced.app/donate")
-        },
-        Triple(Icons.Outlined.Language, stringResource(R.string.website), third = {
-            context.openUrl("https://revanced.app")
-        }),
-    )
+    val (preferred, notPreferred) = viewModel.socials.partition(ReVancedSocial::preferred)
```
Axelen123, line 63: Wrap the `partition` in a `remember` block and rename the variables to something like `preferredSocials`

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -9,69 +9,98 @@ import androidx.compose.foundation.layout.Column
...
 
-    val outlinedButton = listOf(
-        Triple(Icons.Outlined.Code, stringResource(R.string.github), third = {
-            context.openUrl("https://revanced.app/github")
-        }),
-        Triple(Icons.Outlined.MailOutline, stringResource(R.string.contact), third = {
-            context.openUrl("mailto:nosupport@revanced.app")
-        }),
-    )
+    val filledButtons = remember(preferred) {
```
Axelen123, line 65: Rename to `preferredSocialButtons`.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -9,69 +9,98 @@ import androidx.compose.foundation.layout.Column
...
+        filledButtons.add(
+            Triple(
+                Icons.Outlined.MailOutline,
+                stringResource(R.string.contact),
+                third = {
+                    context.openUrl(viewModel.contact.value!!)
+                }
+            )
+        )
+    }
```
Axelen123, line 95: Create a list containing the hardcoded socials in the earlier `remember` block and use the `+` operator.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -9,69 +9,98 @@ import androidx.compose.foundation.layout.Column
...
+        }
+    }.toMutableList()
+
+    viewModel.donate.value?.let {
+        filledButtons.add(
+            Triple(
+                Icons.Outlined.FavoriteBorder,
+                stringResource(R.string.donate),
+                third = {
+                    context.openUrl(viewModel.donate.value!!)
```
Axelen123, line 79:
```suggestion
                    context.openUrl(it)
```

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -152,23 +182,20 @@ fun AboutSettingsScreen(
...
-                            )
-                            Text(
-                                text,
-                                style = MaterialTheme.typography.labelLarge,
-                                color = MaterialTheme.colorScheme.primary
+                                modifier = Modifier.size(28.dp),
+                                tint = MaterialTheme.colorScheme.secondary
                             )
                         }
                     }
```
Ushie, line 201: Please move this out of the FlowRow, any new entry added in the API will cause this to break

Ushie (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -132,9 +161,10 @@ fun AboutSettingsScreen(
                 maxItemsInEachRow = 2,
                 horizontalArrangement = Arrangement.spacedBy(8.dp, Alignment.CenterHorizontally)
             ) {
-                filledButton.forEach { (icon, text, onClick) ->
+                filledButtons.forEach { (icon, text, onClick) ->
                     FilledTonalButton(
-                        onClick = onClick
+                        onClick = onClick,
+                        modifier = Modifier.defaultMinSize(minWidth = 130.dp),
```
Ushie, line 167: Maybe better to find a solution instead of arbitrarily choose a width, as an entry being added in the API that has slightly more characters can cause this to fail

Ushie (changes requested):

BenjaminHalko:
I think it is finally done.

BenjaminHalko:
Screenshot_20240124-094455_ReVanced Manager Debug

Ushie:
Is there anyway to stop the Filled Buttons from taking all the available width? it looks odd

BenjaminHalko:
We could always increase the side padding, but there is no way to make it use the minimum space possible while also having the buttons be the same size. The fact that the buttons were the same size before was just a coincidence.

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -35,43 +33,76 @@ import androidx.compose.ui.res.stringResource
...
+            viewModel.contact?.let {
+                Triple(
+                    Icons.Outlined.MailOutline,
+                    contactText,
+                    third = {
+                        context.openUrl("mailto:$it")
+                    }
+                )
+            }
+        )
```
Axelen123, line 96: You can use the `+` operator on the list you get from `map`. This would avoid having to use `toTypedArray()`.

app/src/main/java/app/revanced/manager/ui/viewmodel/AboutViewModel.kt
```diff
@@ -0,0 +1,56 @@
...
+import compose.icons.fontawesomeicons.brands.XTwitter
+import compose.icons.fontawesomeicons.brands.Youtube
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.withContext
+
+class AboutViewModel(private val reVancedAPI: ReVancedAPI) : ViewModel() {
+    var socials by mutableStateOf(emptyList<ReVancedSocial>())
+    var contact by mutableStateOf<String?>(null)
+    var donate by mutableStateOf<String?>(null)
```
Axelen123, line 29:
```suggestion
    var socials by mutableStateOf(emptyList<ReVancedSocial>())
    	private set
    var contact by mutableStateOf<String?>(null)
      	private set
    var donate by mutableStateOf<String?>(null)
      	private set
```

gradle/libs.versions.toml
```diff
@@ -110,6 +111,9 @@ libsu-nio = { group = "com.github.topjohnwu.libsu", name = "nio", version.ref =
 # Scrollbars
 scrollbars = { group = "com.github.GIGAMOLE", name = "ComposeScrollbars", version.ref = "scrollbars" }
 
+# Compose Icons
+compose-icons-fontawesome = { group = "com.github.BenjaminHalko.compose-icons", name = "font-awesome", version.ref = "compose-icons" }
```
Axelen123, line 115: What changes are there compared to the upstream version of the library?

Axelen123 (changes requested):

gradle/libs.versions.toml
BenjaminHalko: My version uses font awesome 6. The biggest change is that it uses XTwitter instead of the Twitter icon.

Axelen123: Can you add that as a comment in the toml file?

BenjaminHalko:
Ok it might be done now...

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -35,43 +33,76 @@ import androidx.compose.ui.res.stringResource
...
+            viewModel.donate?.let {
+                Triple(
+                    Icons.Outlined.FavoriteBorder,
+                    donateText,
+                    third = {
+                        context.openUrl(it)
+                    }
+                )
+            }
+        ) + listOfNotNull(
```
Axelen123, line 86: We only need one list
```suggestion
            },
```

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -130,19 +164,21 @@ fun AboutSettingsScreen(
...
+                        modifier = Modifier.weight(1f),
                     ) {
                         Row(
                             horizontalArrangement = Arrangement.spacedBy(8.dp),
                             verticalAlignment = Alignment.CenterVertically
                         ) {
                             Icon(
                                 icon,
-                                contentDescription = null,
+                                contentDescription = text,
```
Axelen123, line 181: This icon should not have a content description because the preferred social buttons already have `Text` elements associated with them unlike the non-preferred socials.

```suggestion
                                contentDescription = null,
```

app/src/main/java/app/revanced/manager/ui/screen/settings/AboutSettingsScreen.kt
```diff
@@ -35,43 +33,79 @@ import androidx.compose.ui.res.stringResource
...
+                    donateText,
+                    third = {
+                        context.openUrl(it)
+                    }
+                )
+            },
+            viewModel.contact?.let {
+                Triple(
+                    Icons.Outlined.MailOutline,
+                    contactText,
```
Axelen123, line 89: `context.getString` can be used to get the contact text and donate text

Axelen123 (changes requested):

Axelen123 (approved):

Issue #1176, opened 2023-08-28
feat: add debug info to patching screen


taku-nm:
### Type

Functionality

### Issue

An astounding amount of users simply posts a screenshot of the patching screen saying "Aborted". When they're feeling gracious, they might add some info like "latest version" themselves, but often the user is unaware that there is a newer version available.

So the core issue is manager only showing the errors the patcher throws, without adding any debug info.
Sometimes its helpful, sometimes it isn't. 

### Feature

Display additional debug info.

this should include:
- Manager Version
- Package name of app selected
- Version of app selected
- Number of patches selected
- Default patches?
- APK loaded from storage or app list?
- Split APK for not?
- Android Version

in order to minimize the screenspace required for that info I'd suggest packing it all into a short string:

1.9.3:com.google.android.youtube:18.23.35:52:Y:S:NS:11

This way there is some flexibility as to where to include that information.
It's important that the debug information is on display regardless of where the user might've scrolled to in the log.

Poorly photoshopped example:
![aborted](https://github.com/ReVanced/revanced-manager/assets/23640508/211a9634-bc5a-4fff-83cc-6fca73c41a4d)
(android version not pictured as I thought about this after the fact)

Additionally, one might drop the information on "Number of patches selected" and "Default patches" in favour of an integer that is computed based on the selected patches. Similar to discords bot permissions integer.

### Motivation

Efficiency and speed.
Having this debug information would immediately outline the source of the error in the overwhelming majority of cases that land in the support channel.
This reduces additional questioning and guidance, especially when the user for example doesn't know what versions he's using.

It gives all supporters a fairly detailed insight into what the user is doing (wrong).

### Additional context

One might also consider a pop-up that notifies the user to either send a screenshot or to click [this] button to export the log to ask for further assistance. In the later case one should attach the debug string at the beginning of the exported logfile. 

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

validcube:
This feature request looks interesting, I think the information could be outputed to the installer log instead of the app bar, would that looks good?

Ushie:
[...]

PR #1299, opened 2023-09-22
feat(patcher): improve logs

BenjaminHalko:
Adds a debug popup upon crash, making it easier for users to report.

Screenshots:

Clipboard Output:
```
~ Device Info
Manager: 1.10.2
Build: debug
Model: SM-G973U1
Android Version: 12
Supported Archs: arm64-v8a, armeabi-v7a, armeabi
Rooted: false

~ Patch Info
App: com.benjaminhalko.bladesofdoom v1.0.2
Patch Version: v2.190.0
Patches: Override certificate pinning

~ Settings
Enabled changing patches: true
Enabled universal patches: true
Enabled experimental patches: true

~ Logs
Copying APK
Reading APK
Decoding app manifest
Loading patches
Deleting existing resource cache directory
Decoding resources
Executing patches
Override certificate pinning succeeded
Compiling modified dex files
Compiling modified resources

~ Exception
An error occurred:
brut.androlib.exceptions.AndrolibException: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~yX9hzw4GXSh5MAaEn7upIQ==/app.revanced.manager.flutter.debug-9bLAJBhUXGH8yrLPrQBL-Q==/lib/arm64/libaapt2.so, compile, --dir, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-UQBTXQ/cache/res, --legacy, -o, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-UQBTXQ/cache/build/resources.zip]
	at brut.androlib.AaptInvoker.invokeAapt2(AaptInvoker.java:123)
	at brut.androlib.AaptInvoker.invokeAapt(AaptInvoker.java:391)
	at app.revanced.patcher.data.ResourceContext.get(ResourceContext.kt:126)
	at app.revanced.patcher.Patcher.get(Patcher.kt:232)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$17(MainActivity.kt:270)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$ano5ZEDeC0qujvhPEYSwCbNnLNQ(Unknown Source:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda3.run(Unknown Source:24)
	at java.lang.Thread.run(Thread.java:920)
Caused by: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~yX9hzw4GXSh5MAaEn7upIQ==/app.revanced.manager.flutter.debug-9bLAJBhUXGH8yrLPrQBL-Q==/lib/arm64/libaapt2.so, compile, --dir, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-UQBTXQ/cache/res, --legacy, -o, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-UQBTXQ/cache/build/resources.zip]
	at brut.util.OS.exec(OS.java:104)
	at brut.androlib.AaptInvoker.invokeAapt2(AaptInvoker.java:119)
	... 7 more

```
(the triple backticks are part of the output)


BenjaminHalko:
Honestly I think the design of the popup can be changed, instead of `Show Exception` replace it with `Copy Exception to Clipboard` and then don't have the screen that shows the exception since currently the exception still shows up in the output window behind the dialog.

We should probably add a dismiss button too.

Ushie:
Instead of "Copy Exception to Clipboard", it should be simplified to something like "Copy exception", as the clipboard part isnt exactly relevant

Ushie:
If possible, include the custom source settings too into the "Settings" section of the log output

oSumAtrIX:
Regarding the last popup refer to https://github.com/ReVanced/revanced-manager/issues/1176#issuecomment-1730506029

BenjaminHalko:
A single screen like this would be better right?
Screenshot_20230922-120948_ReVanced Manager Debug

Though I don't like how connected the 2 buttons look. Maybe if they had an outline like other buttons had??

BenjaminHalko:
I could add padding to the buttons to give them more space as well.

BenjaminHalko:
This seems better:
Screenshot_20230922-124308_ReVanced Manager Debug

BenjaminHalko:
Updated output with patch/integration repo:
```
~ Device Info
Manager: 1.10.2
Build: debug
Model: SM-G973U1
Android Version: 12
Supported Archs: arm64-v8a, armeabi-v7a, armeabi

~ Patch Info
App: com.benjaminhalko.bladesofdoom v1.0.2
Patch Version: v2.190.0
Patches: Override certificate pinning

Settings
Enabled changing patches: true
Enabled universal patches: true
Enabled experimental patches: true
Patches repo: revanced/revanced-patches
Integration repo: revanced/revanced-integrations

~ Logs
Copying APK
Reading APK
Decoding app manifest
Loading patches
Deleting existing resource cache directory
Decoding resources
Executing patches
Override certificate pinning succeeded
Compiling modified dex files
Compiling modified resources

~ Exception
An error occurred:
brut.androlib.exceptions.AndrolibException: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~TgFNemG2J2VM_Zgkj5WXzQ==/app.revanced.manager.flutter.debug-YHK3W1X6ePkcwIE45TiaIg==/lib/arm64/libaapt2.so, compile, --dir, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-EBVINP/cache/res, --legacy, -o, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-EBVINP/cache/build/resources.zip]
	at brut.androlib.AaptInvoker.invokeAapt2(AaptInvoker.java:123)
	at brut.androlib.AaptInvoker.invokeAapt(AaptInvoker.java:391)
	at app.revanced.patcher.data.ResourceContext.get(ResourceContext.kt:126)
	at app.revanced.patcher.Patcher.get(Patcher.kt:232)
	at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$17(MainActivity.kt:270)
	at app.revanced.manager.flutter.MainActivity.$r8$lambda$ano5ZEDeC0qujvhPEYSwCbNnLNQ(Unknown Source:0)
	at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda3.run(Unknown Source:24)
	at java.lang.Thread.run(Thread.java:920)
Caused by: brut.common.BrutException: could not exec (exit code = 1): [/data/app/~~TgFNemG2J2VM_Zgkj5WXzQ==/app.revanced.manager.flutter.debug-YHK3W1X6ePkcwIE45TiaIg==/lib/arm64/libaapt2.so, compile, --dir, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-EBVINP/cache/res, --legacy, -o, /data/user/0/app.revanced.manager.flutter.debug/cache/patcher/tmp-EBVINP/cache/build/resources.zip]
	at brut.util.OS.exec(OS.java:104)
	at brut.androlib.AaptInvoker.invokeAapt2(AaptInvoker.java:119)
	... 7 more

```

Ushie:
Maybe "Copy logs" would make more sense and look better

BenjaminHalko:
@Ushie done.

I was also wondering if we should also update the export log button that is in the bottom left to also output the various info that this dialog outputs like device info, relevant settings, etc.

BenjaminHalko:
Screenshot_20230922-132048_ReVanced Manager Debug

BenjaminHalko:
huh. why does the pr build keep failing?

Ushie:
That's an issue with the latest commit on the dev branch, I'll push a fix in a sec

Ushie:
> @Ushie done.
> 
> I was also wondering if we should also update the export log button that is in the bottom left to also output the various info that this dialog outputs like device info, relevant settings, etc.

Yeah totally

BenjaminHalko:
@Ushie Should that be in a different PR or can I update that output in this.

Ushie:
This PR can be generalized to "improving logs"

assets/i18n/en_US.json
```diff
@@ -162,6 +162,12 @@
         "exportApkButtonTooltip": "Export patched APK",
         "exportLogButtonTooltip": "Export log",
 
+        "patchErrorDialogTitle": "Something went wrong",
+        "patchErrorDialogText": "During patching an exception occurred.",
```
Ushie, line 166:
```suggestion
        "patchErrorDialogText": "An error occured while patching.",
```

Maybe this would be better?

BenjaminHalko:
Can a button be added to the bottom next to the export log button. A share log button that opens up the share menu. It would allow sharing to places like Discord and allow users to copy to clipboard if they accidentally close the popup.

Ushie (changes requested):

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -169,6 +176,77 @@ class InstallerViewModel extends BaseViewModel {
...
+                  '~ Device Info',
+                  'ReVanced Manager: ${info['version']}',
+                  'Build: ${info['flavor']}',
+                  'Model: ${info['model']}',
+                  'Android version: ${info['androidVersion']}',
+                  'Supported Archs: ${info['supportedArch'].join(", ")}',
+
+                  '\n~ Patch Info',
+                  'App: ${_app.packageName} v${_app.version}',
+                  'Patch Version: ${_managerAPI.patchesVersion}',
```
Ushie, line 221:
```suggestion
                  'Patches version: ${_managerAPI.patchesVersion}',
```

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -169,6 +176,77 @@ class InstallerViewModel extends BaseViewModel {
...
+                  'Patches: ${_patches.map((p) => p.name).toList().join(", ")}',
+
+                  '\n~ Settings',
+                  'Enabled changing patches: ${_managerAPI.isPatchesChangeEnabled()}',
+                  'Enabled universal patches: ${_managerAPI.areUniversalPatchesEnabled()}',
+                  'Enabled experimental patches: ${_managerAPI.areExperimentalPatchesEnabled()}',
+                  'Patches repo: ${_managerAPI.getPatchesRepo()}',
```
Ushie, line 228:
```suggestion
                  'Patches source: ${_managerAPI.getPatchesRepo()}',
```

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -169,6 +176,77 @@ class InstallerViewModel extends BaseViewModel {
...
+                  'Integration repo: ${_managerAPI.getIntegrationsRepo()}',
```
Ushie, line 229:
```suggestion
                  'Integration source: ${_managerAPI.getIntegrationsRepo()}',
```

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -169,6 +176,77 @@ class InstallerViewModel extends BaseViewModel {
...
+            onPressed: () => {
+              Clipboard.setData(
+                ClipboardData(text: [
+                  '```',
+                  '~ Device Info',
+                  'ReVanced Manager: ${info['version']}',
+                  'Build: ${info['flavor']}',
+                  'Model: ${info['model']}',
+                  'Android version: ${info['androidVersion']}',
+                  'Supported Archs: ${info['supportedArch'].join(", ")}',
```
Ushie, line 217:
```suggestion
                  'Supported architectures: ${info['supportedArch'].join(", ")}',
```

Ushie (approved):

BenjaminHalko:
I was also wondering if the export log function could export a `.txt` instead of a `.log` as txts can be opened much more easily on Android.

Ushie:
.txt should work fine, i think it's only .log because the extension speaks for itself

BenjaminHalko:
I think this might be ready to merge.

BenjaminHalko:
@oSumAtrIX anything else to add?

BenjaminHalko:
@Ushie is there anything else to add or can I merge

Ushie:
Let's wait for Palm's approval for the popup

oSumAtrIX:
The popup doesn't animate the user to copy the exception. This will lead users making a screenshot of that popup. An example solution that may or may not be used can be the following:

If an error occurs, show the popup with the exception as originally in the PR as well as the copy button.

If the user tries to screenshot it, trigger a copy to clipboard event;

The popup when screenshotted could say: Instead of a screenshot, the log has been copied to your clipboard.

Any attempt to screenshot an error can be handled the same way in the log when the popup is closed.

BenjaminHalko:
that sounds like a good idea.

oSumAtrIX:
It may or may not be unexpected UX because the intended action is a screenshot not copy to clipboard, so it may not be the best idea, but if it sounds good to you, we can go with this (sorry for first saying to remove the exception from the popup then suggesting to add it back again, this idea just popped in my head right now)

oSumAtrIX:
The PR can be simplified the following way:

Show the exception as usual in the log.
Add a copy exception button somewhere, in case the user wants to copy it. If they try to screenshot, show a popup saying "Instead copied to clipboard". This was no initial popup is necessary and an appropriate feedback is given to the sudden UX.

BenjaminHalko:
I can move the copy logs button to the bottom beside the export log button.

oSumAtrIX:
An export log and export exception button sounds a little too much. How about merging both to one button. The export log button can copy the log (+ the appended metadata) to clipboard (show a popup on button press explaing that the log has been copied to clipboard)

BenjaminHalko:
that sounds good, exporting logs to a file seemed unnecessary anyways.

oSumAtrIX:
The original reason for that was because the log may be very long, and depending on that it may be difficult to send to places such as Discord.

oSumAtrIX:
In case this is a considerable problem, there is a couple of approaches, such as providing the option to save to file or clipboard, providing the option to only clip the error or full log or the option to upload to services like pastebin (API key needed)

BenjaminHalko:
Does the manager currently have a way of detecting screenshots?

I found [this](https://pub.dev/packages/screenshot_callback) but I don't want to add unnecessary packages.

oSumAtrIX:
I believe there are Android APIs that can prevent a screenshot, but I don't know if they can be disabled on certain activities, and if an event handler can be added. Depending on another package doesn't sound too bad to me

BenjaminHalko:
New design:
Screenshot_20230922-170038_ReVanced Manager Debug

Ushie:
By guidelines you can't put a button there, use an IconButton like previously

oSumAtrIX:
@Ushie would a copy icon do the job?

Ushie:
Why was the existing button removed?

BenjaminHalko:
I changed it so that people would be more drawn to it, as it more obviously shows that you can copy logs than just a simple icon, but if it doesn't follow guidelines I can change it.

oSumAtrIX:
I can see where @BenjaminHalko is coming from. The icon button may not be as appealing to the user as the button with the text. Though, the safe guard here is that the screenshot action would be replaced with the copy action, so it's find

BenjaminHalko:
That makes sense.

BenjaminHalko:
should I change the icon to a copy icon or leave it as is.

BenjaminHalko:
Personally I think leaving it as is might be better, the copy icon seems a bit more vague than the one we have right now.

BenjaminHalko:
It might be a good idea to only disable screenshots on the release version. What if we want to screenshot that section for dev purposes.

oSumAtrIX:
> Personally I think leaving it as is might be better, the copy icon seems a bit more vague than the one we have right now.

Because the safeguard exists, I believe following the guidelines is fine.



> It might be a good idea to only disable screenshots on the release version. What if we want to screenshot that section for dev purposes.

Sure, sounds sensible to me

oSumAtrIX:
Btw, recording the screen should not be restricted. The screenshot functionality is disabled merely to motivate to copy the full log. No one would go out of their way to record instead of copying the logs and recording is used for debugging purposes often

BenjaminHalko:
> Because the safeguard exists, I believe following the guidelines is fine.

Very sorry, I meant keeping the old icon instead of replacing it with a plain "copy icon". Not using a button that has text.

oSumAtrIX:
Interesting. Is this icon really the "copy to clipboard" icon? It looks very different.

BenjaminHalko:
Is the copy to clipboard icon different from the standard [copy button](https://api.flutter.dev/flutter/material/Icons/content_copy-constant.html)?

BenjaminHalko:
You know, after waiting a while, maybe the post_add icon is not really the right choice. I liked it because it showed what could be the logs in its icon but then the + icon makes no sense. We are not adding anything to the logs. We are saving/copying.

If you have a better suggestion let me know.

PalmDevs:
Best take is to design our own icon.

validcube (approved):

BenjaminHalko:
Ok after doing some research, intercepting the screenshot command is only available on Android 14+. But I can run the copy logs function when a screenshot is triggered, the only problem is that a photo is still saved.

I can also disable screenshots during the patching stage by setting the window to secure mode, but then I can't detect when the user screenshots.

There is also a possibility that I did not look hard enough...

BenjaminHalko:
Does anyone have any ideas of what we could do instead given that we can stop them from saving a photo?

oSumAtrIX:
Iirc Telegram can prevent screenshots and show a warning toast, so that may be one source of reference

BenjaminHalko:
Thank you! I will check.

BenjaminHalko:
Unless you mean it shows a warning such as "This screen is secure", because that is the default when windows are set to secure.

oSumAtrIX:
I just checked, and what it does is black out the screenshot, I am not sure how it does though.

BenjaminHalko:
What version of android are you using?

BenjaminHalko:
Which version of Android are you using?

BenjaminHalko:
It might be that Telegram detects when a screenshot is taken then goes to your screenshot location and modifies it afterwards.

I did more research and it is looking like the only way to fully replace the screenshot command is with Android 14. Since they added this as a new feature, I feel like it was previously impossible to do otherwise, why would they add it. That being said I might still do one last bit of research just to double confirm it's impossible.

 But I can still detect when a screenshot is taken. Maybe something like a toast saying "Instead of taking a screenshot, please use the copy button to share logs".

oSumAtrIX:
> It might be that Telegram detects when a screenshot is taken then goes to your screenshot location and modifies it afterwards.

Yes, I also have heard about this approach (and I use Android 13 by the way).
If that is the case, how come the screenshot itself is blank? I'd assume, deleting it wouldn't cause the screenshot to be black.

BenjaminHalko:
The other approach I saw was that it could blank the screen while the screenshot is happening.

BenjaminHalko:
I have temporarily added screenshot restrictions to the patcher screen, can you test it out and let me know what happens? (I use Android 12 and it just shows a security policy toast)

oSumAtrIX:
This sounds like an approach that could always work. This can be done with ReVanced Manager as well. The popup as previously explained should then be shown that explains that the screenshot has been prevented and instead the log has been copied to the clipboard.

It may also be a good idea, to allow a second attempt to screenshot. The first attempt is highly likely done because the user either did not know they could copy to a clipboard, a screenshot seemed easier to them, or the log is assumed to be too long or inconvenient to send. After blocking the attempt and instead copying to the clipboard, the popup can additionally explain, that a second attempt to screenshot will not be blocked in case the user insists on screenshotting for any reasonable case.

BenjaminHalko:
Yeah that sounds like a great idea

But only if I can override the screenshot command. Even when I disable screenshot functionality using the secure flag I do not know if I can control what pops up.

And if I block screenshot functionality then the screenshot detector won't get triggered, at least the one I have been using anyways...

oSumAtrIX:
> But only if I can override the screenshot command. Even when I disable screenshot functionality using the secure flag I do not know if I can control what pops up.

Sorry, I was replying to the comment that was mentioning blanking the screen.

BenjaminHalko:
Oh ok, yeah that sounds great. I might need to add a permission to the manager to view user pictures though, but I do not know.

oSumAtrIX:
I don't think this permission is strictly necessary. See https://github.com/ReVanced/revanced-manager/pull/1299#issuecomment-1732458715:
> even when denying access to images, the screenshot was blank.

BenjaminHalko:
Oh ok, yeah that sounds like a great idea

BenjaminHalko:
It seems like they might just be blanking the screen instead of editing the screenshot then.

oSumAtrIX:
Yes, the handler to blank the screen is called before the screenshot is taken. If this works, the blank screen can be replaced with a view with text saying why the screenshot was blocked (e.g. "The screenshot was blocked because it may not contain the full log and the log has been copied to your clipboard")

BenjaminHalko:
@oSumAtrIX Can you try taking a screenshot in the patcher screen, I would like to see how your phone handles it. [app-debug.apk](https://github.com/ReVanced/revanced-manager/suites/16494277140/artifacts/942331054)

oSumAtrIX:
It worked well. In the patcher screen, when I took a screenshot, a black screen was saved. Additionally the log was copied to clipboard and a toast was shown. A small bug occured when I minimized the app and came back to it, it showed a black screen

oSumAtrIX:
https://github.com/ReVanced/revanced-manager/assets/13122796/bbe93543-92c6-48b3-a54b-b33d116ffa67

oSumAtrIX:
Screen recording causes it to completely blank out. This only is visible in the recording.

BenjaminHalko:
Right ok so it depends on the Android version. Mine just shows a toast and no screenshot.

BenjaminHalko:
This makes it slightly harder to detect screenshots.

BenjaminHalko:
This makes it slightly harder to detect screenshots. The library I am currently using seems to detect if the screenshot overlay is visible to detect screenshots. So if no screenshot is taken it won't trigger.

oSumAtrIX:
It seems to also block recordings completely. As previously explained recording is often used for debugging, which is why it would rather be better to allow recording, but block the first attempt to screenshot

BenjaminHalko:
I do not know if I can block one but not the other, right now I am setting the screen to "secure" which blocks both screenshots and recordings. Though as previously mentioned the way it blocks is different from version to version. I am currently looking at other methods and I will keep that in mind.

BenjaminHalko:
It might not be possible to reliably stop the user from taking a screenshot while also detecting that the user attempted to take a screenshot.

Is there a different way to tell the user that copied logs are preferred? Maybe when they take a screenshot a popup (not a toast) could come up saying `We noticed you took a screenshot. If you would like to share logs, why not use the copy button instead.`

BenjaminHalko:
Unfortunately there is nothing that can detect when a screenshot is currently taking place, only afterwards.

oSumAtrIX:
Does this also apply to Telegram when the permission for pictures is not present?

BenjaminHalko:
Yes because telegram sets the `secure` flag on the app window to true. This is also what I used in the latest commit, but unfortunately setting the flag has varied results on different devices. Yours takes a black screenshot but mine just doesn't take a screenshot at all. While I can detect when a black screenshot is taken I can tell when a screenshot is attempted on a device like mine.

This approach of setting the secure flag also disables taking videos. There is no way to only disable screenshots (except in Android 14)

oSumAtrIX:
Alright then, in this case we can do the following:

Do not prevent screenshots. Instead when a screenshot is taken, show the popup with the notice about the log being copied to clipboard and that it should be sent instead of the screenshot.

BenjaminHalko:
After the first screenshot is taken, do you want me to do nothing on subsequent screenshots?

oSumAtrIX:
I see, in this case it is fine to simply display the popup at the initial attempt of screenshotting.

BenjaminHalko:
How's this?

Screenshot_20230925-104248_ReVanced Manager Debug

oSumAtrIX:
Looks good

BenjaminHalko:
Ok so additional complication, it seems like the screenshot is added to your clipboard when the Android screenshot overlay disappears, so while I can save the logs to the clipboard, doing it automatically will not place the log at the top of the clipboard since the Android screenshot overlay stays visible for around 3 seconds unless the screen is tapped.

My idea is instead to change the message at the end to say "Would you like to copy the logs to your clipboard?" and then have yes / no buttons, this would then allow the screenshot overlay to disappear because of the screen tap.

BenjaminHalko:
Screenshot_20230925-110825_ReVanced Manager Debug

BenjaminHalko:
I think this is ready now

oSumAtrIX:
Is this ready for review?

BenjaminHalko:
Yep

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -338,10 +415,6 @@ class InstallerViewModel extends BaseViewModel {
     }
   }
 
-  void exportLog() {
-    _patcherAPI.exportPatcherLog(logs);
-  }
-
```
Ushie, line 344: Considering this makes exportPatcherLog unused, maybe it should be deleted?

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -169,6 +180,72 @@ class InstallerViewModel extends BaseViewModel {
...
+
+  Future<void> screenshotDetected(BuildContext context) async {
+    await showDialog(
+      context: context,
+      builder: (context) => AlertDialog(
+        title: I18nText(
+          'warning',
+        ),
+        backgroundColor: Theme.of(context).colorScheme.secondaryContainer,
+        icon: const Icon(FontAwesomeIcons.triangleExclamation),
```
Ushie, line 224: You can use Material Icons instead of FontAwesomeIcons

Ushie (changes requested):

BenjaminHalko:
@Ushie done

lib/ui/views/installer/installer_viewmodel.dart
Ushie: This was marked as resolved, but I don't see any changes about this

BenjaminHalko: You're suggesting that `exportLog` be removed right? Well it has been removed, even before you requested the change.

BenjaminHalko: Oh wait... I'm dumb. It was `_patcherAPI.exportPatcherLog` you were wanting removed. My bad

BenjaminHalko: Sorry about that.

BenjaminHalko: I actually think it should be left in, because there was some talk about maybe in the future that exporting to file might return if it is required.

Ushie: I see, fine by me

BenjaminHalko:
Is this ready to merge now?

Ushie (approved):

assets/i18n/en_US.json
```diff
@@ -162,6 +162,9 @@
         "exportApkButtonTooltip": "Export patched APK",
         "exportLogButtonTooltip": "Export log",
 
+        "screenshotDetected": "We noticed you took a screenshot. If you were trying to share the log, please share a text copy instead of a screenshot.\n\nWould you like to copy the log to the clipboard?",
```
oSumAtrIX, line 165:
```suggestion
        "screenshotDetected": "A screenshot has been detected. If you are trying to share the log, please share a text copy instead.\n\nCopy log to clipboard?",
```

assets/i18n/en_US.json
```diff
@@ -162,6 +162,9 @@
...
+        "copiedToClipboard": "Copied logs to clipboard",
```
oSumAtrIX, line 166:
```suggestion
        "copiedToClipboard": "Copied log to clipboard",
```

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -38,6 +41,7 @@ class InstallerViewModel extends BaseViewModel {
   bool hasErrors = false;
   bool isCanceled = false;
   bool cancel = false;
+  bool doNotPopupScreenshotWarning = false;
```
oSumAtrIX, line 44: Rename this to showScreenshotWarning and negate the boolean logic. The code has double negations which make it confusing to understand.

oSumAtrIX (changes requested):
Looks good to me. One last way to improve it would be to automatically copy the log to the clip when the screenshot is detected. This way no additional user input is necessary and it may also overwrite the screenshot from clipboard

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -169,6 +179,72 @@ class InstallerViewModel extends BaseViewModel {
...
+            label: I18nText('noButton'),
+            onPressed: () {
+              Navigator.of(context).pop();
+            },
+          ),
+          CustomMaterialButton(
+            label: I18nText('yesButton'),
+            onPressed: () {
+              copyLogs();
+              showPopupScreenshotWarning = true;
```
oSumAtrIX, line 239: Is this intended? After the screenshot event handler was called, it should not be called for the current installer session anymore, so that the user can copy a screenshot without any intervention.

BenjaminHalko: It was intended because if the user wants to copy logs then receives another error,  they know that screenshoting will open the dialog to copy logs, so that is what they do. They may not know that there is a dedicated button outside the dialog so I programmed it this way to prevent confusion. If the user does not want the popup popping up, they can simply press no.

BenjaminHalko:
@oSumAtrIX [this](https://github.com/ReVanced/revanced-manager/pull/1299#issuecomment-1734232750) explains why auto copying is not done.

oSumAtrIX:
I see, thanks.

oSumAtrIX (approved):

BenjaminHalko:
Is this ready to merge?

Ushie:
LGTM, is there anything you want to add before I merge this? @oSumAtrIX

oSumAtrIX:
lgtm

PR #1303, opened 2023-09-22
fix: trigger workflow on all files in app

BenjaminHalko:
The workflow does not trigger on things like string updates or manifest updates, so now the range has been broadened to anything in the app directory.

I also made it check the whole gradle directory too but that is less important...


BenjaminHalko:
nevermind, the workflow I was basing this off of just took a while to load...

PR #1308, opened 2023-09-24
feat: settings migration (flutter)

BenjaminHalko:
Allows compose to import flutter settings.

~~The provider `content://app.revanced.manager.flutter.provider/settings` is created so that the compose manager can query it.~~

Creates an activity that compose can access.


BenjaminHalko:
I think the Flutter side of settings migration is done, but I will leave this as a draft until the compose side is done.

Currently it exports all of shared preferences, saved patches, and the keystore (in base64)

PalmDevs (commented):
Does this mean any app can access this URI and grab all settings? Ideally Flutter Manager should include an activity which shows a dialog asking the user if they want to import should be in Flutter Manager, so it can ask the user before giving any data to any other app.

BenjaminHalko:
Instead of accessing `content://app.revanced.manager.flutter.provider/settings`, compose can now start the ExportSettings activity which will show this dialog:
Screenshot_20230926-162052_ReVanced Manager Debug

BenjaminHalko:
One thing that still needs updating is changing the color of the top and bottom to match the rest of the background

PalmDevs (approved):
Looks good.

Can you make the background at least transparent so you can see Compose Manager in the background or something? So the user can think the dialog comes from Compose Manager and it's actually "seamless".

validcube (approved):
🥞 the code and the UI looks good to me.

oSumAtrIX (commented):
Two questions:

- Does this approach prevent any app with a different package name other than that of Compose version to access the settings?
- Does this approach prevent any app with a different signature other than that of the Compose version to access the settings?

BenjaminHalko:
Nope to both. BUT in order to access settings they need to press yes on a dialog, so it will seem weird if a different app opens the dialog.

Screenshot_20230927-123522_ReVanced Manager Debug

oSumAtrIX:
This only mitigates hijacking the settings as it now relies on the user to do the verification. You can automate the verification using this method:

https://github.com/ReVanced/revanced-manager/pull/1309#issuecomment-1737983076

BenjaminHalko:
@oSumAtrIX done! I got the fingerprint from the GitHub release of flutter manager, that will be the same fingerprint that the compose manager has right?

BenjaminHalko:
@oSumAtrIX done. Now it will check to see if the fingerprint of the calling activity is the same as its own

oSumAtrIX:
@Ushie Does the compose app use the same signature?

android/app/src/main/kotlin/app/revanced/manager/flutter/ExportSettingsActivity.kt
```diff
@@ -0,0 +1,87 @@
...
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        val callingPackageName = getCallingPackage()!!
+
+        if (getFingerprint(callingPackageName) == getFingerprint(getPackageName())) {
+            // Create JSON Object
+            val json = JSONObject()
+
+            // Default Data
+            json.put("keystorePassword", "s3cur3p@ssw0rd")
```
oSumAtrIX, line 27: What if I changed the password?

Ushie:
likely, but not set in stone, if we for whatever reason decide otherwise we can change the signature here again

android/app/src/main/kotlin/app/revanced/manager/flutter/ExportSettingsActivity.kt
Ushie: @BenjaminHalko I believe there's an variable in settings for the keystore password, it can be used here

BenjaminHalko: Then it will get changed in when it loops through all the settings in shared preferences. This is in case the user doesn't change the password and an entry for the password is not in shared preferences. All other settings use the Compose default if the entry is not in Flutters shared preferences but since the password is required for importing the keystore, I put the default password in just in case Compose every decides to change its default password.

BenjaminHalko: @Ushie wouldn't that require loading Flutter? This activity doesn't use flutter as it is unnecessary.

oSumAtrIX (approved):

PR #1309, opened 2023-09-24
feat: settings migration (compose)

BenjaminHalko:
Allows compose to import flutter settings.

~~On the first run it will detect if the flutter manager is present on the system, and then will query `content://app.revanced.manager.flutter.provider/settings`.~~

Edit: It not longer uses `content://app.revanced.manager.flutter.provider/settings` instead it launches an Activity from Flutter.


BenjaminHalko:
Screenshot_20230925-134153_ReVanced Manager Debug

PalmDevs (approved):
UI looks good to me.

Axelen123 (changes requested):

BenjaminHalko:
Now instead of showing the "Would you like to import legacy settings" dialog on Compose, it will start an export settings activity on Flutter.

Ushie:
What about a popup offering to uninstall ReVanced Manager flutter after importing?

oSumAtrIX:
Ideally, the update happens seamlessly without any interruption. This means that once an update is offered in ReVanced Manager and the user updates, ReVanced Manager Compose imports the settings and uninstalls Flutter if it detects it. After the migration is done, it will disable this functionality in case you ever want to go back. This will feel and look like a usual update without any user interaction.

Ushie:
Is that possible without any user intervention?

oSumAtrIX:
@BenjaminHalko Exposing a content:// URL allows any app on the device to hijack the keystore password or other sensitive data. Is it possible to check the signature of the app that tries to read the settings?

@Ushie  Uninstalling will probably trigger the Android uninstall popup. To not confuse the user after the update. Once the new app is launched, it can show a popup clarifying that ReVanced Manager has been updated and the old one should be installed. Once the user presses next/ok, uninstallation can be triggered. Other than that everything else should be possible to automate.

BenjaminHalko:
@oSumAtrIX It not longer uses content://app.revanced.manager.flutter.provider/settings instead it launches an Activity from Flutter.

PalmDevs:
> @BenjaminHalko Exposing a content:// URL allows any app on the device to hijack the keystore password or other sensitive data. Is it possible to check the signature of the app that tries to read the settings?

This is no longer the case anymore. Now there's an activity on Flutter's side which prompts the user. However, if you want a 0-interaction importing, we can possibly have it if both sides can provide an activity to verify each other.

```
🤮 = Flutter (import activity)
😄 = Any app
🔥 = Compose (verify import activity)

😄 calls 🤮 with randomly generated token
🤮 calls 🔥 with the token 😄 gave

If 🔥 doesn't exist or responds with a no:
- 🤮 responds back, but doesn't give any settings data
else:
- 🤮 responds back to 😄 with settings data
```

oSumAtrIX:
Flutter needs a whitelist of the signature of Compose manager. It checks the callers signature and verifies if it's legitimate. This is similar to GMS verifying signatures

BenjaminHalko:
> However, if you want a 0-interaction importing, we can possibly have it if both sides can provide an activity to verify each other.

Well we want some interaction, what if the user does not want to import their settings

oSumAtrIX:
No interaction should be present. This update should look to the user like a regular update with no intervention. This means the user presses update, is prompted to install the app, compose initiates the migration phase with the least amount of Intervention necessary. One Intervention will be necessary as a restrictioncon Android to uninstall the old package. A popup will explain why and the user is given the option to keep the old version.

BenjaminHalko:
@oSumAtrIX so the migration process should be something like this?

- Compose requests from Flutter
- Flutter checks signature
- Flutter sends settings

oSumAtrIX:
Yes.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -30,4 +42,85 @@ class MainViewModel(
...
+    }
+
+    fun applyLegacySettings(data: String) = viewModelScope.launch {
+        prefs.showAutoUpdatesDialog.update(false)
+        val jsonData = JSONObject(data)
+        val keysIterator = jsonData.keys()
+        while (keysIterator.hasNext()) {
+            try {
+                val key = keysIterator.next()
+                val value = jsonData.get(key)
```
Axelen123, line 67: `kotlinx.serialization` should be used instead because the rest of the app uses it

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
oSumAtrIX: The app does not use `kotlinx.serialization` at all though.

Axelen123: No, it does use the library: https://github.com/ReVanced/revanced-manager/tree/compose-dev/app/src/main/java/app/revanced/manager/network/dto

oSumAtrIX: Oh, I am sorry I was mistaking this PR with mine

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -30,4 +42,85 @@ class MainViewModel(
...
+            )
+            launcher.launch(intent)
+            true
+        } catch (e: Exception) {
+            false
+        }
+    }
+
+    fun applyLegacySettings(data: String) = viewModelScope.launch {
+        prefs.showAutoUpdatesDialog.update(false)
```
oSumAtrIX, line 61: This should be moved to the bottom in case an error occurs.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -30,4 +41,70 @@ class MainViewModel(
...
+
+                updateCheck()
+            }
+        }
+        settings.patchesChangeEnabled?.let {
+            // TODO: Implement setting
+        }
+        settings.showPatchesChangeWarning?.let {
+            // TODO: Implement setting
+        }
```
Axelen123, line 96: The selection warning was recently implemented. Can you implement this?

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -30,4 +41,70 @@ class MainViewModel(
...
+        settings.keystore?.let { keystore ->
+            prefs.keystoreCommonName.update("ReVanced")
+            prefs.keystorePass.update(settings.keystorePassword)
+            val keystorePath = app.getDir("signing", Context.MODE_PRIVATE)
+                .resolve("manager.keystore").toPath()
+            val keystoreBytes = Base64.decode(keystore, Base64.DEFAULT)
+            keystorePath.toFile().writeBytes(keystoreBytes)
```
Axelen123, line 103: Try to use `KeystoreManager` to import it instead.
Make sure to handle exceptions.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
BenjaminHalko: How do I get the latest changes from Compose dev onto this branch?

oSumAtrIX: git fetch
git merge compose-dev

or 

git fetch
git rebase compose-dev

app/src/main/java/app/revanced/manager/network/dto/LegacySettings.kt
```diff
@@ -0,0 +1,18 @@
...
+    val themeMode: Int? = null,
+    val useDynamicTheme: Int? = null,
+    val apiUrl: String? = null,
+    val experimentalPatchesEnabled: Int? = null,
+    val patchesAutoUpdate: Int? = null,
+    val patchesChangeEnabled: Int? = null,
+    val showPatchesChangeWarning: Int? = null,
+    val keystore: String? = null,
+    val patches: SerializedSelection? = null,
+)
```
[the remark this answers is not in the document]
BenjaminHalko: Because when loading the json string. If the values were `true` or `false` boolean, it crashed with unexpected character 't'. Maybe I was loading it wrong?

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
Axelen123: Merging like osu said is the easiest way. You might have to specify `origin/compose-dev` instead of `compose-dev` though

app/src/main/java/app/revanced/manager/network/dto/LegacySettings.kt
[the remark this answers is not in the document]
BenjaminHalko: I can try retooling at making books work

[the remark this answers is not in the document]
Axelen123: Strange, how was the json created?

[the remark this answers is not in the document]
BenjaminHalko: The json is created [here](https://github.com/ReVanced/revanced-manager/blob/main/android/app/src/main/kotlin/app/revanced/manager/flutter/ExportSettingsActivity.kt#L24) in Flutter. Notice the Boolean work around.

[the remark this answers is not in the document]
BenjaminHalko: I will try removing this work around and see if I can load it properly

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -30,4 +41,70 @@ class MainViewModel(
...
+            val intent = Intent()
+            intent.setClassName(
+                "app.revanced.manager.flutter",
+                "app.revanced.manager.flutter.ExportSettingsActivity"
+            )
+            launcher.launch(intent)
+            true
+        } catch (e: Exception) {
+            false
+        }
```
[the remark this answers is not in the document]
BenjaminHalko: No because it will fail when the Flutter manager is not installed, which should be the normal for most cases after the migration.

BenjaminHalko:
Should the `LegacySettings` data type file be stored somewhere other than in `network/dto`? It doesn't really have anything to do with networking.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
BenjaminHalko: Should I use the `patchesChangeEnabled` or the `showPatchesChangeWarning` setting from Flutter to control the Compose preference.

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -37,6 +43,20 @@ class MainActivity : ComponentActivity() {
...
+        ) { result: ActivityResult ->
+            if (result.resultCode == Activity.RESULT_OK) {
+                if (result.data != null) {
+                    val jsonData = result.data!!.getStringExtra("data")!!
+                    vm.applyLegacySettings(jsonData)
+                }
+            } else {
+                launchedActivityState.value = 2
+            }
+        }
```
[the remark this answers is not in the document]
BenjaminHalko: Where should the enum be initialized?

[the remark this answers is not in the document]
Axelen123: It can be a private type in the file it is used in

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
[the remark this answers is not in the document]
Axelen123: I see.
It should be possible to check the type of the exception to determined if the app is installed or not though. I believe PM already does this

Axelen123:
> Should the `LegacySettings` data type file be stored somewhere other than in `network/dto`? It doesn't really have anything to do with networking.

Yeah it doesnt really belong there. Maybe just put it in the viewmodel or activity file and make it private

BenjaminHalko:
@Axelen123 All done

Axelen123:
Still need to switch to `rememberLauncherForActivityResult`

BenjaminHalko:
@Axelen123 done

oSumAtrIX (approved):

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -52,8 +74,17 @@ class MainActivity : ComponentActivity() {
                 NavBackHandler(navController)
 
                 val showAutoUpdatesDialog by vm.prefs.showAutoUpdatesDialog.getAsState()
+
                 if (showAutoUpdatesDialog) {
-                    AutoUpdatesDialog(vm::applyAutoUpdatePrefs)
+                    rememberLauncherForActivityResult = rememberSaveable { mutableStateOf(LegacyActivity.NOT_LAUNCHED) }
```
Axelen123, line 79: I was referring to this: https://developer.android.com/reference/kotlin/androidx/activity/compose/package-summary#rememberLauncherForActivityResult(androidx.activity.result.contract.ActivityResultContract,kotlin.Function1)

Axelen123 (changes requested):

BenjaminHalko:
@Axelen123 done

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -51,9 +60,45 @@ class MainActivity : ComponentActivity() {
...
                 NavBackHandler(navController)
 
-                val showAutoUpdatesDialog by vm.prefs.showAutoUpdatesDialog.getAsState()
-                if (showAutoUpdatesDialog) {
-                    AutoUpdatesDialog(vm::applyAutoUpdatePrefs)
+                val firstLaunch by vm.prefs.firstLaunch.getAsState()
+
+                if (firstLaunch) {
+                    val showAutoUpdatesDialog = rememberSaveable { mutableStateOf(false) }
+                    if (!showAutoUpdatesDialog.value) {
```
Axelen123, line 67: Use the `by` keyword.

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -51,9 +60,45 @@ class MainActivity : ComponentActivity() {
...
+                                "app.revanced.manager.flutter.ExportSettingsActivity"
+                            )
+                        }
+
+                        LaunchedEffect(Unit) {
+                            try {
+                                launcher.launch(intent)
+                            } catch (e: Exception) {
+                                if (e !is ActivityNotFoundException) {
+                                    toast(getString(R.string.legacy_import_failed))
```
Axelen123, line 94: Log the error with `Log.e` as well

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -51,9 +60,45 @@ class MainActivity : ComponentActivity() {
...
+                        val intent = Intent().apply {
+                            setClassName(
+                                "app.revanced.manager.flutter",
+                                "app.revanced.manager.flutter.ExportSettingsActivity"
+                            )
+                        }
+
+                        LaunchedEffect(Unit) {
+                            try {
+                                launcher.launch(intent)
```
Axelen123, line 91: The `LaunchedEffect` block will run again if you rotate the screen for example. Add a variable to keep track if you have attempted to launch or not if this is not intended

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -1,21 +1,42 @@
...
+    fun launchActivity(launcher: ManagedActivityResultLauncher<Intent, ActivityResult>) = viewModelScope.launch {
+        val intent = Intent().apply {
+            setClassName(
+                "app.revanced.manager.flutter",
+                "app.revanced.manager.flutter.ExportSettingsActivity"
+            )
+        }
+
+        launcher.launch(intent)
+    }
```
Axelen123, line 36: This does not appear to be used

Axelen123 (changes requested):

BenjaminHalko:
@Axelen123 Fixed, my bad.

BenjaminHalko:
@Axelen123 done... again.

Axelen123 (approved):

PR #1311, opened 2023-09-26
build: Bump dependencies

oSumAtrIX:
# TODO

- [x] Test patches (and integrations) from `dev` branch
- [x] Check if the removed code should not be removed
- [x] Check if the removal of `path.dependencies` breaks anything
- [x] Resolve the todos 
- [x] Check if installation works and if the app is properly added to the "Patched apps" list
- [x] Monitor network traffic for any unusual behavior


android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
```diff
@@ -93,29 +88,34 @@ class MainActivity : FlutterActivity() {
...
+                    ).map { patch ->
+                        buildMap<String, Any> {
+                            this["\"name\""] =
+                                "\"${patch.name!!.replace("\"", "\\\"")}\""
+
+                            this["\"description\""] =
+                                "\"${patch.description?.replace("\"", "\\\"")}\""
+
+                            this["\"excluded\""] =
+                                !patch.use
```
TheAabedKhan, line 106:
```suggestion
                            this["\"excluded\""] =
                                !patch.use
                            this["\"dependencies\""] =
                                patch.dependencies?.map { "\"${it.java.name}\"" }
                                    ?: emptyList<Any>()
```

Need `dependencies` as well

TheAabedKhan (changes requested):

android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
oSumAtrIX: Why?

TheAabedKhan: It's in use

TheAabedKhan: https://github.com/ReVanced/revanced-manager/blob/58385501881b2d606e4cad925654211c5305d56e/lib/services/patcher_api.dart#L162

here

oSumAtrIX: What use does it have?

Ushie: However the needsSettingsPatch function may be unused/unnecessary now

TheAabedKhan: https://github.com/ReVanced/revanced-manager/blob/58385501881b2d606e4cad925654211c5305d56e/lib/services/patcher_api.dart#L184

To check for dependencies and include them while patching

oSumAtrIX: I am not sure what you mean by "ability to patch from storage"

Ushie: I meant from installed, sorry

Ushie: Patching from installed will work fine if no resource patching is necessary, this is the only way to tell if resource patching is necessary

oSumAtrIX: It will only work fine, if you mount. And if you mount, resource patching will work fine anyways. Not sure why the dependencies are necessary. Also how do you check on flutter side if it's even a resource patch when you only get strings?

oSumAtrIX:
@TheAabedKhan I have updated this branch with TODOs in the PR description.

android/app/build.gradle
```diff
@@ -85,7 +85,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
     // ReVanced
-    implementation "app.revanced:revanced-patcher:14.2.2"
+    implementation "app.revanced:revanced-patcher:15.0.1"
```
oSumAtrIX, line 88:
```suggestion
    implementation "app.revanced:revanced-patcher:15.0.2"
```

Ushie:
Are you going to bring back original package name now that you've updated the code to read it from the input and output APKs?

oSumAtrIX:
No, the "original package name" is of no relevance at any point in time.

Ushie (commented):
What is the motivation behind this? Mount may not be as recognized as root, this popup is only shown to those who have given root permission to ReVanced Manager so Root and Non-root can safely be assumed to be more recognizable

oSumAtrIX:
@Ushie We install by mounting the app using root. Just like the CLI's option is not called `--root` but `--mount`.  So basically root is required to be able to install the app by mounting. Referring to the installation type with "install by root " is wrong, but "install by mount" is correct. If the terms are not explanatory, a subtitle can be added:

Normal: Install the patched app using Android package manager. This is the standard installation type as you know it from installing apps manually. **Learn more**

Mount: Install the patched app by mounting it over an existing installation of the unpatched app. The version of the unpatched app has to match that of the patched app at all time. **Learn more**

android/app/build.gradle
```diff
@@ -85,7 +85,7 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
 
     // ReVanced
-    implementation "app.revanced:revanced-patcher:14.2.2"
+    implementation "app.revanced:revanced-patcher:15.0.2"
```
oSumAtrIX, line 88:
```suggestion
    implementation "app.revanced:revanced-patcher:15.0.3"
```

Issue #1335, opened 2023-10-05
bug: Only app name is searched for if app is installed 


KobeW50:
### Bug description

If an app is installed and you try searching for it in the apps list within the Manager, you can only search for the app name and not the package name.

https://github.com/ReVanced/revanced-manager/assets/84587632/5af3af7c-f9df-49d1-bdf8-cd329594c8bf

This PR may be related https://github.com/ReVanced/revanced-manager/issues/1330

### Version of ReVanced Manager and version & name of application you tried to patch

1.11.1

### Installation type

None

### Device logs

```shell
NA
```

### Acknowledgements

- [X] This request is not a duplicate of an existing issue.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The issue is solely related to the ReVanced Manager

PR #1344, opened 2023-10-07
fix: search with package name

Almostsagar:
fixed : #1335


validcube (approved):
🥞 Tested and Approved

Thanks for your contribution!

Issue #1331, opened 2023-10-05
Selected patches are in strange order


KobeW50:
### Bug description

The selected patches in the card are seemingly in a random order at first. And if you select new patches the most recently selected patches will be at the bottom of the list.

For example, this is the default order for the default YT Music patches.

![SmartSelect_20231005_041125_ReVanced Manager](https://github.com/ReVanced/revanced-manager/assets/84587632/151bc54d-9d24-459c-b054-8ed1cee38277)

They should always be ordered alphabetically just like the patches selection menu list.

### Version of ReVanced Manager and version & name of application you tried to patch

1.11.0

### Installation type

None

### Device logs

```shell
NA
```

### Acknowledgements

- [X] This request is not a duplicate of an existing issue.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The issue is solely related to the ReVanced Manager

validcube:
👋 Hello! Thank for reporting this issue, 

Your issue has been assigned, and will be fix very shortly.

validcube:
Nevermind, ignore that, turn out GitHub didn't give me a notification.

> 👋 Hello! Thank for reporting this issue,
> 
> Your issue has been assigned, and will be fixed very shortly.
> 
> ~wait, what's the point of saying this I've already finished~

PR #1345, opened 2023-10-08
fix: selected patches order

mddanishansari:
fixed: #1331

### DIFF
--- a/lib/ui/widgets/patcherView/patch_selector_card.dart
+++ b/lib/ui/widgets/patcherView/patch_selector_card.dart
@@ -58,7 +58,9 @@ class PatchSelectorCard extends StatelessWidget {
 
   String _getPatchesSelection() {
     String text = '';
-    for (final Patch p in locator<PatcherViewModel>().selectedPatches) {
+    final List<Patch> selectedPatches = locator<PatcherViewModel>().selectedPatches;
+    selectedPatches.sort((a, b) => a.name.compareTo(b.name));
+    for (final Patch p in selectedPatches) {
       text += '\u2022  ${p.getSimpleName()}\n';
     }
     return text.substring(0, text.length - 1);


lib/ui/widgets/patcherView/patch_selector_card.dart
```diff
@@ -58,7 +59,9 @@ class PatchSelectorCard extends StatelessWidget {
 
   String _getPatchesSelection() {
     String text = '';
-    for (final Patch p in locator<PatcherViewModel>().selectedPatches) {
+    for (final Patch p in locator<PatcherViewModel>()
+        .selectedPatches
```
oSumAtrIX, line 63: Why not sort when initializing `selectedPatches`?

Ushie: Why not sort when generating the json overall? that results in it being sorted everywhere, patch selector screen, patchcard, etc

mddanishansari: I tried that first, but it fails in this scenario.


https://github.com/ReVanced/revanced-manager/assets/11080523/c017d67d-7dec-467b-a663-3669e5e14e1b

BTW, I had written sort logic earlier in `loadLastSelectedPatches()` function of `PatcherViewModel` just above `notifyListeners()`. Is that the correct place?

lib/ui/widgets/patcherView/patch_selector_card.dart
```diff
@@ -1,3 +1,4 @@
+import 'package:collection/collection.dart';
```
validcube, line 1:
```suggestion
```

I don't think this package is related to sorting, please remove them or mark this as resolve.

👋 Anyway, thank for your contribution!

lib/ui/widgets/patcherView/patch_selector_card.dart
mddanishansari: > Why not sort when generating the json overall? that results in it being sorted everywhere, patch selector screen, patchcard, etc

List is being sorted only for UI purpose and nothing else. That's why I think sorting it here in the view is correct. What do you think?

Ushie: In that case, I think your placement fits best

oSumAtrIX (approved):

validcube (approved):
🥞 Approved, thanks for your contribution!

PR #1347, opened 2023-10-08
style: sort selected patches "lexicographically"

validcube:


validcube:
Ok, turn out, I'm late see #1345 instead

PR #1354, opened 2023-10-10
feat: Add patch options

TheAabedKhan:
patchOptions


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,5 @@
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
```
validcube, line 5: Why? We already have one in `~/android/gradle/wrapper`.

But if this was intended

You might want to use the same Gradle version as `~/android/gradle/wrapper` (7.5) to avoid downloading two different versions of Gradle for the same project.

oSumAtrIX: So the thing is, the root project has to be a gradle project to be able to use the subgrade project. I removed the wrapper from the subproject and added it to the root project. Is this a problem?

oSumAtrIX (approved):

PR #1369, opened 2023-10-13
refactor(Accessibility): improve patch options

validcube:
## Tooltip

Improve accessibility, and avoid user from accidentally resetting their patch options

image

## Colour

Changed from Secondary Container to Surface to provide slightly better hierarchy.

image

Ignore commit message, something messed up

Yes, my budget for writing PR has temporarily dropped :trollface:

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -1,6 +1,7 @@
 {
     "okButton": "OK",
     "cancelButton": "Cancel",
+    "dismissButton": "Dismiss",
     "quitButton": "Quit",
     "updateButton": "Update",
     "enabledLabel": "Enabled",
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -136,6 +136,7 @@
         "selectAllPatchesWarningContent": "You are about to select all patches, that includes non-suggested patches and can cause unwanted behavior."
     },
     "patchOptionsView": {
+        "resetOptionsTooltip": "Reset options",
         "viewTitle": "Patch options",
         "saveOptions": "Save",
 
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -137,7 +137,7 @@
         "selectAllPatchesWarningContent": "You are about to select all patches, that includes non-suggested patches and can cause unwanted behavior."
     },
     "patchOptionsView": {
-        "resetOptionsTooltip": "Reset options",
+        "resetOptionsTooltip": "Reset patch options",
         "viewTitle": "Patch options",
         "saveOptions": "Save",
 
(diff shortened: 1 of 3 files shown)


PalmDevs:
Change `Dismiss` to `Cancel`

assets/i18n/en_US.json
```diff
@@ -1,6 +1,7 @@
 {
     "okButton": "OK",
     "cancelButton": "Cancel",
+    "dismissButton": "Dismiss",
```
PalmDevs, line 4: Change this to `Cancel` instead.

assets/i18n/en_US.json
```diff
@@ -136,6 +137,7 @@
         "selectAllPatchesWarningContent": "You are about to select all patches, that includes non-suggested patches and can cause unwanted behavior."
     },
     "patchOptionsView": {
+        "resetOptionsTooltip": "Reset options",
```
PalmDevs, line 140: `Reset patch options`

PalmDevs (dismissed):
Needs string changes.

---

Another problem with this page is the dialog. Currently, the "list item(s)" don't look like list items at all. It doesn't really tell the user it's clickable and is a choice.

What we could do is revert to the default list item styles and add a **dummy** radio button which will do absolutely nothing other than convince the user the list item is clickable like in the image below. 

The behavior won't change, clicking on the list item won't make it selected, it will add the clicked option automatically.

Image example

PalmDevs:
@oSumAtrIX What do you think about the solution I've proposed in the review?

assets/i18n/en_US.json
validcube: 🔧 Resolved @ 00e49dba0cdcc71e07edad417800a6e2d4f1a4de

assets/i18n/en_US.json
validcube: 🔧 Resolved @ 00e49dba0cdcc71e07edad417800a6e2d4f1a4de

validcube:
> Needs string changes.
> 
> Another problem with this page is the dialog. Currently, the "list item(s)" don't look like list items at all. It doesn't really tell the user it's clickable and is a choice.
> 
> What we could do is revert to the default list item styles and add a **dummy** radio button which will do absolutely nothing other than convince the user the list item is clickable like in the image below.
> 
> The behavior won't change, clicking on the list item won't make it selected, it will add the clicked option automatically.
> 
> Image example

👋 Hello!

Unfortunately, this is out-of-scope for this PR as this is targeted to improvement to Android TalkBack, screen reader, viewing contrasts, accessibility etc.

Feel like to open an issue about this.

PalmDevs (approved):
LGTM

Ushie (approved):

PR #1376, opened 2023-10-14
fix(app-bar): title not hiding completely

TheAabedKhan:
| Before | After |
|----|-----|
before-appbar|after-appbar|


validcube (approved):
💫 Approved

PR #1402, opened 2023-10-17
fix: specify `multithreadingDexFileWriter` in `PatcherOptions`

BenjaminHalko:
`PatcherOptions` was complaining that calling it without `multithreadingDexFileWriter` was depreciated. So `multithreadingDexFileWriter` is now specified as false as that was what the default was before.


app/src/main/java/app/revanced/manager/patcher/Session.kt
```diff
@@ -30,7 +30,8 @@ class Session(
             inputFile = input,
             resourceCachePath = tempDir.resolve("aapt-resources"),
             frameworkFileDirectory = frameworkDir,
-            aaptBinaryPath = aaptPath
+            aaptBinaryPath = aaptPath,
+            multithreadingDexFileWriter = false
```
oSumAtrIX, line 34: This option may be configurable as multithreading can greatly improve speed but cause failure while writing dex files.

oSumAtrIX (approved):

app/src/main/java/app/revanced/manager/patcher/Session.kt
BenjaminHalko: Should I do that in this pr, or later.

oSumAtrIX: You can do it later if it suits you better but this PR is not blocking so can also be done in this one

Axelen123: I believe this should be set to true by default because multithreading was always enabled in previous releases.

oSumAtrIX: This will keep patching speed like before, but it will consume significantly more RAM, it can be possible to make this parameter an integer for thread count.

Axelen123: Setting it to true should make it behave the same as before, so I don't think there is a problem with that.

oSumAtrIX: The problem is that the memory issue remains unsolved wheras the parameter has been added to solve it

Axelen123: Most devices don't run into it and not enabling it slows things down for devices that don't need it

oSumAtrIX: Nearly any device runs into memory issues when patching certain apps

Axelen123: Most devices can patch most apps without issue with multithreading enabled, so it should be enabled by default.

oSumAtrIX: All devices being able to patch slightly slower is better than some not at all slightly faster

Axelen123: No, the majority of devices don't need to disable multithreading and some devices may still hit the limit regardless of whether multithreading is disabled or not. All other frontends enable multithreading and there is no reason not to enable it by default.

oSumAtrIX: > No, the majority of devices don't need to disable multithreading

Yes, but some do.

> and some devices may still hit the limit regardless of whether multithreading is disabled or not

Yes, but way less.

> All other frontends enable multithreading and there is no reason not to enable it by default.

The other frontend doing that is ReVanced CLI as it's heap is not limited by Android.

BenjaminHalko: Can multithreading be auto enabled based on how much ram the device has?

oSumAtrIX: You can't preemptively enable multithreading as it is an app by app case

Axelen123: > > No, the majority of devices don't need to disable multithreading
> 
> Yes, but some do.

The default should be optimized for the majority.

> 
> > and some devices may still hit the limit regardless of whether multithreading is disabled or not
> 
> Yes, but way less.

Can you give an example of a device where success is solely determined by this settings?

> 
> > All other frontends enable multithreading and there is no reason not to enable it by default.
> 
> The other frontend doing that is ReVanced CLI as it's heap is not limited by Android.

Flutter manager also does it.

> Can multithreading be auto enabled based on how much ram the device has?

It is possible to get the memory limit, so yeah I think that is possible.

> You can't preemptively enable multithreading as it is an app by app case

If it works for the majority of cases then its valid. Also, you could probably just anticipate memory requirements by checking how many `.dex` files the app has.

oSumAtrIX: > The default should be optimized for the majority.

No, because the issue with multithreading outweighs the minimal benefit of slightly less patching time. It should be toggleable.

> Can you give an example of a device where success is solely determined by this settings?

Yes, a Pixel 8 Pro.

> Flutter manager also does it.

Should be fixed then.

> It is possible to get the memory limit, so yeah I think that is possible.

No: https://github.com/ReVanced/revanced-manager/pull/1402#discussion_r1365751725

> If it works for the majority of cases then its valid

No, because the issue with multithreading outweighs the minimal benefit of slightly less patching time. It should be toggleable.

> Also, you could probably just anticipate memory requirements by checking how many .dex files the app has.

Not within the scope of ReVanced Manager. Also, you can not solely assume by the dex files count.

oSumAtrIX: If it is enabled by default a part of users will encounter an error without knowing that a toggle in the settings can fix it which worsens UX considerably. 

> The issue with multithreading outweighs the minimal benefit of slightly less patching time.

Axelen123: Defaults should be determined based on the scenario at hand. If it is not possible to pick automatically then the default should be whatever works best in the most common cases.

Not doing this would be objectively wrong.

oSumAtrIX: > should be whatever works best in the most common cases.

What works best is what works, at least.

Axelen123: Multithreading works better in most cases so its the better choice.

oSumAtrIX: It causes issues though:

> The issue with multithreading outweighs the minimal benefit of slightly less patching time.

Axelen123: > It causes issues though:
> 
> > The issue with multithreading outweighs the minimal benefit of slightly less patching time.

The toggle solves that.

oSumAtrIX: The user can not know that it will fail for them, meaning they would have to patch, experience the issue, then somehow figure out the existence of the switch, and then turn it on and try patching again. In contrast, when it's off by default, users can either keep it off or turn it on with a warning saying that it may not work for all apps due to memory limitation, which is miles better than impairing patching by default.

oSumAtrIX: If you can, you could try to check for the OOM exception passed via the PatchResult object and, after patching failure, offer to disable multithreading in order to attempt to fix the issue. In this case, this would be a good compromise to enable the multithreading.

Axelen123: Detecting OOMs is an acceptable solution. I will implement an external process runtime sometime after release as well, which will get rid of the memory limit completely.

oSumAtrIX: I think you can simply catch the exception.

oSumAtrIX: @Axelen123 Should this PR be merged until then or do you plan to push to here?

Axelen123: > @Axelen123 Should this PR be merged until then or do you plan to push to here?

A setting should be added first. Its easy to add because we already have UI components specifically for settings.

Axelen123: I have added the setting, so I will merge this PR. Exceptions can be dealt with another time

PR #1420, opened 2023-10-24
fix(copy-logs): log the used values of patch options

TheAabedKhan:


validcube (approved):
🥞 Tested and Approve-y!

Ushie:
Is there no concern towards this leaking tokens and such?

validcube:
> Is there no concern towards this leaking tokens and such?

Ah, I have forgot about that during review.

I'll dismiss my review for now until this can be resolved.

TheAabedKhan:
> Is there no concern towards this leaking tokens and such? 

But this will help us troubleshoot any issues as we can find out whether the issue is caused due to the improper/incorrect use of the tokens.

oSumAtrIX:
I'd say, add a warning label somewhere that says "Could contain sensitive data"

BenjaminHalko:
Could an option be marked as sensitive and then the user is given the option whether or not to include them, or is this too unnecessarily complicated for copying logs.

oSumAtrIX:
If an option is sensitive or not is too abstract to be present in patcher. Otherwise we'd add countlees other nieche properties like this and eventually breaking abstraction. I suggest to generalize. The user ideally understands what they are doing, which is sharing manual input which can be sensitive and should be reviewed before sharing

oSumAtrIX (approved):

PR #1456, opened 2023-11-01
fix: option state crash

BenjaminHalko:
When loading an APK, manager crashes with `java.lang.IllegalStateException: Reading a state that was created after the snapshot was taken or in a snapshot that has not yet been applied`.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
+++ b/app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
@@ -62,7 +62,7 @@ class SelectedAppInfoViewModel(input: Params) : ViewModel(), KoinComponent {
         viewModelScope.launch {
             if (!persistConfiguration) return@launch // TODO: save options for patched apps.
 
-            val packageName = selectedApp.packageName
+            val packageName = selectedApp.packageName // Accessing this from another thread may cause crashes.
             state.value = withContext(Dispatchers.Default) { optionsRepository.getOptions(packageName) }
         }
 
--- a/app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
+++ b/app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
@@ -62,7 +62,9 @@ class SelectedAppInfoViewModel(input: Params) : ViewModel(), KoinComponent {
         viewModelScope.launch(Dispatchers.Default) {
             if (!persistConfiguration) return@launch // TODO: save options for patched apps.
 
-            state.value = optionsRepository.getOptions(selectedApp.packageName)
+            withContext(Dispatchers.Main) {
+                state.value = optionsRepository.getOptions(selectedApp.packageName)
+            }
         }
 
         state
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
```diff
@@ -62,7 +62,9 @@ class SelectedAppInfoViewModel(input: Params) : ViewModel(), KoinComponent {
         viewModelScope.launch(Dispatchers.Default) {
             if (!persistConfiguration) return@launch // TODO: save options for patched apps.
 
-            state.value = optionsRepository.getOptions(selectedApp.packageName)
+            withContext(Dispatchers.Main) {
+                state.value = optionsRepository.getOptions(selectedApp.packageName)
+            }
```
Axelen123, line 67: Thank you for catching this. It would be better to remove the `Dispatchers.Default` parameter from the `launch` call and do something like `state.value = withContext(Dispatchers.Default) { optionsRepository.getOptions() }` instead. This will run the database query on another thread, which was the original goal of the code.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
BenjaminHalko:
```
viewModelScope.launch {
    if (!persistConfiguration) return@launch // TODO: save options for patched apps.

    state.value = withContext(Dispatchers.Default) { optionsRepository.getOptions(selectedApp.packageName) }
}
```
This still causes it to crash. Did I do something wrong?

Axelen123: You did everything correctly. It appears accessing `selectedApp.packageName` in the `Default` dispatcher is causing the issue, since that is also a compose state. Strange behavior by compose. Lets just get rid of the `withContext` instead I guess.

BenjaminHalko:
```
viewModelScope.launch {
    if (!persistConfiguration) return@launch // TODO: save options for patched apps.

    val packageName = selectedApp.packageName
    state.value = withContext(Dispatchers.Default) { optionsRepository.getOptions(packageName) }
}
```

Would this work if we wanted to keep `withContext`?

Axelen123: That would work

app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
```diff
@@ -59,10 +59,11 @@ class SelectedAppInfoViewModel(input: Params) : ViewModel(), KoinComponent {
     var options: Options by savedStateHandle.saveable {
         val state = mutableStateOf<Options>(emptyMap())
 
-        viewModelScope.launch(Dispatchers.Default) {
+        viewModelScope.launch {
             if (!persistConfiguration) return@launch // TODO: save options for patched apps.
 
-            state.value = optionsRepository.getOptions(selectedApp.packageName)
+            val packageName = selectedApp.packageName
```
Axelen123, line 65:
```suggestion
            val packageName = selectedApp.packageName // Accessing this from another thread may cause crashes.
```

PR #1464, opened 2023-11-04
feat(Update Screen): changelogs & handle states

Ushie:
- [x] Need help with intents not working
- [x] Add a parameter option to start downloading right away
- [x] Check metered connection


Axelen123:
Would be good to check if the network is metered as well.

Ushie:
> Would be good to check if the network is metered as well.

Hm, maybe there's no need to

The update doesn't start downloading until the user manually clicks update, the only instance where the screen will auto-download the update is if it's specifically told to which in our case would be the auto-update check, which already has a metered connection check

What do you think?

Axelen123:
> > Would be good to check if the network is metered as well.
> 
> Hm, maybe there's no need to
> 
> The update doesn't start downloading until the user manually clicks update, the only instance where the screen will auto-download the update is if it's specifically told to which in our case would be the auto-update check, which already has a metered connection check
> 
> What do you think?

The manager update checker only checks if there is an available network, not if its metered. This makes sense in my opinion because manager updates are important and the check itself isn't very expensive.

I think displaying a warning and not automatically downloading on metered networks makes the most sense. What do you think about this?

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
```diff
@@ -0,0 +1,197 @@
...
+    Column(verticalArrangement = Arrangement.spacedBy(16.dp)) {
+        Title(
+            when (vm.state) {
+                State.CAN_DOWNLOAD -> R.string.update_available
+                State.DOWNLOADING -> R.string.downloading_manager_update
+                State.CAN_INSTALL -> R.string.ready_to_install_update
+                State.INSTALLING -> R.string.installing_manager_update
+                State.FAILED -> R.string.install_update_manager_failed
+                State.SUCCESS -> R.string.update_completed
+            }
```
Axelen123, line 89: Enum variants can have fields, which could be used here instead of `when`

Ushie:
> > > Would be good to check if the network is metered as well.
> > 
> > 
> > Hm, maybe there's no need to
> > The update doesn't start downloading until the user manually clicks update, the only instance where the screen will auto-download the update is if it's specifically told to which in our case would be the auto-update check, which already has a metered connection check
> > What do you think?
> 
> The manager update checker only checks if there is an available network, not if its metered. This makes sense in my opinion because manager updates are important and the check itself isn't very expensive.
> 
> I think displaying a warning and not automatically downloading on metered networks makes the most sense. What do you think about this?

I'll add the metered connection check then

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
```diff
@@ -0,0 +1,224 @@
...
+                Divider()
+                Changelog(changelog)
+            } ?: Spacer(modifier = Modifier.weight(1f))
+            Buttons(vm, onBackClick)
+        }
+    }
+}
+
+@Composable
+private fun meteredDownloadConfirmationDialog(vm: UpdateViewModel) {
```
Axelen123, line 86: This should be called `MeteredDownloadConfirmationDialog` because it returns `Unit`

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
```diff
@@ -0,0 +1,224 @@
...
+                modifier = Modifier.align(Alignment.CenterHorizontally)
+            )
+        }
+    }
+}
+
+@Composable
+private fun ColumnScope.Changelog(changelog: Changelog) {
+    val scrollState = rememberScrollState()
+    Column(
```
Axelen123, line 171: Maybe use `Box` instead since there is only one "child".

app/src/main/res/values/strings.xml
```diff
@@ -300,6 +306,10 @@
 
     <string name="failed_to_check_updates">Failed to check for updates</string>
     <string name="dismiss_temporary">Not now</string>
-    <string name="update_available">New update available</string>
-    <string name="update_available_description">A new version (%s) is available for download.</string>
+    <string name="new_update_available">New update available</string>
+    <string name="new_update_available_description">A new version (%s) is available for download.</string>
```
Axelen123, line 310: Why is there a `new_` version and a "regular" version of these strings? Seems confusing

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
```diff
@@ -0,0 +1,224 @@
...
+                }
+            ) {
+                Text(stringResource(id = R.string.download))
+            }
+        },
+        title = {
+            Text(text = stringResource(id = R.string.download_update_confirmation))
+        },
+        icon = {
+            Icon(Icons.Outlined.Update, stringResource(id = R.string.update))
```
Axelen123, line 112: The description is unnecessary since the title **should** convey enough information

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
```diff
@@ -0,0 +1,224 @@
...
+        },
+        text = {
+            Text(stringResource(id = R.string.download_confirmation_metered))
+        }
+    )
+}
+
+@Composable
+private fun Header(vm: UpdateViewModel) {
```
Axelen123, line 121: Try passing callbacks and data instead of the viewmodel

Axelen123 (changes requested):

app/src/main/res/values/strings.xml
Ushie: I merged changes from dev and strings started conflicting, supposedly from the new update pop-up on launch

I can rename it to update_available_dialog?

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
Ushie: A lot of dialogs can have their purpose conveyed by the title, a description is an addition to go a little more in-depth

Axelen123: Yes, but decorative icons don't need a text replacement

app/src/main/res/values/strings.xml
Axelen123: I don't know. Try to reuse strings if possible

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
Ushie: I misunderstood your initial message, I thought it was about the text field of the AlertDialog

I can't use an ImageVector without also providing a contentDescription, I thought it would be better to set a proper one than setting an empty string

app/src/main/java/app/revanced/manager/ui/screen/settings/update/UpdateScreen.kt
Ushie: It seems like a Column is necessary for this, using Box does this:

image

Axelen123: Alright. Column is fine then

app/src/main/res/values/strings.xml
Ushie: I've updated the names of these strings to be consistent with other strings in the app, but wasn't able to reuse them

app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
```diff
@@ -0,0 +1,179 @@
...
+                UninstallService.APP_UNINSTALL_ACTION -> {
+                }
+            }
+        }
+    }
+
+    init {
+        ContextCompat.registerReceiver(app, installBroadcastReceiver, IntentFilter().apply {
+            addAction(InstallService.APP_INSTALL_ACTION)
+            addAction(UninstallService.APP_UNINSTALL_ACTION)
```
Axelen123, line 152: I don't think checking for uninstalls is necessary in this screen

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
```diff
@@ -0,0 +1,175 @@
...
+    fun installUpdate() = viewModelScope.launch {
+        state = State.INSTALLING
+
+        pm.installApp(listOf(location))
+    }
+
+    private val installBroadcastReceiver = object : BroadcastReceiver() {
+        override fun onReceive(context: Context?, intent: Intent?) {
+            when (intent?.action) {
+                InstallService.APP_INSTALL_ACTION -> {
```
Axelen123, line 127: We dont need to use `when`. All other actions are filtered by the intent filter

app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
```diff
@@ -0,0 +1,175 @@
...
+                            totalSize = contentLength
+                        }
+                    }
+                    state = State.CAN_INSTALL
+                }
+            }
+        } catch (e: Exception) {
+            Log.e(tag, "Failed to download update", e)
+            app.toast(app.getString(R.string.failed_to_download_update, e.simpleMessage()))
+        }
```
Axelen123, line 115: Can't `uiSafe` be used here?

Axelen123 (changes requested):

Axelen123 (approved):

PR #1465, opened 2023-11-04
feat(Contributors Screen): implement design from Figma

Ushie:
- [ ] ~~Proper dots indicator~~
- [x] Figure out the spacing issue between users
- [x] Figure out card colors


app/src/main/java/app/revanced/manager/ui/screen/settings/ContributorScreen.kt
```diff
@@ -50,87 +54,139 @@ fun ContributorScreen(
...
+                    ) {
+                        repositories.forEach {
+                            ContributorsCard(
+                                title = it.name,
+                                contributors = it.contributors
+                            )
+                        }
+                    }
+                }
+            } ?: LoadingIndicator()
```
Axelen123, line 79: A `LazyColumn` would probably be better here considering the time complexity of the UI code

app/src/main/java/app/revanced/manager/ui/screen/settings/ContributorScreen.kt
```diff
@@ -50,87 +54,139 @@ fun ContributorScreen(
...
-    contributors: List<ReVancedContributor>
+    contributors: List<ReVancedContributor>,
+    itemsPerPage: Int = 12,
+    numberOfRows: Int = 2
 ) {
-    var expanded by remember { mutableStateOf(false) }
+    val itemsPerRow = (itemsPerPage / numberOfRows)
+
+    // Create a list of contributors grouped by itemsPerPage
+    val contributorsByPage = contributors.chunked(itemsPerPage)
```
Axelen123, line 95:
```suggestion
    val contributorsByPage = remember(itemsPerPage, contributors) {
    	contributors.chunked(itemsPerPage)
    }
```

app/src/main/java/app/revanced/manager/ui/screen/settings/ContributorScreen.kt
```diff
@@ -50,87 +54,139 @@ fun ContributorScreen(
...
+            modifier = Modifier.padding(16.dp),
+            verticalArrangement = Arrangement.spacedBy(16.dp)
+        ) {
+            Row(
+                horizontalArrangement = Arrangement.spacedBy(8.dp),
+                verticalAlignment = Alignment.CenterVertically
+            ) {
+                Text(
+                    text = processHeadlineText(title),
+                    style = MaterialTheme.typography.titleMedium.copy(fontWeight = FontWeight(500))
```
Axelen123, line 118: Aren't there font weights in `MaterialTheme`?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/settings/ContributorScreen.kt
Ushie: How can I keep the padding of that Column in the middle? applying it to the parent LazyColumn does this
image

Axelen123: Turning the `Column` that contains the `ContributorsCard`s into a `LazyColumn` should work then. The parent's `verticalScroll` would have to be removed though.

app/src/main/java/app/revanced/manager/ui/viewmodel/ContributorViewModel.kt
```diff
@@ -11,13 +14,13 @@ import kotlinx.coroutines.launch
...
+    var repositories: List<ReVancedGitRepository>? by mutableStateOf(null)
 
     init {
         viewModelScope.launch {
-            withContext(Dispatchers.IO) { reVancedAPI.getContributors().getOrNull() }?.let(
-                repositories::addAll
-            )
+            withContext(Dispatchers.IO) {
+                repositories = reVancedAPI.getContributors().getOrNull()
+            }
```
CnC-Robert, line 23: This should be done on the main thread.
```suggestion
            repositories = withContext(Dispatchers.IO) {
                reVancedAPI.getContributors().getOrNull()
            }
```

CnC-Robert (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/settings/ContributorScreen.kt
Ushie: That didn't work, it's because the LazyColumn is the part that's scrolling, instead of the whole page

Axelen123: I'll look into this later today to see if there is a way to use `LazyColumn`.

Axelen123: It seems I will be kinda busy this weekend. Can you do a basic performance test on a release build when you have time?

We can keep it as a regular column if the performance is fine.

Ushie: I have no experience with profiling apps, but following https://developer.android.com/studio/profile I have these results:

Column:
- CPU: 0% to 29%
- Memory: 49MB to 61MB

image


LazyColumn:
- CPU: 0% to 18%
- Memory: 49MB to 52MB

image


Although the results I've gotten were not consistent in each run, it consistently had a reasonable difference between LazyColumn and Column

Ushie: Turns out, LazyColumn has a "contentPadding" parameter that acts the same as padding in web does, so this is solved

Axelen123 (changes requested):

gradle/libs.versions.toml
```diff
@@ -43,7 +43,7 @@ compose-bom = { group = "androidx.compose", name = "compose-bom", version.ref =
 compose-ui = { group = "androidx.compose.ui", name = "ui" }
 compose-ui-preview = { group = "androidx.compose.ui", name = "ui-tooling-preview" }
 compose-livedata = { group = "androidx.compose.runtime", name = "runtime-livedata" }
-compose-material3 = { group = "androidx.compose.material3", name = "material3" }
+compose-material3 = { group = "androidx.compose.material3", name = "material3", version = "1.2.0-alpha10"}
```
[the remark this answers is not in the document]
Ushie: Doesn't seem like it, newest BOM uses 1.1.2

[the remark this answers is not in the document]
Axelen123: Alright

Axelen123 (approved):

PR #1471, opened 2023-11-06
feat: Disable selection of unsuggested app version by default

TheAabedKhan:
<image height="640" width="572" src="https://github.com/ReVanced/revanced-manager/assets/39409020/e3cc52a4-5a4d-4775-be04-783570cfd106"/>

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -241,7 +241,7 @@
         "versionCompatibilityCheckHint": "Restricts patches to supported app versions",
         "requireSuggestedAppVersionLabel": "Require suggested app version",
         "requireSuggestedAppVersionHint": "Enforce selection of suggested app version",
-        "requireSuggestedAppVersionDialogText": "Selecting an app that does not have the suggested version may cause unexpected issues.\n\nDisable anyways?",
+        "requireSuggestedAppVersionDialogText": "Selecting an app that is not the suggested version may cause unexpected issues.\n\nDo you want to proceed anyways?",
 
         "aboutLabel": "About",
         "snackbarMessage": "Copied to clipboard",
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -111,7 +111,7 @@
 
         "downloadToast": "Download function is not available yet",
 
-        "requireSuggestedAppVersionDialogText": "The version of the app you have selected does not match the suggested version. Please select the app that matches the recommended version.\n\nSelected version: v{selected}\nSuggested version: v{suggested}\n\n.You can disable \"Require suggested app version\" in the settings to bypass this.",
+        "requireSuggestedAppVersionDialogText": "The version of the app you have selected does not match the suggested version. Please select the app that matches the suggested version.\n\nSelected version: v{selected}\nSuggested version: v{suggested}\n\n.To proceed anyway, disable "Require suggested app version" in the settings.",
 
         "featureNotAvailable": "Feature not implemented",
         "featureNotAvailableText": "This application is a split APK and cannot be selected. Unfortunately, this feature is only available for rooted users at the moment. However, you can still install the application by selecting its APK files from your device's storage instead"
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -111,7 +111,7 @@
 
         "downloadToast": "Download function is not available yet",
 
-        "requireSuggestedAppVersionDialogText": "It is recommended to select a suggested application version to avoid unexpected issues.\n\nSuggested version: v{suggested}\nSelected version: v{selected}\n\nYou'll need to turn off \"Require suggested app version\" in settings to disable this enforcement.",
+        "requireSuggestedAppVersionDialogText": "The version of the app you have selected does not match the suggested version. Please choose the app that matches the recommended version.\n\nSelected version: v{selected}\nSuggested version: v{suggested}\n\n.You can disable the setting \"Require suggested app version\" to bypass this.",
 
         "featureNotAvailable": "Feature not implemented",
         "featureNotAvailableText": "This application is a split APK and cannot be selected. Unfortunately, this feature is only available for rooted users at the moment. However, you can still install the application by selecting its APK files from your device's storage instead"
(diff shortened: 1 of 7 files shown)


assets/i18n/en_US.json
```diff
@@ -234,8 +236,12 @@
         "autoUpdatePatchesHint": "Automatically update patches to the latest version",
         "universalPatchesLabel": "Show universal patches",
         "universalPatchesHint": "Display all apps and universal patches (may slow down the app list)",
+
         "versionCompatibilityCheckLabel": "Version compatibility check",
         "versionCompatibilityCheckHint": "Restricts patches to supported app versions",
+        "requireSuggestedAppVersionLabel": "Require suggested app version",
+        "requireSuggestedAppVersionHint": "Enforce selection of suggested app version",
+        "requireSuggestedAppVersionDialogText": "Selecting unsuggested application version may cause unexpected issues.\n\nDisable anyways?",
```
oSumAtrIX, line 244:
```suggestion
        "requireSuggestedAppVersionDialogText": "Selecting an app that does not have the suggested version may cause unexpected issues.\n\nDisable anyways?",
```

assets/i18n/en_US.json
```diff
@@ -111,6 +111,8 @@
 
         "downloadToast": "Download function is not available yet",
 
+        "requireSuggestedAppVersionDialogText": "The version of the app you have selected does not match the suggested version. Please choose the app that matches the recommended version.\n\nSelected version: v{selected}\nSuggested version: v{suggested}\n\n.You can disable the setting \"Require suggested app version\" to bypass this.",
```
oSumAtrIX, line 114:
```suggestion
        "requireSuggestedAppVersionDialogText": "The version of the app you have selected does not match the suggested version. Please select the app that matches the recommended version.\n\nSelected version: v{selected}\nSuggested version: v{suggested}\n\n.You can disable the setting \"Require suggested app version\" to bypass this.",
```

lib/ui/widgets/settingsView/settings_advanced_section.dart
```diff
@@ -21,6 +22,7 @@ class SAdvancedSection extends StatelessWidget {
         SAutoUpdatePatches(),
         SEnablePatchesSelection(),
         SVersionCompatibilityCheck(),
+        SRequireSuggestedAppVersion(),
```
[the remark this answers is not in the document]
TheAabedKhan: Good point. @oSumAtrIX, what do you say?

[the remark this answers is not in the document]
oSumAtrIX: Sounds good to me

validcube:
I have mixed opinions about this. I don't really like this proposed solution; this will make ReVanced Manager a bit more cumbersome to use. 

(cc: @PalmDevs)

The implementation could be something like this:

> [!NOTE]
> This is an example solution; the text needs to be changed to something like "This application only supports (x)" or something.

image

or

image

*Google Material 3 - [Common Buttons](https://m3.material.io/components/buttons/guidelines); [Dialogs](https://m3.material.io/components/dialogs/overview)*

oSumAtrIX:
The point of requiring the user go to the settings to toggle it on is intended design. The idea is to leverage lazyness so that the user does not attempt to proceed with an unsuggested version but still allow the user to do it if they really intend to. The same design is applied to changing the selection of patches. You need to go to the settings to enable changing the selection before you can change it to your liking. Once again, here, the default selection is suggested which is why we leverage lazyness to prevent the user from using a selection that is not recommended. 

If you allow to proceed from right within the popup the user will attempt to continue out of curiously as well as out of lazyness to get the suggested APK version. Note how now our entire leverage is used for the complete opposite. Now the user is animated to continue instead of chosing the suggested APK. And because in the most cases when the popup is shown, the patch will not not work as intended for the user, now instead of leveraging lazyness to prevent the user from causing issues, we animate them to do it which is why, just like the toggle to change the selection of patches, this toggle should be in the settings

assets/i18n/en_US.json
```diff
@@ -234,8 +236,12 @@
         "autoUpdatePatchesHint": "Automatically update patches to the latest version",
         "universalPatchesLabel": "Show universal patches",
         "universalPatchesHint": "Display all apps and universal patches (may slow down the app list)",
+
         "versionCompatibilityCheckLabel": "Version compatibility check",
         "versionCompatibilityCheckHint": "Restricts patches to supported app versions",
+        "requireSuggestedAppVersionLabel": "Require suggested app version",
+        "requireSuggestedAppVersionHint": "Enforce selection of suggested app version",
+        "requireSuggestedAppVersionDialogText": "Selecting an app that does not have the suggested version may cause unexpected issues.\n\nDisable anyways?",
```
PalmDevs, line 244: How about `Selecting an app that is not the suggested version may cause unexpected issues. Do you want to proceed anyways?`

assets/i18n/en_US.json
```diff
@@ -111,6 +111,8 @@
 
         "downloadToast": "Download function is not available yet",
 
+        "requireSuggestedAppVersionDialogText": "The version of the app you have selected does not match the suggested version. Please select the app that matches the recommended version.\n\nSelected version: v{selected}\nSuggested version: v{suggested}\n\n.You can disable \"Require suggested app version\" in the settings to bypass this.",
```
PalmDevs, line 114: I see some consistencies here. Please replace `the recommended version` with `the suggested version`.

and for the last sentence, I think `You can turn "Require suggested app version" off in settings if you want to proceed anyways.` is a better fit.

PalmDevs (changes requested):
Some string changes 👍

assets/i18n/en_US.json
oSumAtrIX:
```suggestion
        "requireSuggestedAppVersionDialogText": "The version of the app you have selected does not match the suggested version. Please select the app that matches the suggested version.\n\nSelected version: v{selected}\nSuggested version: v{suggested}\n\n.To proceed anyways, disable "Require suggested app version" in the settings.",
```

assets/i18n/en_US.json
oSumAtrIX:
```suggestion
        "requireSuggestedAppVersionDialogText": "Selecting an app that is not the suggested version may cause unexpected issues.\n\nDo you want to proceed anyways?",
```

PR #1473, opened 2023-11-08
feat: Add installer status dialog 

oSumAtrIX:
This PR shows a native dialogue after the installation finishes. It can handle all status flags. Business logic may move. 

# Todo

- [ ] Handle errors such as:
   - [ ] Installing regularly while a mount installation is currently active (In this case, uninstall the mount installation)
   - [ ] Mount errors such as wrong base APK or no base APK present (In this case, show an appropriate dialog)

# Review

- [x] Check if https://github.com/ReVanced/revanced-manager/pull/1473/files#diff-bd497764717bba99312a9c9db9f60eaa4c5d090266194189db6855574572564bR55 is placed in a suitable location
- [ ] When uninstallation is necessary, first the uninstallation prompt is shown, then the installation prompt
- [x] All strings are chosen appropriately
- [x] All icons are chosen appropriately
- [x] ~~Showing a SUCCESS status dialog is wanted~~
- [x] All TODOs in this PR are considered or done
- [x] `uiSafe` has been used correctly (In some places, it seems it should be used but it seems out of scope for this PR)

> [!WARNING]
> On installation, the install button is not greyed out, and no progress is visible, so it appears as if nothing happened.


app/src/main/res/values/strings.xml
```diff
@@ -302,4 +302,25 @@
...
+    <string name="installer_result_timeout">Timeout</string>
+    <string name="installer_result_success">Success</string>
+    <string name="installer_result_unknown">Unknown</string>
+
+    <string name="installer_result_failed_description">The installation failed for an unknown reason.</string>
+    <string name="installer_result_aborted_description">The installation was manually cancelled.</string>
+    <string name="installer_result_blocked_description">The installation was blocked.</string>
+    <string name="installer_result_conflict_description">An existing installation of the app prevents the installation. Uninstall the app and try again?</string>
+    <string name="installer_result_incompatible_description">The app is incompatible with this device.</string>
+    <string name="installer_result_invalid_description">The app is invalid.</string>
```
KobeW50, line 322: I think it would be helpful to append `Uninstall the existing app and try again?` to this description as well, as this error is most often seen when the user tries downgrading. When downgrading, the package installer will prioritize this error over the conflict error, even when the signatures don't match.

KobeW50 (commented):
Suggestion for the "invalid" error

app/src/main/res/values/strings.xml
```diff
@@ -302,4 +302,27 @@
...
+    <string name="installer_result_invalid">Invalid</string>
+    <string name="installer_result_storage_issue">Storage issue</string>
+    <string name="installer_result_timeout">Timeout</string>
+    <string name="installer_result_success">Success</string>
+    <string name="installer_result_unknown">Unknown</string>
+
+    <string name="installer_result_failed_description">The installation failed for an unknown reason.</string>
+    <string name="installer_result_aborted_description">The installation was manually cancelled.</string>
+    <string name="installer_result_blocked_description">The installation was blocked.</string>
+    <string name="installer_result_conflict_description">An existing installation of the app prevents the installation. Reinstall by uninstalling the app and try again?</string>
```
KobeW50, line 320:
```suggestion
    <string name="installer_result_conflict_description">An existing installation of the app prevents the installation. Uninstall the app and try again?</string>
```

`Reinstall by uninstalling the app and try[ing] again.` is slightly confusing, especially if the currently installed app is an unpatched app.

app/src/main/res/values/strings.xml
```diff
@@ -302,4 +302,27 @@
...
+    <string name="installer_result_incompatible_description">The app is incompatible with this device.</string>
+    <string name="installer_result_invalid_description">The app is invalid. Reinstall by uninstalling the app and try again?</string>
```
KobeW50, line 322:
```suggestion
    <string name="installer_result_invalid_description">The app is invalid. Uninstall the current installation and try again?</string>
```

KobeW50 (commented):
String changes in error dialog

app/src/main/res/values/strings.xml
oSumAtrIX: In https://github.com/ReVanced/revanced-manager/pull/1473/files#r1388757664 you suggest `Uninstall the current installation and try again?` but here you suggest `Uninstall the app and try again?`. Please chose either.

KobeW50: Here in the case of a conflict, the first sentence in the dialog gives context to what app should be uninstalled, so `Uninstall the app and try again?` is sufficient.

In the case of an invalid package, more clarification is needed due to the lack of context from the first sentence in that dialog. (This is also why I suggested the term "current installation" for that dialog instead of "existing installation", which is the term used in the conflict dialog.)

If you must use the same sentence for both, you can use `Uninstall the existing installation and try again?`

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -81,61 +86,34 @@ class InstallerViewModel(
...
-                    logger,
-                    setInputFile = { inputFile = it }
-                )
+    val progress: MutableStateFlow<ImmutableList<Step>> = MutableStateFlow(
+        PatcherProgressManager.generateSteps(
+            app,
+            input.selectedPatches.flatMap { (_, selected) -> selected },
+            input.selectedApp
+        ).toImmutableList()
+    )
```
Axelen123, line 95: UI code shouldn't be able to write to this

app/src/main/java/app/revanced/manager/ui/component/InstallerDialog.kt
```diff
@@ -0,0 +1,144 @@
...
+
+
+@Composable
+fun InstallerDialog(
+    vm: InstallerViewModel,
+) {
+    val dialogKind = DialogKind.fromValue(vm.installerStatus!!) ?: DialogKind.FAILURE
+
+    AlertDialog(
+        onDismissRequest = {},
```
Axelen123, line 37: This makes it impossible to dismiss the dialog by clicking outside of it. Is this intentional?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -147,24 +125,61 @@ class InstallerViewModel(
...
+            app, installBroadcastReceiver,
+            IntentFilter(InstallService.APP_INSTALL_ACTION),
+            ContextCompat.RECEIVER_NOT_EXPORTED
+        )
+
+        ContextCompat.registerReceiver(
+            app, uninstallBroadcastReceiver,
+            IntentFilter(UninstallService.APP_UNINSTALL_ACTION),
+            ContextCompat.RECEIVER_NOT_EXPORTED
+        )
```
Axelen123, line 171: Why do the receivers need to be separated?

app/src/main/java/app/revanced/manager/ui/component/InstallerDialog.kt
```diff
@@ -0,0 +1,144 @@
...
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import app.revanced.manager.R
+import app.revanced.manager.ui.viewmodel.InstallerViewModel
+import com.github.materiiapps.enumutil.FromValue
+
+private typealias InstallerDialogButtonHandler = ((vm: InstallerViewModel) -> Unit)
+private typealias InstallerDialogButton = @Composable (vm: InstallerViewModel) -> Unit
```
Axelen123, line 27: We are going to need the dialog outside the installer screen.
An interface could work for this.

```kt
interface InstallerDialogActions {
    fun reinstall()
    fun uninstall()
}
```

app/src/main/java/app/revanced/manager/ui/component/InstallerDialog.kt
```diff
@@ -0,0 +1,144 @@
...
+
+private fun installerDialogButton(
+    @StringRes buttonStringResId: Int,
+    handler: InstallerDialogButtonHandler = { },
+): InstallerDialogButton = { vm ->
+    TextButton(onClick = {
+        vm.showInstallerDialog = false
+        handler(vm)
+    }) { Text(stringResource(buttonStringResId)) }
+}
```
Axelen123, line 74: How about puttiing the `handler` in the enum instead of a composable?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/InstallerDialog.kt
oSumAtrIX: It is not intentional, but considering that, I'd keep it this way, because the dialogs have buttons to dismiss them and if you accidentally tap outside of it, it'll be annoying.

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
oSumAtrIX: I am not sure of any other implementation.

app/src/main/java/app/revanced/manager/ui/component/InstallerDialog.kt
oSumAtrIX: If we put the handler in the enum, then we would need an additional field for `buttonStringResId`, whereas if we keep `InstallerDialogButton` in the enum, you can pass `buttonStringResId` and `handler`.

app/src/main/res/values/strings.xml
oSumAtrIX: @Axelen123 A problem with this one is that an app we try to install can be invalid, but no installation of the app already exists. Do you have any suggestions?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
Axelen123: It is possible to handle both with a single receiver. That is what was done before. What I am asking is: Is there a reason to separate them?

app/src/main/java/app/revanced/manager/ui/component/InstallerDialog.kt
Axelen123: The dialogs created by this `AlertDialog` overload are usually quite big if I recall correctly, so I don't think people will accidentally click outside of them. All other dialogs in the app are dismissable like this, so not being able to dismiss it like this would be inconsistent. The only exception to this is the first launch auto update settings dialog, because it is absolutely required to read and acknowledge it.

app/src/main/res/values/strings.xml
Axelen123: Is there anything stopping you from checking if the app is installed and showing a different message if it is?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
oSumAtrIX: I am unaware of an implementation that allows handling both with a single receiver. If possible, please link to one as a reference.

Axelen123: [This](https://github.com/ReVanced/revanced-manager/blob/ac561e7acaf190605d02ff22ba49df6cd655a8e9/app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt#L130-L166) is what the code looked like previously. It handled both uninstall and install events with the same receiver using `when`. One of your commits seems to have separated them into two different broadcast receivers. Is there a reason for it?

oSumAtrIX: I will use a single receiver in this case. There is no need for two.

app/src/main/res/values/strings.xml
oSumAtrIX: I am not sure how I would restructure the code. Any suggestions? Alternative I could say:

```diff
+ Uninstall any existing installation
- Uninstall the current installation
```

KobeW50:
Is this also for when you update the Manager?

app/src/main/res/values/strings.xml
Axelen123: Maybe a function that returns a string resource could work: https://kotlinlang.org/docs/enum-classes.html#anonymous-classes

We could call it with a class containing the information we need, but it might be better to just change the wording instead.

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -215,6 +220,14 @@ class InstallerViewModel(
         }
     }
 
+    fun reinstall() = viewModelScope.launch {
+        pm.getPackageInfo(outputFile)?.packageName?.let { pm.uninstallPackage(it) }
+            ?: throw Exception("Failed to load application info")
+
+        pm.uninstallPackage(packageName)
```
CnC-Robert, line 227: `pm.uninstallPackage` is run twice here
Also is the exception intended here? It's not handled so the whole app will crash if the app isnt installed.

CnC-Robert (changes requested):

Ushie:
ReVanced Manager Updates will also use this, but this PR focuses on patched apps

oSumAtrIX:
@ Axelen123 Currently, the extra status message is not used, but I don't know which extra status messages exist so that I can implement them. For example, if an existing installation is present, but the patched app mismatches signatures, then the dialogue will say:

image

This flag is returned from installation, and there is no way to differentiate between an APK being invalid, for example, or a signature mismatch.

Furthermore, when uninstalling, the dialogue may also be shown. But once again, there is no way to differentiate between installations, uninstallations or the status message, so the dialogue may say that the installation has failed while the uninstallation failed.

Axelen123:
> @ Axelen123 Currently, the extra status message is not used, but I don't know which extra status messages exist so that I can implement them. For example, if an existing installation is present, but the patched app mismatches signatures, then the dialogue will say:
> 
> image
> 
> This flag is returned from installation, and there is no way to differentiate between an APK being invalid, for example, or a signature mismatch.

Not sure why Android says the app is invalid. Android should give you [CONFLICT](https://developer.android.com/reference/android/content/pm/PackageInstaller#STATUS_FAILURE_CONFLICT) on signature mismatch. What is the content of the `EXTRA_STATUS_MESSAGE` here?

> 
> Furthermore, when uninstalling, the dialogue may also be shown. But once again, there is no way to differentiate between installations, uninstallations or the status message, so the dialogue may say that the installation has failed while the uninstallation failed.

Isn't it possible to make a separate dialog or enum for uninstalls?

oSumAtrIX:
> Isn't it possible to make a separate dialog or enum for uninstalls?

Yes, but ideally we would have the existing Flags enum have variants of the install and uninstall strings.

Ushie:
Will root install type be handled in this?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -73,69 +79,49 @@ class InstallerViewModel(
     private var inputFile: File? = null
 
     private var installedApp: InstalledApp? = null
-    var isInstalling by mutableStateOf(false)
+    internal var isInstalling by mutableStateOf(false)
```
Ushie, line 82:
```suggestion
    var isInstalling by mutableStateOf(false)
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -73,69 +79,49 @@ class InstallerViewModel(
...
         private set
-    var installedPackageName by mutableStateOf<String?>(null)
+    internal var installedPackageName by mutableStateOf<String?>(null)
```
Ushie, line 84:
```suggestion
    var installedPackageName by mutableStateOf<String?>(null)
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -73,69 +79,49 @@ class InstallerViewModel(
...
         private set
-    val appButtonText by derivedStateOf { if (installedPackageName == null) R.string.install_app else R.string.open_app }
+    internal val appButtonText by derivedStateOf { if (installedPackageName == null) R.string.install_app else R.string.open_app }
```
Ushie, line 86:
```suggestion
    val appButtonText by derivedStateOf { if (installedPackageName == null) R.string.install_app else R.string.open_app }
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -73,69 +79,49 @@ class InstallerViewModel(
...
-                )
+    private val _progress: MutableStateFlow<ImmutableList<Step>> = MutableStateFlow(
+        PatcherProgressManager.generateSteps(
+            app,
+            input.selectedPatches.flatMap { (_, selected) -> selected },
+            input.selectedApp
+        ).toImmutableList()
+    )
+
+    internal val progress = _progress.asStateFlow()
```
Ushie, line 98:
```suggestion
    val progress = _progress.asStateFlow()
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import app.revanced.manager.R
+import app.revanced.manager.ui.viewmodel.PackageInstallerResult
+import com.github.materiiapps.enumutil.FromValue
+
+internal abstract class InstallerStatusDialogModel {
```
Ushie, line 26:
```suggestion
abstract class InstallerStatusDialogModel {
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+    abstract fun reinstall()
+    abstract fun install()
+}
+
+private typealias InstallerStatusDialogButtonHandler = ((model: InstallerStatusDialogModel) -> Unit)
+private typealias InstallerStatusDialogButton = @Composable (model: InstallerStatusDialogModel) -> Unit
+
+
+@Composable
+internal fun InstallerStatusDialog(
```
Ushie, line 39:
```suggestion
fun InstallerStatusDialog(
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+
+// TODO: Change to internal once https://github.com/MateriiApps/enumutil-kt/issues/4 is fixed
+@FromValue("flag")
+enum class DialogKind(
+    internal val flag: Int,
+    internal val title: Int,
+    @StringRes internal val contentStringResId: Int,
+    internal val icon: ImageVector = Icons.Outlined.ErrorOutline,
+    internal val confirmButton: InstallerStatusDialogButton = installerDialogButton(R.string.ok),
+    internal val dismissButton: InstallerStatusDialogButton? = null,
```
Ushie, line 94:
```suggestion
    val flag: Int,
    val title: Int,
    @StringRes internal val contentStringResId: Int,
    val icon: ImageVector = Icons.Outlined.ErrorOutline,
    val confirmButton: InstallerStatusDialogButton = installerDialogButton(R.string.ok),
    internal val dismissButton: InstallerStatusDialogButton? = null,
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -56,14 +56,20 @@ import java.util.logging.LogRecord
...
     private val installedAppRepository: InstalledAppRepository by inject()
     private val rootInstaller: RootInstaller by inject()
 
-    val packageName: String = input.selectedApp.packageName
+    internal var packageInstallerResult by mutableStateOf<PackageInstallerResult?>(null)
+        private set
+
+    internal var showInstallerStatusDialog by mutableStateOf(false)
+
+    internal val packageName: String = input.selectedApp.packageName
```
Ushie, line 72:
```suggestion
    var packageInstallerResult by mutableStateOf<PackageInstallerResult?>(null)
        private set

    var showInstallerStatusDialog by mutableStateOf(false)

    val packageName: String = input.selectedApp.packageName
```
Is internal necessary?

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -276,6 +308,12 @@ class InstallerViewModel(
     }
 }
 
+internal data class PackageInstallerResult(
```
Ushie, line 311:
```suggestion
data class PackageInstallerResult(
```
Is internal necessary?

Ushie (changes requested):

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
Axelen123, line 26: Shouldn't this be an interface? There are no non-abstract fields or methods

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+
+private typealias InstallerStatusDialogButtonHandler = ((model: InstallerStatusDialogModel) -> Unit)
+private typealias InstallerStatusDialogButton = @Composable (model: InstallerStatusDialogModel) -> Unit
+
+
+@Composable
+internal fun InstallerStatusDialog(
+    model: InstallerStatusDialogModel,
+) {
+    val dialogKind = DialogKind.fromValue(model.packageInstallerResult.status) ?: DialogKind.FAILURE
```
Axelen123, line 42: Is `fromValue` expensive? `remember` should be used if it is.

app/src/main/java/app/revanced/manager/ui/screen/InstallerScreen.kt
```diff
@@ -69,6 +71,24 @@ fun InstallerScreen(
...
+                }
+
+            override fun reinstall() {
+                vm.reinstall()
+            }
+
+            override fun install() {
+                vm.install(InstallType.DEFAULT) // Always default for now.
+            }
+        })
```
Axelen123, line 90: This isn't great because the initialization code will run every time the screen recomposes.
Use the view model (preferred) or `remember` to avoid it.

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -147,24 +133,57 @@ class InstallerViewModel(
...
         }
     }
 
     init {
-        ContextCompat.registerReceiver(app, installBroadcastReceiver, IntentFilter().apply {
-            addAction(InstallService.APP_INSTALL_ACTION)
-            addAction(UninstallService.APP_UNINSTALL_ACTION)
-        }, ContextCompat.RECEIVER_NOT_EXPORTED)
+        // TODO: navigate away when system-initiated process death is detected
+        //  because it is not possible to recover from it.
```
Axelen123, line 163: Move this to the `launchExpedited` call site

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
oSumAtrIX: I am not sure what `launchExpedited` is. This LoC is not part of the PR.

app/src/main/java/app/revanced/manager/ui/screen/InstallerScreen.kt
oSumAtrIX: > Use the view model

What do you mean by that? Should I move line 75-90 to a method in the view model?

Axelen123: Yes. Move the anonymous `InstallerStatusDialogModel` object to the viewmodel or make the viewmodel implement `InstallerStatusDialogModel`.

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
Axelen123: The comment was above an init block before this PR, but code was moved around so the comment is in a weird place.

It would be best to have it above the call to `workerRepository.launchExpedited`, which is the function that launches the patcher worker. There might be another place for the comment, but the current location is weird and should be changed.

oSumAtrIX: Marking as resolved as this exceeds my field of expertise and is off-topic for this PR

Axelen123: It is on-topic because this PR moves and removes a lot of code in the file, but the comment wasn't moved to account for it.

Just move [here](https://github.com/ReVanced/revanced-manager/blob/63f2e2f167d0800db4408c470a35cddbf1e7807e/app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt#L101)

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.unit.dp
+import app.revanced.manager.R
+import app.revanced.manager.ui.viewmodel.PackageInstallerResult
+import com.github.materiiapps.enumutil.FromValue
+
+interface InstallerStatusDialogModel {
+    var packageInstallerResult: PackageInstallerResult
```
Axelen123, line 28: Will compilation fail if you change this to a `val`? The UI does not need to change this value

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+}
+
+private fun installerDialogButton(
+    @StringRes buttonStringResId: Int,
+    handler: InstallerStatusDialogButtonHandler = { },
+): InstallerStatusDialogButton = { model ->
+    TextButton(onClick = {
+        model.showInstallerStatusDialog = false
+        handler(model)
+    }) { Text(stringResource(buttonStringResId)) }
```
Axelen123, line 84: Please adjust the formatting to be more inline with the rest of the codebase:
```kt
    TextButton(
    	onClick = {
        	model.showInstallerStatusDialog = false
        	handler(model)
    	}
) {
	Text(stringResource(buttonStringResId))
}
```

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+}
+
+@FromValue("flag")
+enum class DialogKind(
+    val flag: Int,
+    val title: Int,
+    @StringRes internal val contentStringResId: Int,
+    val icon: ImageVector = Icons.Outlined.ErrorOutline,
+    val confirmButton: InstallerStatusDialogButton = installerDialogButton(R.string.ok),
+    internal val dismissButton: InstallerStatusDialogButton? = null,
```
Axelen123, line 94: `internal` isn't useful here. Wasn't there already a review comment for this?

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,158 @@
...
+        title = R.string.installation_incompatible,
+        contentStringResId = R.string.installation_incompatible_description,
+    ),
+    FAILURE_INVALID(
+        flag = PackageInstaller.STATUS_FAILURE_INVALID,
+        title = R.string.installation_invalid,
+        contentStringResId = R.string.installation_invalid_description,
+        confirmButton = installerDialogButton(R.string.reinstall) { vm ->
+            vm.reinstall()
+        },
```
Axelen123, line 131: Maybe use a different name since you are no longer interacting with the viewmodel directly

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -56,13 +57,36 @@ import java.util.logging.LogRecord
...
     private val fs: Filesystem by inject()
     private val pm: PM by inject()
     private val workerRepository: WorkerRepository by inject()
     private val installedAppRepository: InstalledAppRepository by inject()
     private val rootInstaller: RootInstaller by inject()
 
+    var packageInstallerResult by mutableStateOf<PackageInstallerResult?>(null)
+        private set
+
+    var showInstallerStatusDialog by mutableStateOf(false)
```
Axelen123, line 71: I think we could just not show the dialog if `packageInstallerResult` is non-null instead of using an extra boolean.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
oSumAtrIX: I am not sure what to do here. What do you mean by "Just move here"

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
oSumAtrIX: So whenever a result is present, a dialog is shown, which means we would need to delete the result to hide the dialog.

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
Axelen123: Take the comment marked by the review and move it to the specified location

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
Axelen123: Do you need to hide and then unhide a dialog? If not, it would work.

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
oSumAtrIX: showInstallerStatusDialog is set to false in order to close the dialog

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
oSumAtrIX: What are you referring to? Should I rename vm to "model"?

Axelen123: Yes

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,163 @@
...
+            )
+        },
+        text = {
+            Column(
+                horizontalAlignment = Alignment.CenterHorizontally,
+                verticalArrangement = Arrangement.spacedBy(16.dp),
+                content = {
+                    Text(stringResource(dialogKind.contentStringResId))
+                }
+            )
```
Axelen123, line 73:
```suggestion
            Column(
                horizontalAlignment = Alignment.CenterHorizontally,
                verticalArrangement = Arrangement.spacedBy(16.dp),
            ) {
            	Text(stringResource(dialogKind.contentStringResId))
            }
```

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,163 @@
...
+    ) {
+        Text(stringResource(buttonStringResId))
+    }
+}
+
+@FromValue("flag")
+enum class DialogKind(
+    val flag: Int,
+    val title: Int,
+    @StringRes internal val contentStringResId: Int,
```
Axelen123, line 96: This is still internal

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -56,13 +57,26 @@ import java.util.logging.LogRecord
...
+    val installerStatusDialogModel = object : InstallerStatusDialogModel {
+        override var packageInstallerResult by mutableStateOf<PackageInstallerResult?>(null)
+
+        override fun reinstall() {
+            this@InstallerViewModel.reinstall()
+        }
+
+        override fun install() {
+            install(InstallType.DEFAULT) // Always default for now.
+        }
```
Axelen123, line 77: This is better as long as the formatter permits it:
```suggestion
        override fun reinstall() = this@InstallerViewModel.reinstall()

        override fun install() = install(InstallType.DEFAULT) // Always default for now.
```

app/src/main/java/app/revanced/manager/ui/viewmodel/InstallerViewModel.kt
```diff
@@ -215,6 +241,19 @@ class InstallerViewModel(
...
+        try {
+            pm.getPackageInfo(outputFile)?.packageName?.let { pm.uninstallPackage(it) }
+                ?: throw Exception("Failed to load application info")
+
+            pm.installApp(listOf(outputFile))
+        } catch (e: Exception) {
+            Log.e(tag, "Failed to reinstall", e)
+            app.toast(app.getString(R.string.reinstall_app_fail, e.simpleMessage()))
+        }
+
```
Axelen123, line 254: We have a function in `util/Util.kt`. I think it can be used here to deal with exceptions

app/src/main/res/values/strings.xml
```diff
@@ -315,4 +316,26 @@
...
+
+    <string name="installation_failed_description">The installation failed due to an unknown reason. Please try again.</string>
+    <string name="installation_aborted_description">The installation was cancelled manually.</string>
+    <string name="installation_blocked_description">The installation was blocked. Adjust your security settings and try again.</string>
+    <string name="installation_conflict_description">An existing installation of the app prevents the installation. Uninstall the app and try again?</string>
+    <string name="installation_incompatible_description">The app is incompatible with this device. Contact the developer of the app and ask for support.</string>
+    <string name="installation_invalid_description">The app is invalid. Uninstall the app and try again?</string>
+    <string name="installation_storage_issue_description">The app could not be installed due to insufficient storage. Free up some space and try again.</string>
+    <string name="installation_timeout_description">The installation took too long.</string>
+    <string name="installation_success_description">The app has been installed successfully.</string>
```
Axelen123, line 338: Didn't we say that these should be changed to something like "action" instead of "installation"?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
oSumAtrIX:
```suggestion
    @StringRes val contentStringResId: Int,
```

app/src/main/res/values/strings.xml
oSumAtrIX: The strings refer to the entire installation activity that includes uninstallation as an action, and hence, if we need to uninstall to be able to install, but the uninstallation action fails, the entire activity is terminated, and the strings are displayed accordingly.

Axelen123: Right. I forgot about that

oSumAtrIX:
@Axelen123 I have updated the PR description with up to date information.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -49,6 +52,11 @@ import java.io.File
 import java.nio.file.Files
 import java.util.UUID
 
+data class PackageInstallerResult(
+    val status: Int = PackageInstaller.STATUS_FAILURE,
+    val extraStatusMessage: String?,
+)
```
Axelen123, line 58: This seems to be unused

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -144,18 +171,37 @@ class PatcherViewModel(
...
+                }
+
+                UninstallService.APP_UNINSTALL_ACTION -> {
+                    val pmStatus = intent.getIntExtra(
+                        UninstallService.EXTRA_UNINSTALL_STATUS,
+                        PackageInstaller.STATUS_FAILURE
+                    )
+
+                    // TODO: This is unused and may be completely removed everywhere.
+                    // intent.getStringExtra(UninstallService.EXTRA_UNINSTALL_STATUS_MESSAGE)
```
Axelen123, line 186: Consider logging the extra status message. It might be useful for debugging

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -144,18 +171,37 @@ class PatcherViewModel(
...
+
+                    if (pmStatus != PackageInstaller.STATUS_SUCCESS) {
+                        installerStatusDialogModel.packageInstallerStatus = pmStatus
                     }
                 }
```
Axelen123, line 191: No logic for reattempting an installation that required uninstalling the installed app is present here. Is this intended?

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -144,18 +171,37 @@ class PatcherViewModel(
                                 input.selectedPatches
                             )
                         }
-                    } else {
-                        app.toast(app.getString(R.string.install_app_fail, extra))
+                    }
+
+                    installerStatusDialogModel.packageInstallerStatus = pmStatus
```
Axelen123, line 176: @PalmDevs should the dialog be shown for successful installs?

PalmDevs:
Please send a recording of the latest changes or link me to them, thanks 👍

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
oSumAtrIX: I am not sure if I understood correctly. If uninstallation fails, the appropriate dialog will be shown.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
oSumAtrIX: If not, after installation the only thing that changes is the button from "Install" to "Ok"

oSumAtrIX:
@PalmDevs 
https://github.com/ReVanced/revanced-manager/assets/13122796/48e874e6-84a4-48cf-9cd4-f9be7442aded

app/src/main/res/values/strings.xml
```diff
@@ -328,4 +329,25 @@
...
     <string name="remote_bundle_description">Import remote files from a URL, can automatically update</string>
     <string name="recommended">Recommended</string>
+
+    <string name="installation_failed_dialog_title">Installation failed</string>
+    <string name="installation_cancelled_dialog_title">Installation cancelled</string>
+    <string name="installation_blocked_dialog_title">Installation blocked</string>
+    <string name="installation_conflict_dialog_title">Installation conflict</string>
+    <string name="installation_incompatible_dialog_title">Installation incompatible</string>
+    <string name="installation_invalid_dialog_title">Installation invalid</string>
+    <string name="installation_storage_issue_dialog_title">Installation storage issue</string>
```
PalmDevs, line 339: Does this just mean there's not enough storage? If so, how about `Not enough storage` or something similar, so the user can clearly tell what the problem is if they just read the title and not anything else.

app/src/main/res/values/strings.xml
```diff
@@ -328,4 +329,25 @@
...
+    <string name="installation_timeout_dialog_title">Installation timeout</string>
```
PalmDevs, line 340: How about `Installation timed out`?

app/src/main/res/values/strings.xml
```diff
@@ -328,4 +329,25 @@
     <string name="local_bundle_description">Import local files from your storage, does not automatically update</string>
     <string name="remote_bundle_description">Import remote files from a URL, can automatically update</string>
     <string name="recommended">Recommended</string>
+
+    <string name="installation_failed_dialog_title">Installation failed</string>
+    <string name="installation_cancelled_dialog_title">Installation cancelled</string>
+    <string name="installation_blocked_dialog_title">Installation blocked</string>
+    <string name="installation_conflict_dialog_title">Installation conflict</string>
+    <string name="installation_incompatible_dialog_title">Installation incompatible</string>
+    <string name="installation_invalid_dialog_title">Installation invalid</string>
```
PalmDevs, line 338: When does this usually happen? Is it due to invalid APKs, is it due to a signature mismatch?

app/src/main/res/values/strings.xml
```diff
@@ -328,4 +329,25 @@
...
+    <string name="installation_storage_issue_dialog_title">Installation storage issue</string>
+    <string name="installation_timeout_dialog_title">Installation timeout</string>
+    <string name="installation_success_dialog_title">Installation succeeded</string>
+    <string name="installation_failed_description">The installation failed due to an unknown reason. Try again?</string>
+    <string name="installation_aborted_description">The installation was cancelled manually. Try again?</string>
+    <string name="installation_blocked_description">The installation was blocked. Review your device security settings and try again.</string>
+    <string name="installation_conflict_description">The installation was prevented by an existing installation of the app. Uninstall the installed app and try again?</string>
+    <string name="installation_incompatible_description">The app is incompatible with this device. Contact the developer of the app and ask for support.</string>
```
PalmDevs, line 346: Ideally, this should *probably* be checked before patching in the first place, you can place a dismissable warning dialog or something. The string looks fine though.

PalmDevs (changes requested):
A few string changes, need opinions.

app/src/main/res/values/strings.xml
oSumAtrIX: Unfortunately, there is no specification on what kind of issue could occur. Not enough storage is the most likelihood, though; other storage issues could occur, for example no file system permission for some reason, but due to the likelihood I'd keep it simple with "Not enough storage"

app/src/main/res/values/strings.xml
oSumAtrIX: This PR is primarily introducing a Dialog that handles status codes and accepts a view model for taking action for certain status codes Introducing such a warning is out of scope for the PR.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
PalmDevs: If the package installer dialog is hidden after the user clicks Install (some apps do that), yes. Otherwise, maybe not considering the package manager dialog already shows the same thing.

app/src/main/res/values/strings.xml
PalmDevs: I'll resolve this for now. If we can find out what actually causes this to appear, then we'll change it. 👍

PalmDevs (approved):
Strings and UI, LGTM.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
oSumAtrIX: The popup is hidden once the user clicks install.

Axelen123: Should this be marked as resolved?

Axelen123 (approved):
I have nothing else to suggest that isn't already being discussed.

Axelen123:
@oSumAtrIX 

> When uninstallation is necessary, first the uninstallation prompt is shown, then the installation prompt

Not sure this is within the scope of the PR. One of the cases requires checking the signature of the installed app, which would require functionality that isn't currently present in revanced library.

>  Mount errors such as wrong base APK or no base APK present (In this case, show an appropriate dialog)

Isn't this already handled?

oSumAtrIX:
> Not sure this is within the scope of the PR. One of the cases requires checking the signature of the installed app, which would require functionality that isn't currently present in revanced library.

The dialog model has two functions for installing and uninstalling apps.
The installer returns a result at which the dialog shows the corresponding dialogue. If the signature mismatches, it will suggest to uninstall the app using the models uninstall function and attempting the installation again, which, the dialog will handle the result of:

https://github.com/ReVanced/revanced-manager/assets/13122796/48e874e6-84a4-48cf-9cd4-f9be7442aded

What I meant with that message is the issue shown in the recording which shows that the installer runs before the uninstaller. The cause is not known and its unclear if it is an issue outside my env.

> Isn't this already handled?

The dialog currently ONLY handles Android installer status. It has no context about root installations and root installers also don't return any status code like the Android installer. As it is right now, the dialog is only handling Android installations.

BenjaminHalko:
OK, I _think_ this may be done. Though probably something will come up in review.

oSumAtrIX:
Some todos are unchecked in OP. Is this intended?

BenjaminHalko:
Oh, I was waiting for someone to review / confirm them before ticking them.

BenjaminHalko:
However, the success dialogue will not be implemented right?

BenjaminHalko:
Oh I just realized, while installing should the installation button be greyed out?

Ushie:
> However, the success dialogue will not be implemented right?

Success dialog won't be implemented, as that would be indicated by the "Open" FAB and  Android Package Installer's own success dialog

> Oh I just realized, while installing should the installation button be greyed out?

It should disappear, there's no concept of a disabled FAB

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -0,0 +1,166 @@
...
+            model.install()
+        },
+    ),
+    SUCCESS(
+        flag = PackageInstaller.STATUS_SUCCESS,
+        title = R.string.installation_success_dialog_title,
+        contentStringResId = R.string.installation_success_description,
+        icon = Icons.Outlined.Check,
+    );
+
```
Ushie, line 163:
```suggestion
```

Because of https://github.com/ReVanced/revanced-manager/pull/1473#issuecomment-2254259701

app/src/main/res/values/strings.xml
```diff
@@ -359,6 +360,27 @@
...
+    <string name="installation_failed_dialog_title">Installation failed</string>
+    <string name="installation_cancelled_dialog_title">Installation cancelled</string>
+    <string name="installation_blocked_dialog_title">Installation blocked</string>
+    <string name="installation_conflict_dialog_title">Installation conflict</string>
+    <string name="installation_incompatible_dialog_title">Installation incompatible</string>
+    <string name="installation_invalid_dialog_title">Installation invalid</string>
+    <string name="installation_storage_issue_dialog_title">Not enough storage</string>
+    <string name="installation_timeout_dialog_title">Installation timed out</string>
+    <string name="installation_success_dialog_title">Installation succeeded</string>
+    <string name="installation_failed_description">The installation failed due to an unknown reason. Try again?</string>
```
Ushie, line 373:
```suggestion
    <string name="installation_failed_description">The installation failed due to an unknown reason. Try again?</string>
```

Because of https://github.com/ReVanced/revanced-manager/pull/1473#issuecomment-2254259701

app/src/main/res/values/strings.xml
```diff
@@ -359,6 +360,27 @@
...
+    <string name="installation_aborted_description">The installation was cancelled manually. Try again?</string>
+    <string name="installation_blocked_description">The installation was blocked. Review your device security settings and try again.</string>
+    <string name="installation_conflict_description">The installation was prevented by an existing installation of the app. Uninstall the installed app and try again?</string>
+    <string name="installation_incompatible_description">The app is incompatible with this device. Contact the developer of the app and ask for support.</string>
+    <string name="installation_invalid_description">The app is invalid. Uninstall the app and try again?</string>
+    <string name="installation_storage_issue_description">The app could not be installed due to insufficient storage. Free up some space and try again.</string>
+    <string name="installation_timeout_description">The installation took too long. Try again?</string>
+    <string name="installation_success_description">The app has been installed successfully.</string>
+
```
Ushie, line 382:
```suggestion
```

Because of https://github.com/ReVanced/revanced-manager/pull/1473#issuecomment-2254259701

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -238,23 +283,67 @@ class PatcherViewModel(
         context.startActivity(shareIntent)
     }
 
+    private fun versionNameToInt(versionName: String): Int {
+        val versionParts = versionName.split(".")
+        return versionParts[0].toInt() * 10000 + versionParts[1].toInt() * 100 + versionParts[2].toInt()
+    }
```
Ushie, line 289: Please compare the versionCodes through Package Manager

Ushie (changes requested):

gradle/libs.versions.toml
```diff
@@ -32,6 +32,8 @@ app-icon-loader-coil = "1.5.0"
 skrapeit = "1.2.2"
 libsu = "5.2.2"
 scrollbars = "1.0.4"
+enumutil = "1.1.0"
+enumutil-ksp = "1.1.0"
```
Axelen123, line 36: Defining two versions is unnecessary unless `enumutil` and `enumutil-ksp` are actually versioned separately

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -238,23 +283,67 @@ class PatcherViewModel(
...
                 }
 
                 InstallType.ROOT -> {
                     try {
+                        // Check for base APK, first check if the app is already installed
+                        if (existingPackageInfo == null) {
+                            val inputPackageInfo = if (inputFile != null)
+                                pm.getPackageInfo(inputFile!!)
+                            else
+                                null
```
Axelen123, line 330: Use `?.let` instead. `val inputPackageInfo = inputFile?.let { pm.getPackageInfo(it) }`

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -243,18 +288,54 @@ class PatcherViewModel(
...
+            // If the app is currently installed
+            val existingPackageInfo = pm.getPackageInfo(packageName)
+            if (existingPackageInfo != null) {
+                // Check if the app version is less than the installed version
+                if (pm.versionNameToInt(input.selectedApp.version) < pm.versionNameToInt(existingPackageInfo.versionName)) {
+                    // Exit if the selected app version is less than the installed version
+                    installerStatusDialogModel.packageInstallerStatus = PackageInstaller.STATUS_FAILURE_CONFLICT
+                    return@launch
+                }
+            }
```
Ushie, line 301: That's not what I meant in my review, use versionCodes, not versionName

versionName doesn't determine if it can be installed or not, versionCode does, if its equal or bigger then it can be installed

Ushie (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
BenjaminHalko: I have now switched it to use:
```kotlin
fun getVersionCode(packageInfo: PackageInfo): Int {
    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
        packageInfo.longVersionCode.toInt()
    } else {
        packageInfo.versionCode
    }
}
```
Is this correct?

Ushie: Make sure to check if its equal too, because equal can be installed

app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
```diff
@@ -172,4 +177,6 @@ fun PatcherScreen(
             }
         }
     }
-}
\ No newline at end of file
+}
+
+// Create an uninstall dialogue
```
Axelen123, line 182: What is this?

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -286,11 +369,20 @@ class PatcherViewModel(
                     }
                 }
             }
-        } finally {
+        } catch(_: Exception) {
             isInstalling = false
```
Axelen123, line 373: The exception should be rethrown or handled in some other way. Will `isInstalling` be unset properly with root installs?

app/src/main/java/app/revanced/manager/util/PM.kt
Axelen123, line 1: What happened here? the diff looks weird

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
BenjaminHalko: I think I was going to create my own uninstall dialog but then I saw it had already been created

app/src/main/java/app/revanced/manager/util/PM.kt
BenjaminHalko: I honestly have no clue. I even tried taking the old version and re-applying just the changes I've made, yet it still remains like this.

All I added was the `getVersionCode` function.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
BenjaminHalko: OK, it now prints a message just like when the root installation fails. And if the root installation fails it also sets `isInstalling` to false

Ushie (approved):

app/src/main/java/app/revanced/manager/util/PM.kt
Axelen123: It appears the reason why the diff looks weird is because the file was switched to CRLF line endings for some reason. Android Studio can switch it back.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -165,18 +191,39 @@ class PatcherViewModel(
...
             }
         }
     }
 
     init { // TODO: navigate away when system-initiated process death is detected because it is not possible to recover from it.
-        ContextCompat.registerReceiver(app, installBroadcastReceiver, IntentFilter().apply {
-            addAction(InstallService.APP_INSTALL_ACTION)
-        }, ContextCompat.RECEIVER_NOT_EXPORTED)
+        ContextCompat.registerReceiver(
+            app, installerBroadcastReceiver,
```
Axelen123, line 220:
```suggestion
            app,
            installerBroadcastReceiver,
```

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -283,14 +367,26 @@ class PatcherViewModel(
                             rootInstaller.uninstall(packageName)
                         } catch (_: Exception) {
                         }
+                        isInstalling = false
                     }
                 }
             }
-        } finally {
```
Axelen123, line 289: There are still cases where `isInstalling` will not have the correct value, such as:
- Successful root install
- [Version conflict early return](https://github.com/ReVanced/revanced-manager/blob/970c3fcd38fbdd595d73e6dd2093820d75b32fa0/app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt#L302-L305)
- [Split APK check early return](https://github.com/ReVanced/revanced-manager/blob/970c3fcd38fbdd595d73e6dd2093820d75b32fa0/app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt#L328-L332)

The easiest way to ensure `isInstalling` is set properly is to put it in a `finally` block like before. The only case where we don't want to unset it is if we have started a regular `PM` install, so create a boolean `var` to track that and check the variable inside the `finally` block.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -283,14 +367,26 @@ class PatcherViewModel(
...
     }
 
+    fun reinstall() = viewModelScope.launch {
+        uiSafe(app, R.string.reinstall_app_fail, "Failed to reinstall") {
+            pm.getPackageInfo(outputFile)?.packageName?.let { pm.uninstallPackage(it) }
+                ?: throw Exception("Failed to load application info")
+            
+            pm.installApp(listOf(outputFile))
+        }
+    }
```
Axelen123, line 388: Shouldn't `isInstalling` be set here?

app/src/main/java/app/revanced/manager/util/PM.kt
Axelen123, line 1: The codebase uses LF line endings, but this file was converted to CRLF.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -241,20 +289,58 @@ class PatcherViewModel(
...
             when (installType) {
                 InstallType.DEFAULT -> {
+                    // Check if the app is mounted as root
+                    // If it is, unmount it first, silently
+                    if (rootInstaller.hasRootAccess() && rootInstaller.isAppMounted(packageName)) {
+                        rootInstaller.unmount(packageName)
+                    }
+
+                    // Install regularly
+                    isInstallingRegularly = true
```
Axelen123, line 319: Move this after `installApp` so it is unset if `installApp` throws an exception for some reason

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -241,20 +289,58 @@ class PatcherViewModel(
     fun open() = installedPackageName?.let(pm::launch)
 
     fun install(installType: InstallType) = viewModelScope.launch {
+        var isInstallingRegularly = false
```
Axelen123, line 292: Rename it to something like `pmInstallStarted`.

app/src/main/java/app/revanced/manager/util/PM.kt
```diff
@@ -115,6 +115,14 @@ class PM(
 
     fun PackageInfo.label() = this.applicationInfo.loadLabel(app.packageManager).toString()
 
+    fun getVersionCode(packageInfo: PackageInfo): Int {
+        return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.P) {
+            packageInfo.longVersionCode.toInt()
+        } else {
+            packageInfo.versionCode
+        }
+    }
```
Axelen123, line 124: There is an official function that can be used instead: https://developer.android.com/reference/kotlin/androidx/core/content/pm/PackageInfoCompat#getLongVersionCode(android.content.pm.PackageInfo)

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/util/PM.kt
BenjaminHalko: `getLongVersionCode` was added in API 28, we target a minimum of API 26.

app/src/main/java/app/revanced/manager/util/PM.kt
BenjaminHalko: It is already switch to LF. At least that is what Android Studio says.

Axelen123: Yep, it has been switched, although it wasn't at the time I wrote the comment.

app/src/main/java/app/revanced/manager/util/PM.kt
```diff
@@ -115,6 +116,10 @@ class PM(
 
     fun PackageInfo.label() = this.applicationInfo.loadLabel(app.packageManager).toString()
 
+    fun getVersionCode(packageInfo: PackageInfo): Long {
+        return PackageInfoCompat.getLongVersionCode(packageInfo)
+    }
```
Axelen123, line 121:
```suggestion
    fun getVersionCode(packageInfo: PackageInfo) = PackageInfoCompat.getLongVersionCode(packageInfo)
```

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -51,8 +53,12 @@ import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeout
 import org.koin.core.component.KoinComponent
 import org.koin.core.component.inject
+import java.io.ByteArrayInputStream
 import java.io.File
 import java.nio.file.Files
+import java.security.cert.CertificateFactory
+import java.security.cert.X509Certificate
+import java.security.MessageDigest
```
Axelen123, line 61: The new imports seem unused

app/src/main/res/values/strings.xml
```diff
@@ -361,6 +362,24 @@
...
+    <string name="installation_conflict_dialog_title">Installation conflict</string>
+    <string name="installation_incompatible_dialog_title">Installation incompatible</string>
+    <string name="installation_invalid_dialog_title">Installation invalid</string>
+    <string name="installation_storage_issue_dialog_title">Not enough storage</string>
+    <string name="installation_timeout_dialog_title">Installation timed out</string>
+    <string name="installation_failed_description">The installation failed due to an unknown reason. Try again?</string>
+    <string name="installation_aborted_description">The installation was cancelled manually. Try again?</string>
+    <string name="installation_blocked_description">The installation was blocked. Review your device security settings and try again.</string>
+    <string name="installation_conflict_description">The installation was prevented by an existing installation of the app. Uninstall the installed app and try again?</string>
+    <string name="installation_incompatible_description">The app is incompatible with this device. Contact the developer of the app and ask for support.</string>
```
Axelen123, line 378: Suggested by @kitadai31

```suggestion
    <string name="installation_incompatible_description">The APK is incompatible with this device. Use the correct APK for your device and try again.</string>
```

Axelen123 (approved):
~~Code looks good to me. The PR can be merged after dealing with the final comment.~~

oSumAtrIX:
Not sure, if caused by this PR but I tried patching YT via the asset produced by a workflow run on this PR, and after clicking on install, it mentioned an installation conflict and blocked installation. Saving the APK file was also disabled. ReVanced Manager Flutter did  install the patched app

Axelen123:
Looks like I missed that. This PR does introduce a check, but it does not account for patches that change the package name.

oSumAtrIX:
The issue stems from the fact that the dialog uses the input apk, but since we install the output apk, the output should be used

app/src/main/res/values/strings.xml
```diff
@@ -375,7 +375,7 @@
     <string name="installation_aborted_description">The installation was cancelled manually. Try again?</string>
     <string name="installation_blocked_description">The installation was blocked. Review your device security settings and try again.</string>
     <string name="installation_conflict_description">The installation was prevented by an existing installation of the app. Uninstall the installed app and try again?</string>
-    <string name="installation_incompatible_description">The app is incompatible with this device. Contact the developer of the app and ask for support.</string>
+    <string name="installation_incompatible_description">The APK is incompatible with this device. Use the correct APK for your device and try again.</string>
```
kitadai31, line 378: I think "The APK" at the beginning should be remain "The app".
Because it's inconsistent with other messages and it is talking about the patched **app** here.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -228,20 +270,58 @@ class PatcherViewModel(
...
     fun install(installType: InstallType) = viewModelScope.launch {
+        var pmInstallStarted = false
         try {
             isInstalling = true
+
+            val currentPackageInfo = pm.getPackageInfo(outputFile)
+                ?: throw Exception("Failed to load application info")
+
+            // If the app is currently installed
+            val existingPackageInfo = pm.getPackageInfo(packageName)
```
Axelen123, line 281: This needs to account for patches that change the package name.
```suggestion
            val existingPackageInfo = pm.getPackageInfo(currentPackageInfo.packageName)
```

Axelen123 (changes requested):

PR #1490, opened 2023-11-12
feat: Select bundle type before adding bundle

### LINKED ISSUE #1250: feat: Select bundle type, before entering the screen to add it
Before you enter the screen to import a patch bundle, let the user first choose if they want to import a remote or local bundle:

image

UX-wise it makes no sense to put the bundle type here as a button. Either a dropdown or a switch should be used. At best the user should first decide if they want a local or remote bundle, then enter this screen. This simplifies the screen and gets rid of the confusing UX of switching between bundle types changing the entire screens content.

The bundle type option also exists after adding the bundle. For the same reason, this should also be removed here. If the user wants to change the type, they intend to add a completely new bundle to begin with. There is no purpose in having this setting after adding the bundle.

Ushie:
closes #1250

https://github.com/ReVanced/revanced-manager/assets/79272171/9e588297-7283-4358-b3cd-4cea3acacd84

- [x] Remove horizontal padding in the AlertDialog


app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleTypeSelectorDialog.kt
```diff
@@ -0,0 +1,95 @@
...
+    onDismiss: () -> Unit,
+    onConfirm: (BundleType) -> Unit,
+) {
+    var bundleType: BundleType by rememberSaveable { mutableStateOf(BundleType.Remote) }
+
+    AlertDialogExtended(
+        onDismissRequest = onDismiss,
+        confirmButton = {
+            TextButton(
+                onClick = { bundleType.let(onConfirm) }
```
Axelen123, line 37: `bundleType` is not nullable and there is no long function chain, so `.let` isn't very useful here

app/src/main/java/app/revanced/manager/ui/component/AlertDialogExtended.kt
```diff
@@ -0,0 +1,133 @@
...
+                                    .align(Alignment.CenterHorizontally)
+                            ) {
+                                icon()
+                            }
+                        }
+                    }
+                    title?.let {
+                        CompositionLocalProvider(LocalContentColor provides titleContentColor) {
+                            val textStyle = MaterialTheme.typography.headlineSmall
+                            ProvideTextStyle(textStyle) {
```
Axelen123, line 72: Make a function instead of duplicating this

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleTypeSelectorDialog.kt
```diff
@@ -0,0 +1,95 @@
...
+                )
+                Column {
+                    ListItem(
+                        modifier = Modifier.clickable(
+                            role = Role.RadioButton,
+                            onClick = { bundleType = BundleType.Local }
+                        ),
+                        headlineContent = { Text(stringResource(R.string.local)) },
+                        supportingContent = { Text(stringResource(R.string.local_bundle_description)) },
+                        overlineContent = { },
```
Axelen123, line 66: Why does this line need to be here?

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleTypeSelectorDialog.kt
```diff
@@ -0,0 +1,95 @@
...
+                        leadingContent = {
+                            RadioButton(
+                                selected = bundleType == BundleType.Local,
+                                onClick = null
```
Axelen123, line 70: Will this work if you click on the radio itself instead of the parent list item?

app/src/main/java/app/revanced/manager/ui/component/AlertDialogExtended.kt
```diff
@@ -0,0 +1,153 @@
...
+                        FlowRow(
+                            modifier = Modifier.fillMaxWidth(),
+                            horizontalArrangement = Arrangement.spacedBy(
+                                12.dp,
+                                if (tertiaryButton != null) Alignment.Start else Alignment.End
+                            ),
+                            verticalArrangement = Arrangement.spacedBy(8.dp)
+                        ) {
+                            tertiaryButton?.let {
+                                it.invoke()
```
Axelen123, line 126: Explicit `invoke` is not needed since `it` is not nullable in this context

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleTypeSelectorDialog.kt
Ushie: To force the ListItem to use 3-Line state, unfortunately Jetpack Compose doesn't follow the spec exactly and only triggers the 3-Line state when all content parameters are used, instead of when the total line count is 3

This is consistent with the proposed design in Figma and the ListItem right below, it looked weird to have one centered radiobutton and the other top aligned

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleTypeSelectorDialog.kt
Ushie: Yes, the entire ListItem is clickable so clicking on the RadioButton specifically will trigger the ListItem's onClick

It would've been nice to keep the RadioButton's onClick, so that the ripple can happen on the RadioButton directly, but using the onClick parameter on RadioButton adds a huge padding that ruins the look of the popup

Axelen123: Alright. Thats a valid reason for not specifying an onClick callback

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleTypeSelectorDialog.kt
Axelen123: In that case, can you add a comment explaining the reasoning?

app/src/main/java/app/revanced/manager/ui/screen/DashboardScreen.kt
```diff
@@ -60,6 +71,8 @@ fun DashboardScreen(
     onAppClick: (InstalledApp) -> Unit
 ) {
     var showImportBundleDialog by rememberSaveable { mutableStateOf(false) }
+    var showBundleTypeSelectorDialog by rememberSaveable { mutableStateOf(false) }
+    var bundleType by rememberSaveable { mutableStateOf(BundleType.Remote) }
```
Axelen123, line 75: You could remove `showImportBundleDialog` by making `bundleType` nullable and using `?.let`. Just remember to rename it.

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleDialog.kt
```diff
@@ -35,18 +36,19 @@ import app.revanced.manager.util.JAR_MIMETYPE
 fun ImportBundleDialog(
```
Axelen123, line 36: Shouldn't this also use the bundle type dialog?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleDialog.kt
Ushie: What do you mean? integrate the BundleTypeDialog into the ImportBundleDialog directly?

app/src/main/java/app/revanced/manager/ui/screen/DashboardScreen.kt
Ushie: I don't think that's good for code readability, and the default state of the variable is BundleType.Remote so that would be the preselected bundleType is Remote 
image

CnC-Robert:
Why is the remote option at the bottom? Shouldn't it be the primary option?

Ushie:
That's how it is in the design

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleDialog.kt
Axelen123: This composable switches type when you click on the bundle type list item. I think the item should be removed or open the BundleTypeDialog instead of switching type immediately

Ushie: Ah, now I understand, the reviewed line really didn't get the point across

Ushie:
@PalmDevs what do you think?

PalmDevs:
What Robert said is okay, you can put it at the top.

PalmDevs:
Also does Compose not have list dialogs? The ripple effect looks really wonky here when you select an option. Maybe try removing the dialog's padding?

Ushie:
The screen recording is outdated, the dialog has been updated to resemble the design as much as possible

oSumAtrIX:
This change bundle type setting should not exist. Once you select the bundle type and add it, its there. You may edit metadata such as name, but not the ty pe It's like creating a file in windows, but now you decided it should be a folder. Additionally the fact that this setting has absolutely no purpose complements this

app/src/main/java/app/revanced/manager/ui/component/bundle/ImportBundleDialog.kt
Ushie: https://github.com/ReVanced/revanced-manager/pull/1490#issuecomment-1849069638 considering this, which is out of scope, I'll be marking this as resolved considering that the changes would be immediately deleted in a follow-up commit

PalmDevs:
> This change bundle type setting should not exist. Once you select the bundle type and add it, its there. You may edit metadata such as name, but not the ty pe It's like creating a file in windows, but now you decided it should be a folder. Additionally the fact that this setting has absolutely no purpose complements this

It's gonna have to be this way until we get integrations and patched merged together, since you'd need to have 2 fields available for both of the files, but then you only need to input a URL.

Showing both of them wouldn't make much sense either, the user can get confused on why the field is there and why it's disabled even after entering a bundle URL.

And showing only one of them until a file is selected can also still cause confusion, because if the user just finishes downloading a `patches.jar` file, picks it from storage, and then sees another field, it's likely they'll be frustrated because this information was never presented to them.

If you have any other design suggestions, please send them here. I can also make it a "remote" bundle by default after clicking the + FAB, but the dialog would still need to be there so the user can start picking local files after clicking on the `Bundle type` item in the `Import patch bundle` screen and selecting the local choice.

Ushie:
He meant about editing an existing bundle, not adding a new one, just commented outside of review mistakenly

PalmDevs:
Then what he said is okay to me 👍

oSumAtrIX:
Currently, the popup leaves the choice to the user to choose what type of bundle they want. This is problematic as one will not know what type to choose. A better approach is to let the system choose the correct type automatically. Szenario:

1. The user downloads a patches.jar file
2. The user opens ReVanced Manager, clicks +, and is presented with a popup with a textbox and the option to choose from storage.
3. Because the user initially downloaded a file, they will implicitly know to select it.
4. The system reads it and adds it as a "Local" bundle

Szenario two:

1. The user is sent a URL
2. The user opens ReVanced Manager, clicks +, and is presented with a popup with a textbox and the option to choose from storage.
3. Because the user initially was sent a URL, they implicitly know to paste it into the textbox
4. The system reads it and adds it as a "Remote" or "Local" bundle (depending on the protocol, it can be a file:// or http:// URL, for example, so you can think about supporting a range of protocols or just HTTP if its easier at first)

This eliminates the user having to determine and choose the correct type and avoids the potential mistake of choosing the incorrect type alone because the system does it for them. 

A good UI could present "Chose file" and "Paste URL". If the user clicks on "Paste URL" it will automatically try to parse the clipboard or offer an input text box.

app/src/main/java/app/revanced/manager/ui/component/AlertDialogExtended.kt
```diff
@@ -0,0 +1,153 @@
...
+        Surface(
+            modifier = modifier,
+            shape = shape,
+            color = containerColor,
+            tonalElevation = tonalElevation,
+        ) {
+            Column(modifier = Modifier.padding(vertical = 24.dp)) {
+                Column(
+                    modifier = Modifier.padding(horizontal = 24.dp)
+                ) {
```
Axelen123, line 59: Can't this be done in the parent?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/AlertDialogExtended.kt
Ushie: No, this is necessary for `textHorizontalPadding`

app/src/main/java/app/revanced/manager/ui/component/AlertDialogExtended.kt
```diff
@@ -0,0 +1,153 @@
...
+) {
+    val contentColor = color ?: LocalContentColor.current
+    val finalTextStyle = textStyle ?: LocalTextStyle.current
+
+    CompositionLocalProvider(LocalContentColor provides contentColor) {
+        ProvideTextStyle(finalTextStyle) {
+            content()
+        }
+    }
+}
```
Axelen123, line 153: I think I have seen compose-m3 code do something like this:

```suggestion
@Composable
private fun ContentStyle(
    color: Color = LocalContentColor.current,
    textStyle: TextStyle = LocalTextStyle.current,
    content: @Composable () -> Unit
) {
    CompositionLocalProvider(LocalContentColor provides color) {
        ProvideTextStyle(textStyle) {
            content()
        }
    }
}
```

Make sure the code still compiles though.

Axelen123 (changes requested):
Should be good after this final change.

Axelen123 (approved):

PR #1496, opened 2023-11-16
fix(app-selector): show patchable installed apps first

TheAabedKhan:
|Before|After|
|-----|-----|
|image|image|

### DIFF
--- a/app/src/main/java/app/revanced/manager/util/PM.kt
+++ b/app/src/main/java/app/revanced/manager/util/PM.kt
@@ -80,9 +80,9 @@ class PM(
             (compatibleApps.await() + installedApps.await())
                 .distinctBy { it.packageName }
                 .sortedWith(
-                    compareBy<AppInfo>{
-                        it.packageInfo == null || it.patches == 0
-                    }.thenByDescending{
+                    compareByDescending<AppInfo>{
+                        it.packageInfo != null && (it.patches ?: 0) > 0
+                    }.thenByDescending {
                         it.patches
                     }.thenBy {
                         it.packageInfo?.label()
--- a/app/src/main/java/app/revanced/manager/util/PM.kt
+++ b/app/src/main/java/app/revanced/manager/util/PM.kt
@@ -80,7 +80,9 @@ class PM(
             (compatibleApps.await() + installedApps.await())
                 .distinctBy { it.packageName }
                 .sortedWith(
-                    compareByDescending<AppInfo> {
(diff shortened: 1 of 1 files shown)


app/src/main/java/app/revanced/manager/ui/screen/AppSelectorScreen.kt
```diff
@@ -60,8 +60,11 @@ fun AppSelectorScreen(
     var search by rememberSaveable { mutableStateOf(false) }
 
     val appList by vm.appList.collectAsStateWithLifecycle(initialValue = emptyList())
-    val filteredAppList = remember(appList, filterText) {
-        appList.filter { app ->
+    val sortedAppList = remember(appList) {
+        appList.sortedWith(compareBy { !vm.isAppInstalled(it.packageInfo) || it.patches == 0 })
+    }
```
CnC-Robert, line 65: Sorting should be done in `PM`

app/src/main/java/app/revanced/manager/ui/viewmodel/AppSelectorViewModel.kt
```diff
@@ -20,6 +20,8 @@ class AppSelectorViewModel(
 
     fun loadLabel(app: PackageInfo?) = with(pm) { app?.label() ?: "Not installed" }
 
+    fun isAppInstalled(app: PackageInfo?) = with(pm) { app?.label() != null }
```
CnC-Robert, line 23: Instead of this just use `pm.getPackageInfo() != null` in the sorter

CnC-Robert (changes requested):

app/src/main/java/app/revanced/manager/util/PM.kt
```diff
@@ -80,7 +80,9 @@ class PM(
             (compatibleApps.await() + installedApps.await())
                 .distinctBy { it.packageName }
                 .sortedWith(
-                    compareByDescending<AppInfo> {
+                    compareBy<AppInfo>{
+                        it.packageInfo == null || it.patches == 0
```
CnC-Robert, line 84: It might be more readable to use `compareByDescending` instead
```suggestion
                    compareByDescending<AppInfo>{
                        it.packageInfo != null && it.patches > 0
```

CnC-Robert (changes requested):

Ushie (approved):

CnC-Robert (approved):

Issue #1525, opened 2023-11-25
feat: Trim debug log output


taku-nm:
### Feature description

I'd suggest trimming any patch that has been included by default from log.
Anything that has been altered by the user should remain included.

Conceivably one might exclude all statements following the regex `.* succeeded` but this might go against attempts of removing hard-coded variables.

## Checklist
- [ ] Trim patches included by default
- [ ] Include note for all patches excluded manually
- [ ] Include note wherever something has been trimmed

### Motivation

After a while of seeing feat #1176 deployed, it has become apparent that the Discord message limit has become the main barrier for most users seeking support. The message limit is easily reached by the fact that Apps like YouTube have a lot of patches. 
This surely does not affect all possibilities one might reach out for support, but it does hinder the main one. 

tl;dr: To remove overall clutter and making it easier to share the logs.

### Additional context

Original comment moved from #1176 ;
I'm not entirely sure if this is a feat or bug? 
This could be extended to trim even more irrelevant information from the log as well, but isn't really needed.

### Acknowledgements

- [X] This request is not a duplicate of an existing issue.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The issue is solely related to the ReVanced Manager

oSumAtrIX:
Every setting or patch option that has not been touched can be trimmed. Exceptions and everything that was touched should be included

taku-nm:
> Every setting or patch option that has not been touched can be trimmed

Couldn't this result in issues? Hypothetically there could be a patch that does require a value from the user for the patched app to function properly but does not throw an exception when fed the default value. 
To notice that scenario the supporter must be aware that the patch needs something else than the default value to function properly, and that the patch is included by default and therefore trimmed.

oSumAtrIX:
Of something is not included it is implicitly clear, that it's default

Ushie:
SGTM

taku-nm:
Relying on implications within a debug log is not very helpful. This is meant for supporters who may simply have forgotten that some patch requires something else besides default options.
Printing those with options costs nothing and would serve as a helpful reminder.

oSumAtrIX:
Nothing that is default has to be included because it serves no purpose. Simply give an example if it were.

taku-nm:
What about the reddit patches -> missing client ID?
If the convention is going to be that every patch that requires a value besides default throws ans exception, then this can be left out.

oSumAtrIX:
[...]

PR #1526, opened 2023-11-25
feat: updated logs

BenjaminHalko:
closes: #1525

### DIFF
--- a/lib/ui/views/installer/installer_viewmodel.dart
+++ b/lib/ui/views/installer/installer_viewmodel.dart
@@ -179,19 +179,49 @@ class InstallerViewModel extends BaseViewModel {
     }
   }
 
+  void _trimLogs(List<String> logLines, String keyword, String? newString) {
+    final lineCount = logLines.where((line) => line.endsWith(keyword)).length;
+    final index = logLines.indexWhere((line) => line.endsWith(keyword));
+    if (newString != null && lineCount > 0) {
+      logLines.insert(index, newString.replaceAll('{lineCount}', lineCount.toString()));
+    }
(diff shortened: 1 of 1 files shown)


lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -212,10 +245,9 @@ class InstallerViewModel extends BaseViewModel {
       'Integration source: ${_managerAPI.getIntegrationsRepo()}',
       
       '\n- Logs',
-      logs,
```
Ushie, line 215: Why not put logsTrimmed here instead of appending it in the clipboard copy?

Ushie (changes requested):

lib/ui/views/installer/installer_viewmodel.dart
BenjaminHalko: Because the I would have to write .join('\n').

BenjaminHalko: Although it might be cleaner?

BenjaminHalko: Maybe I should do that....

BenjaminHalko: There, changed it.

Ushie: Are you sure you need to join \n here? it's still being joined by the clipboard copy anyways

Ushie: Nvm

Ushie (approved):

Issue #1531, opened 2023-11-27
Exported patch logs show bugged patch selection


KobeW50:
### Bug description

I've noticed 2  issues with the new patch logs that are copied to clipboard.

A) The log lists all of the included patched in the `Patches removed:` field. (In the logs below, i added a line-break before this field so it is easier to find)

B) The log lists all of the included patches in the `Patches added:` field, even though there were only default patches included. (Tbh i'm not sure if this one is a bug)

## Steps to reproduce

1. Clear the app data of the Manager 
2. Patch the suggested YT version with the 61 default patches (without changing any of the settings in the Manager).
3. Export and inspect the patch log.

Bug A and B do not always occur. Once bug A occurs, the only way it can go away is by clearing the app data, which may temporarily fix the issue. I'm not too sure about when bug B occurs, or if it is related to bug A.

https://github.com/ReVanced/revanced-manager/assets/84587632/0f9042a1-cabc-42c1-87a9-661dfe1f249e

### Version of ReVanced Manager and version & name of application you tried to patch

1.17.0.
YT 18.45.41

### Installation type

None

### Device logs

```shell
.
```

### Patcher logs

```shell
- Device Info
ReVanced Manager: 1.17.0
Build: release
Model: SM-S906U1
Android version: 13
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.google.android.youtube v18.45.41
Patches version: v2.201.0
[...]

PR #1535, opened 2023-11-28
fix: exported logs patch selection

BenjaminHalko:
fixes: #1531


Ushie (changes requested):
```
App: com.laurencedawson.reddit_sync vv23.06.30-13:39
Patches version: v2.201.1
Patches added: Spoof client [OAuth client ID: j]
Patches removed: Disable ads, Disable Sync for Lemmy bottom sheet, Spoof client [OAuth client ID: j]
Options changed: None
```
Applying options makes it think that its both added and removed, and the options changed field stays None

BenjaminHalko:
Did you disable the other 2 patches, or is that another bug as well?

Ushie:
I did disable them

Ushie:
The testing seems to have been effected by #1540, will retest when possible

Ushie:
Hey, @validcube or @TheAabedKhan do you think it's possible to test this?

I don't have the time to build and test this

validcube (approved):
🥞 Tested & Approved

Issue #1511, opened 2023-11-22
Unable to patch youtube


Latooo:
### Bug description

i try 5 times to patch youtube with revanced manager. i have the good youtube version and i'm correctly rooted.

### Version of ReVanced Manager and version & name of application you tried to patch

Manager: 1.16.0
Youtube v18.45.41

### Installation type

Root

### Device logs

```shell
- Device Info
ReVanced Manager: 1.16.0
Build: release
Model: Xiaomi 12
Android version: 13
Supported architectures: arm64-v8a, armeabi-v7a
Root permissions: Yes

- Patch Info
App: com.google.android.youtube v18.45.41
Patches version: v2.199.0
Patches: Alternative thumbnails, Always autorepeat, Announcements, Bypass URL redirects, Client spoof, Comments, Copy video url, Custom player overlay opacity, Disable auto captions, Disable fullscreen ambient mode, Disable player popup panels, Disable precise seeking gesture, Disable resuming Shorts on startup, Disable rolling number animations, Disable suggested video end screen, Disable zoom haptics, Enable debugging, Enable slide to seek, Enable tablet layout, External downloads, HDR auto brightness, Hide 'Load more' button, Hide Shorts components, Hide ads, Hide album cards, Hide autoplay button, Hide breaking news shelf, Hide captions button, Hide cast button, Hide crowdfunding box, Hide email address, Hide endscreen cards, Hide filter bar, Hide floating microphone button, Hide info cards, Hide layout components, Hide player buttons, Hide seekbar, Hide timestamp, Hide video action buttons, Minimized playback, Navigation buttons, Open links externally, Playback speed, Player flyout menu, Premium heading [Use premium heading: true], Remember video quality, Remove tracking query parameter, Restore old seekbar thumbnails, Restore old video quality menu, Return YouTube Dislike, Seekbar tapping, SponsorBlock, Spoof app version, Spoof device dimensions, Swipe controls, Tablet mini player, Theme [Dark theme background color: @android:color/black, Light theme background color: @android:color/white], Vanced MicroG support, Video ads, Wide searchbar

- Settings
Allow changing patch selection: false
Version compatibility check: true
Show universal patches: false
Patches source: revanced/revanced-patches
Integration source: revanced/revanced-integrations

- Logs
Copying APK
Reading APK
Decoding app manifest
Loading patches
Merging integrations
Deleting existing resource cache directory
Decoding resources
Executing patches
Alternative thumbnails succeeded
[...]

PR #1543, opened 2023-12-02
docs: Improve troubleshooting solutions

raxod502:
Document advice that could help users troubleshoot issues like https://github.com/ReVanced/revanced-manager/issues/1493 and https://github.com/ReVanced/revanced-manager/issues/1511 and https://github.com/ReVanced/revanced-patches-template/issues/114 without needing to contact support.

### DIFF
--- a/docs/3_troubleshooting.md
+++ b/docs/3_troubleshooting.md
@@ -5,14 +5,16 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 - 💉 Patching fails with an error
 
   Make sure ReVanced Manager is up to date by following [🔄 Updating ReVanced Manager](2_3_updating.md) and select the **Default** button when choosing patches.
-  
+
 - 🚫 App not installed as package conflicts with an existing package
 
-  An existing installation of the app you're trying to patch is conflicting with the patched app. Uninstall the existing app before installing the patched app.
+  An existing installation of the app you're trying to patch is conflicting with the patched app (i.e., the existing version of the app doesn't allow upgrading to the patched version, or a different copy of the app causes a conflict). Uninstall the existing app before installing the patched app.
 
 - ❗️ Error code `135`, `139` or `1` when patching the app
 
-  Your device is not supported. Refer to the [Prerequisites](0_prerequisites.md) page for supported devices.
+  You may be trying to patch a split APK[^1], which is sometimes not supported by ReVanced. Sometimes apps are available in both split and non-split forms. If your existing installation of the app is a split APK, try obtaining a non-split version and see if that can be patched correctly.
+
+  You may also get this error if your device is not supported. Refer to the [Prerequisites](0_prerequisites.md) page for supported devices.
 
   Alternatively, you can use [ReVanced CLI](https://github.com/revanced/revanced-cli) to patch the app.
 
@@ -25,3 +27,5 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 The next page will teach you how to build ReVanced Manager from source.
 
 Continue: [🔨 Building from source](4_building.md)
+
+[^1]: https://old.reddit.com/r/revancedapp/comments/xna2gn/youtube_unsupported_version_and_split_apk/
--- a/docs/3_troubleshooting.md
+++ b/docs/3_troubleshooting.md
@@ -28,4 +28,4 @@ The next page will teach you how to build ReVanced Manager from source.
 
 Continue: [🔨 Building from source](4_building.md)
 
-[^1]: https://old.reddit.com/r/revancedapp/comments/xna2gn/youtube_unsupported_version_and_split_apk/
+[^1]: https://developer.android.com/guide/app-bundle/app-bundle-format
--- a/docs/3_troubleshooting.md
+++ b/docs/3_troubleshooting.md
@@ -14,7 +14,7 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 
   You may be trying to patch a split APK[^1], which is sometimes not supported by ReVanced. Sometimes apps are available in both split and non-split forms. If your existing installation of the app is a split APK, try obtaining a non-split version and see if that can be patched correctly.
 
-  You may also get this error if your device is not supported. Refer to the [Prerequisites](0_prerequisites.md) page for supported devices.
+  Your device may otherwise be unsupported. Please look at the [Prerequisites](0_prerequisites.md) page for supported devices.
 
   Alternatively, you can use [ReVanced CLI](https://github.com/revanced/revanced-cli) to patch the app.
 
--- a/docs/3_troubleshooting.md
+++ b/docs/3_troubleshooting.md
@@ -8,7 +8,7 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 
 - 🚫 App not installed as package conflicts with an existing package
 
-  An existing installation of the app you're trying to patch is conflicting with the patched app (i.e., the existing version of the app doesn't allow upgrading to the patched version, or a different copy of the app causes a conflict). Uninstall the existing app before installing the patched app.
+  An existing installation of the app you're trying to patch conflicts with the patched app (i.e., signature mismatch or downgrade). Uninstall the existing app before installing the patched app.
 
 - ❗️ Error code `135`, `139` or `1` when patching the app
 
--- a/docs/3_troubleshooting.md
+++ b/docs/3_troubleshooting.md
@@ -12,7 +12,7 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 
 - ❗️ Error code `135`, `139` or `1` when patching the app
 
-  You may be trying to patch a split APK[^1], which is sometimes not supported by ReVanced. Sometimes apps are available in both split and non-split forms. If your existing installation of the app is a split APK, try obtaining a non-split version and see if that can be patched correctly.
+  You may be trying to patch a split APK[^1]. This may not work under certain circumstances. In such a case, patch a full APK.
 
   Your device may otherwise be unsupported. Please look at the [Prerequisites](0_prerequisites.md) page for supported devices.
 


oSumAtrIX (changes requested):

docs/3_troubleshooting.md
```diff
@@ -5,13 +5,17 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
...
+
 - 🚫 App not installed as package conflicts with an existing package
 
   An existing installation of the app you're trying to patch is conflicting with the patched app. Uninstall the existing app before installing the patched app.
 
+  This may also occur when trying to install the previous version of an app. Android sometimes blocks downgrading apps. However, if you uninstall the existing app, and try to install the downgrade again, it usually works.
+
 - ❗️ Error code `135`, `139` or `1` when patching the app
 
+  You may be trying to patch a split APK. This is common with the YouTube app. You can visit APKMirror to download an alternate version of the app that is not split (the `bundle` label means it is split) and install it.
```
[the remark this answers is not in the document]
raxod502: How can the user tell whether the APK they have is a split APK? If we don't explain how to make this determination via APKMirror, then we should provide some other explanation.

> It is the user's responsibility to source APK files, a precondition also stated in the documentation of ReVanced Manager.

Sorry, I looked for the place in the documentation where the requirement to use non-split APKs was mentioned, but I didn't manage to find it. Would you mind pointing me in the right direction?

docs/3_troubleshooting.md
```diff
@@ -5,13 +5,17 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 - 💉 Patching fails with an error
 
   Make sure ReVanced Manager is up to date by following [🔄 Updating ReVanced Manager](2_3_updating.md) and select the **Default** button when choosing patches.
-  
+
 - 🚫 App not installed as package conflicts with an existing package
 
   An existing installation of the app you're trying to patch is conflicting with the patched app. Uninstall the existing app before installing the patched app.
 
+  This may also occur when trying to install the previous version of an app. Android sometimes blocks downgrading apps. However, if you uninstall the existing app, and try to install the downgrade again, it usually works.
```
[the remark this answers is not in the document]
raxod502: Reading the existing text, I assumed it was referring to multiple copies of the same app being installed - something that's common with ReVanced with the "change package identifier" option. Not that it was referring to downgrading.

In my opinion, explaining that downgrading is the same as "installing over an existing installation" makes it easier for the user to understand why they encountered the error. I could combine the sentences if you feel there is too much text.

kitadai31:
> - ❗️ Error code `135`, `139` or `1` when patching the app

This section should be separated into Exit code `135` or `139` and Exit code `1`.
Because the Exit code 1 has nothing to do with unsupported device issues.
(Error code 1 is usually caused by split apk, incorrect user input (like color codes in `Theme` patch), or a issue on the patch side.)

docs/3_troubleshooting.md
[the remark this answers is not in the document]
oSumAtrIX: > Reading the existing text, I assumed it was referring to multiple copies of the same app being installed

The existing text says that an existing app conflicts with the app you are trying to install. Why it does (e.g Downgrade or signature mismatch) is a technicality that is irrelevant to the user. The (only) solution is applicable to all cases, which is uninstalling the app before installing the patched app.

docs/3_troubleshooting.md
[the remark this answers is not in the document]
oSumAtrIX: > How can the user tell whether the APK they have is a split APK?

You can add and reference a footnote to the term "split" to an external source in a similar fashion to how Wikipedia does. The determination should not be made using APKMirror. It may be a popular source of APK binaries but it is not connected with ReVanced whatsoever, whereas the official explanation of what a split APK is, is on the Android documentation website.

> where the requirement to use non-split APKs was mentioned

You may have misunderstood what I said with:

> It is the user's responsibility to source APK files, a precondition also stated in the documentation of ReVanced Manager.

Source: https://github.com/ReVanced/revanced-manager/blob/main/docs/2_1_patching.md#user-content-fn-1-0509e8ec511256471426023611fdf9d4 (first footnote)

This is not about split APK files. Split APK files can be patched in certain circumstances.

oSumAtrIX:
@kitadai31 Instead of splitting it into a separate section, simply mentioning either causes should be enough.

docs/3_troubleshooting.md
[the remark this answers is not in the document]
raxod502: Okay, I've rewritten the text to simply use wording that is more clear than what was present before, without treating downgrades as different from other installations.

docs/3_troubleshooting.md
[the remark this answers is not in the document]
raxod502: > We should not mention any specific apps because ReVanced is agnostic, so the documentation must also apply to any app.

Okay, I've updated the documentation to not mention YouTube, although I do feel that it would be helpful for ReVanced to document the requirements for what versions of apps it is able to patch (e.g., only the non-split version of YouTube can be patched, at least according to [this comment](https://old.reddit.com/r/revancedapp/comments/xna2gn/youtube_unsupported_version_and_split_apk/ipsg1b4/)).

> You can add and reference a footnote to the term "split" to an external source in a similar fashion to how Wikipedia does

Okay, I've updated the documentation to use a footnote. Currently, it links to a Reddit post. I am happy to update this to point at some official Android documentation about how to tell whether an APK file is split or not, if any such documentation exists. I wasn't able to find any when I searched.

> The determination should not be made using APKMirror

I'd love to know how the determination should be made, if not that way. This would surely be helpful information to other users, as well.

> Split APK files can be patched in certain circumstances.

It would be very helpful to know what circumstances these are.

raxod502:
> Instead of splitting it into a separate section, simply mentioning both causes should be enough

While I think having separate sections would be more informative, I have followed your suggestion here and updated the text accordingly.

docs/3_troubleshooting.md
```diff
@@ -5,14 +5,16 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 - 💉 Patching fails with an error
 
   Make sure ReVanced Manager is up to date by following [🔄 Updating ReVanced Manager](2_3_updating.md) and select the **Default** button when choosing patches.
-  
+
 - 🚫 App not installed as package conflicts with an existing package
 
-  An existing installation of the app you're trying to patch is conflicting with the patched app. Uninstall the existing app before installing the patched app.
+  An existing installation of the app you're trying to patch is conflicting with the patched app (i.e., the existing version of the app doesn't allow upgrading to the patched version, or a different copy of the app causes a conflict). Uninstall the existing app before installing the patched app.
```
oSumAtrIX, line 11: Simplifying should reduce introducing new/niche terminology such as "copy of" or "upgrading":

```suggestion
  An existing installation of the app you're trying to patch conflicts with the patched app (i.e., signature mismatch or downgrade). Uninstall the existing app before installing the patched app.
```

docs/3_troubleshooting.md
```diff
@@ -5,14 +5,16 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
...
 
 - ❗️ Error code `135`, `139` or `1` when patching the app
 
-  Your device is not supported. Refer to the [Prerequisites](0_prerequisites.md) page for supported devices.
+  You may be trying to patch a split APK[^1], which is sometimes not supported by ReVanced. Sometimes apps are available in both split and non-split forms. If your existing installation of the app is a split APK, try obtaining a non-split version and see if that can be patched correctly.
```
oSumAtrIX, line 15: Extracting the essence of the text so that it is shorter:

```suggestion
  You may be trying to patch a split APK[^1]. This may not work under certain circumstances. In such a case, patch a full APK.
```

docs/3_troubleshooting.md
```diff
@@ -5,14 +5,16 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
...
+
+  You may also get this error if your device is not supported. Refer to the [Prerequisites](0_prerequisites.md) page for supported devices.
```
oSumAtrIX, line 17:
```suggestion
  Your device may otherwise be unsupported. Please look at the [Prerequisites](0_prerequisites.md) page for supported devices.
```

docs/3_troubleshooting.md
```diff
@@ -25,3 +27,5 @@ In case you encounter any issues while using ReVanced Manager, please refer to t
 The next page will teach you how to build ReVanced Manager from source.
 
 Continue: [🔨 Building from source](4_building.md)
+
+[^1]: https://old.reddit.com/r/revancedapp/comments/xna2gn/youtube_unsupported_version_and_split_apk/
```
oSumAtrIX, line 31: Here is an official Android documentation of a split architecture:

```suggestion
[^1]: https://developer.android.com/guide/app-bundle/app-bundle-format
```

Because the docs must be agnostic towards any app, I suggest against linking to a YouTube specific solution for a generalized problem. You may want us to introduce app-specific FAQs, though this would work against our design. Problems such as those mentioned in this docs should be solvable in general. If you there are issues that require specific attention, these issues should be solved somewhere else (e.g split files cause problems in many cases, so we add a warning popup if we detect its a split and warn about the possibility of patching not working)

oSumAtrIX (changes requested):

docs/3_troubleshooting.md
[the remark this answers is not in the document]
oSumAtrIX: > It would be very helpful to know what circumstances these are.

Primarily, patching resources limit us from properly supporting split APK file patching. Regardless, if you mount a resource-patched APK file, it may still work. In some specific known cases, either patching fails or the app crashes regardless of mounting.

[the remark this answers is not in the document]
oSumAtrIX: > I'd love to know how the determination should be made, if not that way. This would surely be helpful information to other users, as well.

The split APK architecture is simple. It works similarly to the dex architecture, where Android iterates over the entries in the zip file to search for a class that is yet to be loaded in a classloader. Android queries Similarily, the internals of an APK in multiple APK files. This means a split architecture is not identifiable in itself as they are regular APK files. What gives it away is, if the APK files miss required files such as dex or resource files. In this case, this implies wether an APK file is a split or not. Each APK file also has to include a manifest, which is encoded but contains the information regarding this as well. Sites such as APKMirror distribute proprietary formats for splits that archive the separate APK files in a single file intended to be installed with their proprietary installers.

docs/3_troubleshooting.md
raxod502: As an end user rather than an Android app developer, I'm not sure how I would use the information in the official Android documentation to determine the appropriate next step. That documentation doesn't say anything about how to tell whether the app your have is split or full, nor does it suggest that some apps might be available in both varieties, nor does it have any information about how one would go about finding other versions of an app, or identifying whether they are split or full once found.

I understand the desire to not associate with specific third party projects unless necessary, but applying the suggested revisions seems to me like it will result in a troubleshooting page that can't actually be used for troubleshooting by anyone except experts who didn't need help in the first place.

oSumAtrIX: This was primarily the reason it was not included in the troubleshooting page to begin with. If necessary, the footnote can have a short paragraph outlining the essence of a split APK file and a link to the documentation page for additional insight which would satisfy both, experts as well as those that seek a short and quick answer.

raxod502: Is it okay if we simply give the quick advice to get the non-`bundle` version from APKMirror which will be what 99% of users reading the page will need to do, and then link to the official documentation as a further reference?

If not, is there any wording that will both be acceptable to this project while also providing the information that users need in order to resolve their problem?

I'm just trying to reduce your support burden from being asked time and time again the same question about YouTube failing to patch, since as far as I can tell there is currently no way for a user to find out why this is happening from the documentation, or by looking at Google's reference that you linked.

I myself still don't know what advice you would give other than visiting APKMirror and downloading the non-`bundle` version of the app. Is there actually any alternative? If there is, could you be more concrete about what it is? And if there isn't any alternative, can we just document it? Neither Android nor ReVanced exist in a void; it seems somewhat counterproductive to me to try to prune out every possible reference to third party software.

oSumAtrIX: We should not mention any third-party software such as APKMirror. The user must source apps themselves and pipe them to ReVanced for patching. If there is an issue with terminology on the side of APKMirror regarding the term `bundle`, this should be solved with APKMirror. If it is not clear enough that ReVanced needs an APK file, this can be outlined by saying that a file with the extension APK is necessary.

raxod502: Okay. It sounds like the information I think users need is for some reason not allowed to be included in the documentation. I cannot say I understand the reasons given, maybe there are some legal issues I am not aware of.

In any case, I suspect that these strict policies around "acceptable content" in official documentation are the reason ReVanced gets so many repetitive support questions that waste maintainers' time, and has such bare-bones documentation in the first place. However, I acknowledge your decision.

Since the proposed improvements to the documentation have been more or less rejected, I don't think there is anything else I can do with this PR. Please feel free to take it or leave it.

oSumAtrIX (approved):

Issue #1516, opened 2023-11-22
feat: Add automatically the patches version when saving APK


Kuddus73:
### Feature description

When saving a patched APK, add also the patches version next to the APK version (something like **_youtube-revanced_v18.45.41, patches_v2.200.0.apk_**).
![Screenshot_20231121_201626_Files](https://github.com/ReVanced/revanced-manager/assets/127908708/dd7de163-5a3d-4888-84b4-63ef12842e5c)

### Motivation

Many patches get released for the same APK version, so automatically adding also the patch version when saving the APK would help to not overwrite by mistake a previously saved APK of the same version.
I know we can just manually write the patch version when saving, but this way would save time and also make the process more automatic and quick.

### Acknowledgements

- [X] This request is not a duplicate of an existing issue.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The issue is solely related to the ReVanced Manager

PR #1547, opened 2023-12-04
feat: append patch version to the APK name while sharing/exporting

dhruvanbhalara:
- fixes: #1516

### DIFF
--- a/lib/services/patcher_api.dart
+++ b/lib/services/patcher_api.dart
@@ -232,11 +232,14 @@ Future<bool> installPatchedFile(PatchedApplication patchedApp) async {
   return false;
 }
 
-void exportPatchedFile(String appName, String version) {
+Future<void> exportPatchedFile(String appName, String version) async{
   try {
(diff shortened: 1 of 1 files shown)


lib/services/patcher_api.dart
```diff
@@ -268,9 +271,9 @@ void sharePatchedFile(String appName, String version) {
   }
 }
 
-String _getFileName(String appName, String version) {
+String _getFileName(String appName, String version, {String patchVersion=''}) {
   final String prefix = appName.toLowerCase().replaceAll(' ', '-');
-  final String newName = '$prefix-revanced_v$version.apk';
+  final String newName = '$prefix-revanced_v$version$patchVersion.apk';
```
oSumAtrIX, line 276: Maybe `some.app.package-v1.2.3-revanced-v3.0.0.apk`:

```suggestion
  final String newName = '$prefix-v$version-revanced-v$patchVersion.apk';
```

PR #1521, opened 2023-11-24
ci(build): refresh

validcube:
- test
- test
- test
- [ ] #1540

PR #1559, opened 2023-12-09
ci: reland Dart analyzer

validcube:
Enforce code styling

### DIFF
--- a/.github/workflows/analyzer.yml
+++ b/.github/workflows/analyzer.yml
@@ -1,6 +1,11 @@
 name: Analyze Code
 
 on:
+  push:
+    branches: dev
+    paths: 
+      - "**.dart"
+      - ".github/workflows/analyze.yml"
   pull_request:
     branches: [ "main", "dev" ]
     paths: 
--- a/.github/workflows/analyzer.yml
+++ b/.github/workflows/analyzer.yml
@@ -1,0 +1,27 @@
+name: Analyze Code
+
+on:
+  pull_request:
+    branches: [ "main", "dev" ]
+    paths: 
+      - "**.dart"
+      - ".github/workflows/analyze.yml"
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Setup Flutter
+        uses: subosito/flutter-action@v2
+        with:
+          channel: 'stable'
+          cache: true
+      - name: Install Flutter dependencies
+        run: flutter pub get
+      - name: Generate files with Builder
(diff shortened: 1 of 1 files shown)


.github/workflows/analyzer.yml
```diff
@@ -0,0 +1,32 @@
+name: Analyze Code
+
+on:
+  push:
+    branches: dev
+    paths: 
+      - "**.dart"
+      - ".github/workflows/analyze.yml"
+  pull_request:
+    branches: [ "main", "dev" ]
```
Ushie, line 10: Wouldn't it be relevant for all branches?

.github/workflows/analyzer.yml
```diff
@@ -0,0 +1,32 @@
+name: Analyze Code
+
+on:
+  push:
+    branches: dev
+    paths: 
+      - "**.dart"
+      - ".github/workflows/analyze.yml"
```
Ushie, line 8: I don't think on push is necessary, the code is already pushed

And this would duplicate runs on PRs from dev to main, which already covers the use of this

Ushie (changes requested):

.github/workflows/analyzer.yml
validcube: This will cause the analyser to run on unintended branches like `compose-dev`

.github/workflows/analyzer.yml
validcube: There are some cases where we would wanna push directly to the repository rather make a pull request, could be situations like—critical bug fixes, strings changes, miscellaneous stuff.

We would wanna have the analyser running at all time so that we can check if one of our commit can causes build to fail.

validcube:
We could set the `fail-on` field to `warning` instead of `info` to allow (slight) leniency towards code styling.

This will cause the analyser to fail when the analyser detected an error(s) within the code.

We can also additionally set required status check for the CI too.

.github/workflows/analyzer.yml
Ushie: No because the workflow isn't in that branch and also even if it is, there's the path check

validcube: Limit to only supported branch is safer than not specifying branch at all; in the future we might introduce analyser for Compose branch or something unrelated to Flutter.

Ushie: This workflow won't run in the compose-dev branches because the workflow file isn't in the branch

validcube: Yep, but in the future we might introduce Compose analyser which can cause problems with running Dart-specific command.

validcube:
Only to realised that it get superseded by my own PR #1521

PR #1560, opened 2023-12-09
feat: dialogs correctly follows Material 3 specifications

validcube:
# Nuke:tm: Task Force

This Nuke Operation deprecates `CustomMaterialButton` and migrates them to native

Additionally, I also remove the custom colour for the dialog.

## 🔘 Button
Button with `CustomMaterialButton` will be replaced with `FilledButton`

**Before**
```dart
CustomMaterialButton(
  label: I18nText('okButton'),
  onPressed: () {
    setPatchesChangeWarning(noShow.value);
    Navigator.of(context).pop();
  },
) 
```

**After**
```dart
FilledButton(
  onPressed: () {
    setPatchesChangeWarning(noShow.value);
    Navigator.of(context).pop();
  },
  child: I18nText('okButton'),
)
```

`CustomMaterialButton` with `isFilled: false` will be replaced with `TextButton`

The `isExpanded` value is ignored


lib/ui/widgets/shared/custom_material_button.dart
```diff
@@ -1,6 +1,7 @@
 import 'package:flutter/material.dart';
 
 class CustomMaterialButton extends StatelessWidget {
+  @Deprecated('Use the native FilledButton instead')
```
Ushie, line 4: Is deprecation even needed? given that this isn't a library, and all instances of CustomMaterialButton have been changed in this PR (supposedly)

validcube: I realised that, well fair enough, I'll remove it

Ushie (approved):

PR #1576, opened 2023-12-17
feat: support down to Android 6

validcube:
ReVanced Manager [can and will work perfectly fine in SDK 21](https://docs.flutter.dev/reference/supported-platforms "Flutter supported platform") (Android Lollipop - 5.0), we don't use native API calls that are limited to just Android 8 or newer, Android 6 sounds like a perfect place to start

People can still use custom patches & integration to patch their favourite applications in older versions.

*hell, Android 5 or KitKat might work if you're desperate enough*

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -38,7 +38,7 @@ android {
     }
     defaultConfig {
         applicationId "app.revanced.manager.flutter"
-        minSdk 26
+        minSdk 23
         targetSdk 34
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -27,6 +27,7 @@ android {
     ndkVersion flutter.ndkVersion
 
     compileOptions {
+        coreLibraryDesugaringEnabled true
         sourceCompatibility JavaVersion.VERSION_17
         targetCompatibility JavaVersion.VERSION_17
     }
@@ -42,6 +43,7 @@ android {
         targetSdk 34
         versionCode flutterVersionCode.toInteger()
         versionName flutterVersionName
+        multiDexEnabled true
     }
     buildTypes {
         release {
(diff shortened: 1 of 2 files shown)


Ushie:
If I remember correctly, patcher and patches don't work in an android 8> environment, due to them using a Files API or something similar

@kitadai31 has achieved compatibility with lower android versions though, it was something called desugaring which basically backports those APIs or something

@kitadai31 can you share if you've faced any problems caused by desugaring and any cons compared to upstream?

kitadai31:
ReVanced Manager works on Android 5.0 completely.
Desugaring has no disadvantages other than slightly increasing APK size.

Some patch is not works on Android 7 or below due to Java nio2 File API.
This applies to some resource patches for YouTube, and `Dynamic color` patch of Twitter. These patches uses File API.
But both patches actually requires newer Android version (A8 for YouTube, A12 for Dynamic color), so this is not a big problem.

But this PR is not enough to make manager works on Android 6.

- apksig won't work on Android 6. You should to switch to [MuntashirAkon/apksig-android](https://github.com/MuntashirAkon/apksig-android)
- flutter_plugin_device_apps needs a fix (https://github.com/kitadai31/flutter_plugin_device_apps/commit/a8bff360982d7acb545b97c19c221560bc5ffa91)
- Use `com.android.tools:desugar_jdk_libs_nio:2.0.4` for the desugaring library because nio is used in revanced-patcher
- multiDex is not needed if you set minSdk to 21+

This is an example implement.
https://github.com/kitadai31/revanced-manager-android5-7/commit/41b3075cf5c069575103efa19401966b0ec61879

Or could you let me open new PR? I can do this.

Ushie:
As long as those issues with the File API are present, I don't believe we should ship ReVanced Manager to android versions where those patches would fail

Is there anyway to "desugar" the File API?

kitadai31:
> Is there anyway to "desugar" the File API?

I have no idea

But I don't think we should cancel this PR because of File API
Because, patching tools should be independent of patches
In fact, there are patch bundles that don't use the File API
[kitadai31/revanced-patches-android6-7](https://github.com/kitadai31/revanced-patches-android6-7)
[d4n3436/revanced-patches-android5](https://github.com/d4n3436/revanced-patches-android5)

I think the patches' problem of not supporting old platform should be fixed on the patches side

Ushie:
@oSumAtrIX thoughts? does patcher use any APIs that'd be problematic like the Files API?

oSumAtrIX:
ReVanced Patcher targets a specific Java API. If the Android version supports that, it should work fine. Am I understanding the question correctly?

kitadai31:
> does patcher use any APIs that'd be problematic like the Files API

It's yes but patcher will be desugared correctly
Because it is integrated in the Manager APK as a dependent library

The only problem is patches
We can't desugar the supplied external dex file

PR #1606, opened 2023-12-29
ci: Build pre-release versions

BenjaminHalko:
Uses [semantic release](https://github.com/semantic-release/semantic-release/tree/master) to build prereleases from dev. (it is now used when making non-pre-release builds as well)

TODO:
- [x] Testing


BenjaminHalko:
[Testing](https://github.com/BenjaminHalko/revanced-manager/actions/runs/7360174911) on my own fork seems to be successful.

PalmDevs:
Does the built APK match the bumped version? Looking into this, it may be one version behind.

BenjaminHalko:
I think you may be correct... `flutter build apk` should be a part of `.releaserc` and not the workflow, so that it gets executed after the version bump.

Ushie:
@PalmDevs What were the complications we had when we wanted semantic releases?

PalmDevs:
> I think you may be correct... `flutter build apk` should be a part of `.releaserc` and not the workflow, so that it gets executed after the version bump.

We can run semantic-release twice with different configs. The first time is run before building which is to bump pubspec.yaml. The second time is to finally upload.

PalmDevs:
> @PalmDevs What were the complications we had when we wanted semantic releases? 

Same issue we currently have at the moment.

Ushie:
How does this plan to handle version code and version name bumps?

PalmDevs:
> How does this plan to handle version code and version name bumps?

It handles version names as is, version code will be incremented by one from what we currently have.

> Version names are decided by semantic release, but we went with a stupid design for version code and I don't think it incorporates dev version

The plugin in this PR bumps by one from what we currently have. Development versions will always be ahead or equal in version code unless someone decides to push to the main branch which we have restrictions for.

> We also need to answer the question of when does the version code bumps happen? 

Before the build process:

1. Bump everything and commit
2. Build APK
3. Upload, generate changelogs, and release

Ushie:
> The plugin in this PR bumps by one from what we currently have. Development versions will always be ahead or equal in version code unless someone decides to push to the main branch which we have restrictions for. 

Every single build is gonna bump by 1, so dev 1 and dev 2 both bump the version codes I assume?

PalmDevs:
> > The plugin in this PR bumps by one from what we currently have. Development versions will always be ahead or equal in version code unless someone decides to push to the main branch which we have restrictions for. 
> 
> Every single build is gonna bump by 1, so dev 1 and dev 2 both bump the version codes I assume? 

Yep, every trigger/build = 1 bump

oSumAtrIX:
I remind here that ReVanced Manager is not a valid candidate for semantic release

BenjaminHalko:
Ok, I think this may work now.

BenjaminHalko:
Ok I think it works now!

.releaserc.bump.json
oSumAtrIX, line 1: Whats the reason for separating the releaserc file to multiple?

BenjaminHalko: It had to be split into 2 steps. 1 for getting the version. Then 1 for uploading after building.

oSumAtrIX (approved):

PR #1615, opened 2024-01-03
feat: Add API migration code

oSumAtrIX:
This PR adds code to migrate the deprecated API to the new one, if not done yet, once.


lib/services/manager_api.dart
```diff
@@ -5,6 +5,7 @@ import 'package:device_info_plus/device_info_plus.dart';
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_i18n/widgets/I18nText.dart';
+import 'package:fluttertoast/fluttertoast.dart';
```
BenjaminHalko, line 8: You may instead want to import `package:revanced_manager/services/toast.dart` instead, as `showBottom` is not part of `fluttertoast`.

BenjaminHalko (changes requested):

BenjaminHalko (approved):
LGTM

lib/services/manager_api.dart
```diff
@@ -82,6 +94,7 @@ class ManagerAPI {
     }
     await _revancedAPI.clearAllCache();
     await _prefs.setString('apiUrl', url);
+    _toast.showBottom('settingsView.restartAppForChanges');
```
oSumAtrIX, line 97: I was wondering, if restarting is actually necessary?

BenjaminHalko: Is there an alt api I can use to test?

BenjaminHalko: Because it is hard to tell since manager caches all data.

oSumAtrIX: You can technically deploy the API locally. Alternatively the endpoint `deimos.revanced.app` exists.

BenjaminHalko: I can confirm that reloading is necessary, most likely because of the `dio` initializing in `revanced_api.dart`. Trying to re-initialize gives the error about dio already being initialized.

PR #1616, opened 2024-01-03
feat: Improve app selector UI

oSumAtrIX:
This PR improves the app selector app item UI by simplifying it:

 Before | After 
-|-
image|image
image|image

## Changes:

- Improved the suggested version chip padding
- Show the suggested version chip only if it does not match the installed version and is not "any version" in the patcher screen
- Moved patch count and version next to app name
- Corrected sizes and paddings
- Simplify strings such as "Select application" -> "Select app"

## Todo:

- Make this look better:
   image

- Correct sizing/padding:
   image

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -72,9 +72,9 @@
         "widgetTitle": "Patcher",
         "patchButton": "Patch",
 
-        "armv7WarningDialogText": "Patching on ARMv7 devices is not yet supported and might fail. Proceed anyways?",
+        "armv7WarningDialogText": "Patching on ARMv7 devices is not yet supported and might fail. Continue anyways?",
 
-        "removedPatchesWarningDialogText": "The following patches have been removed since the last time you used them.\n\n{patches}\n\nProceed anyways?",
+        "removedPatchesWarningDialogText": "The following patches have been removed since the last time you used them.\n\n{patches}\n\nContinue anyways?",
         "requiredOptionDialogText" : "Some patch options have to be set."
     },
     "appSelectorCard": {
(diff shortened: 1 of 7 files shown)


Ushie:
> Show the suggested version chip only if it does not match the installed version and is not "any version"

Ideally the chip would be kept, or displayed another way, as it is the only "official" way of sourcing the APK file

Considering patching from installed (playstore split APKs) is not supported, this should be resolved before merging this PR

oSumAtrIX:
I'd say a click on the card can show a popup that replaces the toast with a description and offers a button that calls the same action as the chip for non installed apps

oSumAtrIX:
The same popup would be able to conveniently show available patches for the app in addition and kill two birds with one stone by doing that

oSumAtrIX:
@Ushie For apps that are already installed, this idea is a problem, as you may want to select that installed app. Say, the suggested version of the app is installed. Then there would not be any motivation for the user to use the suggested version chip button regardless. So keeping it on the card, or not, would make no difference here.

Ushie:
That is true, but with the current design it would be quite terrible UX wise, for a change like this to make sense, there would have to be many changes

To begin with, you'd need to remove the concept of an "installed app", and move it back into a secondary form of selection, at the moment it is the other way around

By making selecting from storage the primary action, ALL app cards primary action would now be the search query in a webview, or the likes, and then handle the APK download

Alternatively, you would just keep the search query logic as is right now, without using a webview, and request the user to select the APK file once they return to the app

oSumAtrIX:
The only solution would be to show a chip for every app, even when the suggested version matches the installed version. This is a problem because it is redundant to show. Additionally the not installed app chip only shows a toast instead of showing a popup and offering a solution which is to launch a search query to download it

PR #1619, opened 2024-01-04
fix: incorrect strings and logics

TheAabedKhan:


oSumAtrIX (commented):
Casing has been a problem twice now. User input should be sanitized in a way so that casing does not affect it. Please add the necessary changes for that

validcube (approved):

PR #1633, opened 2024-01-12
feat: Hide the Install button during installation

### LINKED ISSUE #1629: bug: Disable the install button when installation is in progress
### Bug description

Once patching is completed and the user hits install, the manager should either gray out or the install button should disappear and reappear depending on the outcome, i.e, if the installation failed or if it didn't,  the usual open button.

### Version of ReVanced Manager and version & name of application you tried to patch

ReVanced Manager v1.18.0

### Installation type

Non-root

### Device logs

```shell
Nil
```

### Patcher logs

_No response_

### Acknowledgements

erayerdin:
Closes #1629

Checked on emulator to test the behavior.

One strange behavior is that the install button does not grey-out for some reason when it's set to null.


lib/ui/views/installer/installer_view.dart
```diff
@@ -22,24 +22,26 @@ class InstallerView extends StatelessWidget {
...
+                          ..cleanPatcher();
+                        Navigator.of(context).pop();
+                      },
+                      elevation: 0,
+                    )
+                  : FloatingActionButton.extended(
+                      label: I18nText('installerView.installButton'),
+                      icon: const Icon(Icons.file_download_outlined),
+                      onPressed: model.isInstalling
+                          ? null
```
erayerdin, line 41: It's been a while since I didn't write anything in Flutter (3 months or sth). AFAIK, passing `null` to `onPressed` should also grey-out its styling, but this does not, for some reason (new Metarial 3 thing I suppose).

It behaves, tho.

lib/ui/views/installer/installer_viewmodel.dart
```diff
@@ -453,6 +454,7 @@ class InstallerViewModel extends BaseViewModel {
   }
 
   Future<void> installResult(BuildContext context, bool installAsRoot) async {
```
erayerdin, line 456: I was lazy to comprehend all the `try-catch` thing, so I've set `isInstalling` in each reasonable line in `installResult` method.

lib/ui/views/installer/installer_view.dart
Ushie: FABs can't be disabled

erayerdin: @Ushie I can apply a custom disabled styling or hide FAB altogether. Would you like me to? :)

lib/ui/views/installer/installer_view.dart
```diff
@@ -24,7 +24,7 @@ class InstallerView extends StatelessWidget {
             floatingActionButton: Visibility(
               visible: !model.isPatching && !model.hasErrors,
               child: model.isInstalled
-                  ? FloatingActionButton.extended(
+                  ? HapticFloatingActionButtonExtended(
```
erayerdin, line 27: Changed to `HapticFloatingActionButtonExtended` to resolve merge conflict with `dev` branch.

lib/ui/views/installer/installer_view.dart
Ushie: Hi, sorry for the late reply

By material design guidelines, FABs can be hidden but not disabled, so it'd certainly be better to go with hiding it

erayerdin: Thanks for the feedback. Will apply the patch as soon as I'm available. :)

erayerdin: Applied the patch.

Ushie (approved):
Thank you for the contribution! 🎉

PR #1650, opened 2024-01-23
build: Enable ProGuard

kitadai31:
Enable ProGuard like compose manager.

The size of release APK: 45.7MB -> 39.9MB (5.8MB reduced)

## Motivation
5.8MB is not a small change.

Each version of ReVanced Manager is being downloaded 100,000-300,000 times.
5.8MB * 200,000 times = 1,160,000MB ≒ 1.1TB
So, this reduces 1.1TB of GItHub's traffic from each new versions.

Therefore, if you are planning to continue to release Flutter manager, I think this is effective.

## Other
Also, this PR disables Jetifier.
because all dependent libraries support AndroidX. (checked by `gradlew checkJetifier`)
This solves this build error:

```
[Fatal Error] :-1:-1: Invalid byte 1 of 1-byte UTF-8 sequence.
WARNING: [XmlResourcesTransformer] Received malformed sequence exception when trying to detect the encoding for 'AndroidManifest.xml'. Defaulting to UTF-8.
WARNING: [XmlResourcesTransformer] javax.xml.stream.XMLStreamException: ParseError at [row,col]:[1,1]
(followed by a very long stack trace...)
```

## Test
I tested all features of ReVanced Manager and checked works correctly.
This change was already introduced on my fork ReVanced Manager a month ago, which is 2500 downloads, and any problems are not reported.


android/app/proguard-rules.pro
```diff
@@ -0,0 +1,31 @@
...
+# You can control the set of applied configuration files using the
+# proguardFiles setting in build.gradle.kts.kts.
+#
+# For more details, see
+#   http://developer.android.com/guide/developing/tools/proguard.html
+
+-dontobfuscate
+
+# Required for the patcher to function correctly
+-keep class app.revanced.patcher.** {
```
oSumAtrIX, line 11: Why is this necessary?

kitadai31: I copied it from compose manager
but yes, there are no reason to disable obfuscation

oSumAtrIX: This should be clarified first. @Axelen123 Whats the reason for this?

android/app/proguard-rules.pro
```diff
@@ -0,0 +1,26 @@
...
+}
+
+-keepnames class com.google.common.collect.**
+-keepnames class org.xmlpull.** {
+  *;
+}
+
+-dontwarn java.awt.**
+-dontwarn javax.**
+-dontwarn com.google.j2objc.annotations.*
```
oSumAtrIX, line 26: Please comment on why these packages specifically are kept.

oSumAtrIX (changes requested):

android/app/proguard-rules.pro
kitadai31: OK.

validcube (approved):
💫 Approved

PR #1654, opened 2024-01-24
feat: Show a dialog when an update is available

### LINKED ISSUE #1637: feat: Show a dialog when an update is available
### Feature description

A dialogue should be shown when an update is available.

### Motivation

It has been observed that many people fail to see when an update is available because only a button is being enabled. A dialogue would solve this problem 

### Additional context

The dialogue should only show up when auto updates are enabled. The dialogue should contain a checkmark to not show again, which would disable the autoupdates.

### Acknowledgements

TheAabedKhan:
This PR fixes #1637 and also blocks any API requests until the users agree to download ReVanced Patches.

Screenshots
<image height="640" width="572" src="https://github.com/ReVanced/revanced-manager/assets/39409020/a67e01ba-92db-4771-871f-af3777ca74b1" alt="update-dialog"/>

### DIFF
--- a/assets/i18n/en_US.json
+++ b/assets/i18n/en_US.json
@@ -16,6 +16,8 @@
     "noShowAgain": "Don't show this again",
     "add": "Add",
     "remove": "Remove",
+    "showChangelogButton": "Show changelog",
+    "showUpdateButton": "Show update",
     "navigationView": {
         "dashboardTab": "Dashboard",
         "patcherTab": "Patcher",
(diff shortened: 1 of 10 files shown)


oSumAtrIX:
Some questions regarding behaviour are important to know before merging this PR. I will list some scenarios this PR should fulfil.

I download an old version of the app for the first time. It first shows a popup that asks the user if they want the app to check for updates. If the user presses No, no requests are made.
Because no request are made, the app has to ask the user for permission to connect to the internet in order to download the patches. The user can not proceed using the app if the press no.

If they click yes, the app does not check for updates and no popup is shown. If I accepted to check for updates, the popup that asks the user to download patches is not shown, instead a popup is shown that asks if the user wants to update. The popup is accompanied by a No button and a checkmark not to show the popup again.

If the user clicks yes, patches as well as the app updates are installed. If the user does not press Ok, the Download buttons in the main menu for the app and the patches are enabled, so the user can manually update in case they want.

oSumAtrIX (approved):

Issue #1645, opened 2024-01-21
bug(dev branch): User is able to prematurely exit out of the patching screen 


validcube:
The change is likely the cause of the PopScope Migration

RVMF version: acb1e2434b28efca852297e0a0a8ffce155e67e4

validcube:
https://github.com/ReVanced/revanced-manager/assets/93124920/2b3c01b8-09e7-4aae-a42f-78091a498acd

Ushie:
Given this issue existed only in the dev branch, it makes sense to be closed as completed

PR #1663, opened 2024-01-29
fix(PopScope): User able to exit patch screen when the installer is still running

BenjaminHalko:
fixes: #1645

### DIFF
--- a/lib/ui/views/installer/installer_view.dart
+++ b/lib/ui/views/installer/installer_view.dart
@@ -17,6 +17,8 @@ class InstallerView extends StatelessWidget {
       onViewModelReady: (model) => model.initialize(context),
       viewModelBuilder: () => InstallerViewModel(),
       builder: (context, model, child) => PopScope(
+        onPopInvoked: (bool didPop) => model.onWillPop(context),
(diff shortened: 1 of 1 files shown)


validcube:
Please rebase this on the latest commit.

lib/ui/views/installer/installer_view.dart
```diff
@@ -17,6 +17,8 @@ class InstallerView extends StatelessWidget {
       onViewModelReady: (model) => model.initialize(context),
       viewModelBuilder: () => InstallerViewModel(),
       builder: (context, model, child) => PopScope(
+        onPopInvoked: (bool didPop) => model.onWillPop(context),
+        canPop: false,
```
validcube, line 21: The code seems skeptical to me.

Can you exit out of the screen when you finished patching?

BenjaminHalko: Yep, it does work. At least on my device.

validcube: > Yep, it does work. At least on my device.

I'll revisit this later.:tm:

PR #1664, opened 2024-01-29
ci: Specify paths in release workflow

BenjaminHalko:
Only triggers the release workflow when certain files are changed.


validcube (approved):
🥞 LGTM!

PR #1674, opened 2024-02-03
refactor: PopScope Migration

validcube:
We've tried migrating this two times 💀, and reverted it two times 💀 

for the same reason: unable to exit out of the patching screen.


validcube:
🥞 Approved, (tested)

PR #1692, opened 2024-02-18
ci: Accept more commits + Dependabot

BenjaminHalko:
Allows ci to build on commits with types such as `refactor`.

Also, while making changes, I noticed there were 2 high level vulnerabilities in the npm packages. There seems to be no fix for them at the moment, but I also added Dependabot so that they can be fixed as soon as possible.

### DIFF
--- a/.github/workflows/release-build.yml
+++ b/.github/workflows/release-build.yml
@@ -27,7 +27,7 @@ jobs:
           distribution: "zulu"
 
       - name: Set up Node.js 20
-        uses: actions/setup-node@v2
+        uses: actions/setup-node@v4
         with:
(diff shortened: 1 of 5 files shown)


.github/workflows/release-build.yml
```diff
@@ -26,6 +26,11 @@ jobs:
           java-version: "17"
           distribution: "zulu"
 
+      - name: Set up Node.js 20
+        uses: actions/setup-node@v2
```
validcube, line 30:
```suggestion
        uses: actions/setup-node@v4
```

validcube (approved):
🥞 Explod

PR #1695, opened 2024-02-18
ci: remove unnecessary upload

BenjaminHalko:
Build uploading is handled by semantic release.


validcube:
There are some commit which does not generate a pre-release build like chore, refactor.

Which makes testing a pain.

BenjaminHalko:
Well, now it does trigger on chore and refactor... Plus if it decides not to trigger, then there wouldn't be an APK to upload anyways...

PR #1707, opened 2024-02-26
docs: Fix consistency issues with other repositories

oSumAtrIX:
# About

- This PR changes workflow and issue file names, fixes the consistency in the style of the README, contribution guidelines, issue templates, and workflows

Some inconsistencies would be considered to be kept in this repository as they looked better than in other reports. However, restoring consistency and then improving all repositories together is a safer approach.

# Todo

- [x] Rename GitHub issue tags

# Review

- [ ] Workflows have been modified and need to be reviewed if they still work as expected
- [ ] Spelling mistakes
- [ ] Bad links in markdown or issue templates


.github/workflows/sync_crowdin.yml
oSumAtrIX, line 1: @Ushie Whats the custom workflow needed for again?

.github/workflows/sync_crowdin.yml
```diff
@@ -1,12 +1,14 @@
-name: Crowdin Action
+name: Sync Crowdin
 
 on:
   workflow_dispatch:
   push:
-    branches: [dev]
+    branches:
+      - dev
```
oSumAtrIX, line 7: @Ushie  This is slightly problematic, the workflow shouldn't run on every commit, maybe there's a better trigger?

.github/workflows/sync_crowdin.yml
```diff
@@ -26,10 +28,10 @@ jobs:
           download_translations: true
           localization_branch_name: feat/translations
           create_pull_request: true
-          pull_request_title: "chore(i18n): Sync translations"
-          pull_request_body: "Sync translations from [Crowdin/ReVanced](https://crowdin.com/project/revanced)"
+          pull_request_title: "chore: Sync translations"
+          pull_request_body: "Sync translations from [crowdin.com/project/revanced](https://crowdin.com/project/revanced)"
           pull_request_base_branch_name: "dev"
-          commit_message: "chore(i18n): Sync translations"
```
oSumAtrIX, line 32: We do not have any particular purpose for these scopes yet, keeping it simple by removing them

.github/workflows/sync_crowdin.yml
```diff
@@ -38,16 +40,15 @@ jobs:
       - name: Remove empty values from JSON
         run: |
           cd assets/i18n
-          sudo chmod 777 *
```
oSumAtrIX, line 41: Was this necessary @Ushie ?

.github/workflows/sync_crowdin.yml
```diff
@@ -38,16 +40,15 @@ jobs:
       - name: Remove empty values from JSON
```
oSumAtrIX, line 40: @Ushie What is this step used for?

.github/workflows/sync_crowdin.yml
```diff
@@ -38,16 +40,15 @@ jobs:
...
 
-      - name: Push out changes to i10n
+      - name: Commit translations
         run: |
           git config user.name revanced-bot
           git config user.email github@revanced.app
           sudo chown -R $USER:$USER .git
           git add assets/i18n/*.json
-          git commit -m "chore(Translation): Remove empty values from JSON" assets/i18n/*.json
+          git commit -m "chore: Remove empty values from JSON" assets/i18n/*.json
```
oSumAtrIX, line 51: I am seeing this commit after every push to dev. Not sure what this is for but maybe a proper trigger can be made to reduce this spam @Ushie

.github/workflows/sync_crowdin.yml
```diff
@@ -26,10 +28,10 @@ jobs:
...
           create_pull_request: true
-          pull_request_title: "chore(i18n): Sync translations"
-          pull_request_body: "Sync translations from [Crowdin/ReVanced](https://crowdin.com/project/revanced)"
+          pull_request_title: "chore: Sync translations"
+          pull_request_body: "Sync translations from [crowdin.com/project/revanced](https://crowdin.com/project/revanced)"
           pull_request_base_branch_name: "dev"
-          commit_message: "chore(i18n): Sync translations"
+          commit_message: "chore: Sync translations"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
oSumAtrIX, line 36: As far as I know the checkout step will keep the credentials in cache so that other steps can use it. Is this necessary? @Ushie

.github/workflows/sync_crowdin.yml
```diff
@@ -38,16 +40,15 @@ jobs:
       - name: Remove empty values from JSON
         run: |
           cd assets/i18n
-          sudo chmod 777 *
           dart nuke.dart >> $GITHUB_STEP_SUMMARY
 
-      - name: Push out changes to i10n
+      - name: Commit translations
         run: |
           git config user.name revanced-bot
```
oSumAtrIX, line 47: Instead of setting up the user like this, we have a token for @revanced-bot That can be used instead, if necessary

README.md
```diff
@@ -60,38 +60,50 @@
...
 ## 💪 Features
 
-Some of the features we provide are:
+Some of the features ReVanced Manager provides are:
 
-* 📱 **Portable**: ReVanced Patcher that fit in your pocket;
-* 🤗 **Intuitive UI**: Help you manage your patched applications with easy-to-use interface;
-* 🛠️ **Customization**: Patch with third-party sources;
-* ✨ And a **lot more!**
+- 💉 **Patch apps**: Apply any patch of your choice to Android apps
```
oSumAtrIX, line 76: Maybe this can be improved

.github/workflows/build_pull_request.yml
```diff
@@ -50,73 +43,74 @@ jobs:
           gh repo set-default "${{ github.repository }}"
           gh pr checkout "${{ inputs.pr-number }}"
           
-          echo "DATETIME=$( TZ='UTC+0' date --rfc-email )" >> $GITHUB_ENV
```
oSumAtrIX, line 53: This didnt seem like it had a particular use

.github/workflows/build_pull_request.yml
```diff
@@ -1,23 +1,15 @@
-name: PR Build
+name: Build pull request
 
 on:
   workflow_dispatch:
     inputs:
-      # Flutter
-      flutter-branch:
```
oSumAtrIX, line 7: I removed this because it didn't look like this was necessary. We only really use stable Flutter branch

README.md
```diff
@@ -60,38 +60,50 @@
 
 # 💊 ReVanced Manager
 
-[![GitHub last commit](https://img.shields.io/github/last-commit/ReVanced/revanced-manager)](https://github.com/ReVanced/revanced-manager/commits "")
-[![GitHub commit activity](https://img.shields.io/github/commit-activity/w/ReVanced/revanced-manager)](https://github.com/ReVanced/revanced-manager/commits "")
+![GitHub Workflow Status (with event)](https://img.shields.io/github/actions/workflow/status/ReVanced/revanced-cli/release.yml)
```
oSumAtrIX, line 63:
```suggestion
![GitHub Workflow Status (with event)](https://img.shields.io/github/actions/workflow/status/ReVanced/revanced-manager/release.yml)
```

.github/workflows/sync_crowdin.yml
validcube: In my opinion the workflow should run when we dispatched it or let it automatically run on every week/month

.github/workflows/sync_crowdin.yml
validcube: I remember adding this fixes GitHub action permission issue.

I have no idea why.

.github/workflows/sync_crowdin.yml
validcube: It is pushed to the translation branch before it is squash and merged to dev, this is to remove empty string from JSON to make them cleaner.

I guess we could squash these two before pushing.

oSumAtrIX: Why would empty strings be present?

validcube: > Why would empty strings be present?

Not sure, when we exported them from Crowdin

It would look like this



We decided that it isn't really supposed to be doing that (and it could mess everything up with the language feature), so we decided to work on a postprocessor for it.

I have haven't really decided on a name so I called them Nuke:tm:, a bare-bone (probably not even finished) documentation exist here: https://github.com/ReVanced/revanced-manager/blob/dev/assets%2Fi18n%2FREADME.md

validcube:
🍜 Before merging, merge this first
* #1701

.github/workflows/sync_crowdin.yml
Ushie: > Why would empty strings be present?

Because Crowdin doesn't have a post-processor for nested JSON, so everything that isn't translated is exported as an empty string, no matter the settings in Crowdin

Ushie: > I am seeing this commit after every push to dev. Not sure what this is for but maybe a proper trigger can be made to reduce this spam @Ushie

We can change the trigger of translation sync to manual or to a CRON job

.github/workflows/sync_crowdin.yml
Ushie: Removing empty strings in the JSONs exported from Crowdin, because Crowdin doesn't have a double-nested JSON post-processor to remove them

.github/workflows/sync_crowdin.yml
Ushie: Why remove them, they can stay for consistency with other commits

.github/workflows/sync_crowdin.yml
Ushie: A manual trigger or CRON would work fine, but I feel like those can be easily forgotten

There's no reminder when using a manual trigger, and the CRON job wouldn't be likely to run during the time when someone is working on the project, by running it when pushing to dev you can be reminded by the "Recent pushes" notification

.github/workflows/sync_crowdin.yml
Ushie: It's not necessary to use the token, the token would only be used for push permissions which this workflow already has, we would still have to set the git username

CONTRIBUTING.md
```diff
@@ -71,33 +71,33 @@ This document describes how to contribute to ReVanced Manager.
...
+   it will be merged into the `dev` branch and will be included in the next release of ReVanced Manager
 
 ## 🤚 I want to contribute but don't know how to code
 
 Even if you don't know how to code, you can still contribute by 
 translating ReVanced Manager on [Crowdin](https://translate.revanced.app/).
 
-❤️ Thank you for considering contributing to ReVanced Manager.
+❤️ Thank you for considering contributing to ReVanced Manager,  
+ReVanced
```
Ushie, line 103: Why is this necessary?

.github/ISSUE_TEMPLATE/bup_report.yml
Ushie, line 1: Please correct the filename

.github/workflows/sync_crowdin.yml
validcube: > To make the script executable, I'm not the one who wrote this line so I don't know about if its necessary or not, I think you could run that locally before pushing it and it'll retain permissions

The script can execute perfectly fine, it just that for whatever reason the script can't modify the JSON file.

.github/workflows/sync_crowdin.yml
oSumAtrIX: I am saying that the trigger may be reduced to the JSON file, so when the JSON file is changed, the trigger runs.

.github/workflows/sync_crowdin.yml
oSumAtrIX: They are removed for consistency with other commits. We do not have any use for scopes. Currently, they are used randomly without any convention.

.github/workflows/sync_crowdin.yml
oSumAtrIX: Can you show an example of how such a file looks like that needs this post processing?

.github/workflows/sync_crowdin.yml
oSumAtrIX: > Because Crowdin doesn't have a post-processor for nested JSON, so everything that isn't translated is exported as an empty string, no matter the settings in Crowdin

When does this occur? When stings are deleted?

CONTRIBUTING.md
oSumAtrIX: This PR introduces consistency with other repos. Other repos have it too, so I added it here as well.

.github/workflows/sync_crowdin.yml
validcube: > Are you sure? The permissions on git look fine for this file

It is, but in GitHub Action, it doesn't. It's so confusing.

oSumAtrIX: Can you link me to a run where it occurs?

.github/workflows/sync_crowdin.yml
```diff
@@ -38,16 +40,15 @@ jobs:
       - name: Remove empty values from JSON
         run: |
           cd assets/i18n
```
oSumAtrIX, line 42:
```suggestion
          cd assets/i18n
          sudo chmod 777 *
```

PR #1750, opened 2024-03-07
fix: App list is empty if all apps are installed

Hepolise:
The `model.allApps` is a list of already filtererd applications that are not installed on device. It's needed to check both arrays if they are empty.


Aunali321:
What a stupid bug. But thanks!

Aunali321 (approved):

Issue #1735, opened 2024-03-05
feat: Export manager logs as .txt files instead of .log


KobeW50:
### Feature description

*Uhh, might be controversial.*

Export manager logs with the `.txt` file extension.

Relevant line of code: https://github.com/ReVanced/revanced-manager/blob/main/lib%2Fui%2Fviews%2Fsettings%2Fsettings_viewmodel.dart#L334

### Motivation

Almost all file readers on Android cannot open `.log` files, but users need to copy the contents of the log to their clipboard in order to paste it here in GitHub issues. This means users need to change the file extension to `.txt` in order to read and copy from the log.

### Acknowledgements

- [X] This issue is not a duplicate of an existing feature request.
- [X] I have chosen an appropriate title.
- [X] The feature request is only related to ReVanced Manager

oSumAtrIX:
> Almost all file readers on Android cannot open .log files, but users need to copy the contents of the log to their clipboard in order to paste it here in GitHub issues. This means users need to change the file extension to .txt in order to read and copy from the log.

Isn't this an issue with the respective renderers? `.log` is objectively the more correct extension than `.txt`

KobeW50:
> > Almost all file readers on Android cannot open .log files, but users need to copy the contents of the log to their clipboard in order to paste it here in GitHub issues. This means users need to change the file extension to .txt in order to read and copy from the log.
> 
> Isn't this an issue with the respective renderers? `.log` is objectively the more correct extension than `.txt`

Yes. `.log` is more correct but `.txt` simply makes it easier to read the file in most cases. 

Are there specific benefits of using `.log` aside for that it shows that the file contains log information?

oSumAtrIX:
> but .txt simply makes it easier to read the file in most cases.

That is not correct. The file extension does not change that.

> Are there specific benefits of using .log aside for that it shows that the file contains log information?

The reason why `.log` is used is because of semantics. You don't name an executable file `.txt` for the same reason you don't name a log `.txt`.

KobeW50:
> > but .txt simply makes it easier to read the file in most cases.
> 
> That is not correct. The file extension does not change that.

I mean that most file readers on Android will not open a .log file because they don't recognize the file type.

oSumAtrIX:
> I agree that it is technically an issue with the renderers. But on the other hand, users shouldn't necessarily be expected to have one of the rare Android file readers which can read .log files

This is something you'd need to sort out with the respective renderers.

PR #1752, opened 2024-03-08
feat: Export logs to `.txt` file

KobeW50:
I didn't test the changes but I'm fairly confident this is the only thing that needs to be changed.

Closes https://github.com/ReVanced/revanced-manager/issues/1735


validcube:
I'm not really sure why we would change the format of the log file to txt.

For me it doesn't really make sense to do that since log are *.log*, text are *.txt*

While it's true that Android can't read those *.log* file, perhaps we could try to upstream it to AOSP since it's technically a generalised format.

oSumAtrIX:
Thanks for the PR regardless

PR #1766, opened 2024-03-15
fix: Unable to get changelogs when using GitHub API

TheAabedKhan:


lib/services/manager_api.dart
```diff
@@ -115,6 +119,9 @@ class ManagerAPI {
   }
 
   String getPatchesRepo() {
+    if (!isUsingAlternativeSources()) {
+      return defaultPatchesRepo;
+    }
```
kitadai31, line 124: Shouldn't `getIntegrationsRepo()` also be changed?

lib/services/manager_api.dart
```diff
@@ -115,6 +119,9 @@ class ManagerAPI {
...
     return _prefs.getString('patchesRepo') ?? defaultPatchesRepo;
```
Ushie, line 125: Is that `??` necessary?

ILoveOpenSourceApplications: > Is that `??` necessary?

@TheAabedKhan.

TheAabedKhan: Not needed after committing that logic.

ILoveOpenSourceApplications:
Is this PR ready?

TheAabedKhan:
@ILoveOpenSourceApplications YES! Ready to be merged.

oSumAtrIX (approved):

Issue #821, opened 2023-04-28
feat: allow dev channel patches


SodaWithoutSparkles:
### Type

Functionality

### Issue

The dev channel (pre-releases) of integration and patches are updated far more frequently. Sometimes when there is a very annoying issue and the fix was alreadly in the dev channel, but not yet in the stable release.

### Feature

Allow pre-release integrations and patches to be used in revanced manager. This way, we could use the dev channel when needed.

### Motivation

- allows more users to test out pre-release versions for bugs
- rapid testing of new changes for targeted users/bug reporters
- solve the annoying issue of "the fix is there but you cannot use it yet because you cannot use the dev channel"
- not all people have access to a computer/termux for CLI patching

### Additional context

https://github.com/revanced/revanced-patches/issues/2003#issuecomment-1527226397

Could benefit rapid testing like that in https://github.com/revanced/revanced-patches/issues/1443

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I have written a short but informative title.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

SodaWithoutSparkles:
Potential solutions:
- use existing revanced api
    - benefits
        - using current infrastructure
    - changes
        - api backend: allow dev channel
        - manager: add a toggle to enable dev channel
- use github API
    - benefits
        - reduce revanced backend api load
        - no need to mess with current backend → backwards compatibility
    - changes
        - manager: allow github api as release api
        - manager: send and parse github api
        - manager: add toggle to enable dev channel

kitadai31:
I found a temporary workaround for this issue.

Change any one letter in sources setting to uppercase. (Each of Patches and Integrations)
Then you can use dev patches in Manager.

<img src="https://github.com/revanced/revanced-manager/assets/90122968/afdee91c-271f-4e22-aca1-4d7a135bc9d0" width="240">

<img src="https://github.com/revanced/revanced-manager/assets/90122968/b11ec18b-068e-43cf-abf0-14ce627fa1f1" width="360">

By changing the sources setting, the API used by the Manager is switched from ReVanced API to GitHub API.
In other words, Manager recognizes "revaNced/revanced-patches" as an external source.
Since the GitHub API does not distinguish between dev and release, Manager will fetch the latest release in the repository anyway.
Therefore, Manager uses the dev patches.

[...]

PR #1785, opened 2024-03-25
feat: Pre-release patches toggle

kitadai31:
Add a toggle for pre-release patches to the alternative source settings.
Closes: #821

This PR also fixes an problem where pre-release Integrations are being used when using alternate sources.
At this time, Patches are referring `GitHubAPI#getLatestPatchesRelease()` that doesn't include pre-releases, but Integrations are referring `GitHubAPI#getLatestRelease()` that includes pre-releases.
This problem actually caused patched apps to crash recently in crimera/piko patches (an alternative source) when they released a pre-releases.
(Actually, this is the main motivation)

`getLatestReleaseFile()` in `github_api.dart` was removed because it's unused


Ushie:
Why limit it to alternative sources?

You can pass ?dev=true to achieve the same with the ReVanced API

kitadai31:
I already tried ?dev=true at ReVanced API, but it didn't effect.
https://api.revanced.app/v2/revanced-patches/releases/latest?dev=true
So, I thought it wasn't implemented yet at first.
That's why it is limited to GitHub API only.  I didn't know the cache issue.
If the cache issue will be fixed, I think I should wait for it.

> Are changelogs/updates considered as well?

Yes. These lines change the patches update dialog according to the settings.
https://github.com/ReVanced/revanced-manager/blob/e88e997436d8692023d49480089fe7d71538492b/lib/ui/views/home/home_viewmodel.dart#L482-L490

brian6932:
One thing I noticed after building this branch (rebased) was that RevancedManager, gets stuck on `Loading...`, patches work fine though. Is that just a quirk of debug builds, or an issue with the PR?

nivranaitsirhc:
> One thing I noticed after building this branch (rebased) was that checking for `Revanced Manager` updates gets stuck on `Loading...`, patches work fine though. Is that just a quirk of debug builds, or an issue with the PR?

Can you share the apk here please

brian6932:
@nivranaitsirhc sure https://i.alienpls.org/ReVanced.apk

nivranaitsirhc:
> @nivranaitsirhc sure https://i.alienpls.org/ReVanced.apk

It works like a charm! Thanks!

kitadai31:
Waiting for the parameter issue in ReVanced API to be fixed

I have some questions

1. I added the ReVanced API v2 related logic in revanced_api.dart, but should I split the file into two files like `revanced_api_v2.dart` ?
2. I added a word "(dev)" in settings description because ReVanced calls pre-release versions “dev”. (40b1589)
Is this no problem?

lib/services/github_api.dart
```diff
@@ -37,6 +37,34 @@ class GithubAPI {
...
+      }
+      return latestRelease;
+    } on Exception catch (e) {
+      if (kDebugMode) {
+        print(e);
+      }
+      return null;
+    }
+  }
+
```
Ushie, line 67: Maybe instead of making a new function, add a parameter to the existing one instead

Ushie (changes requested):
@oSumAtrIX Does v3 suffer from the same issues regarding the ?dev parameter, supporting the API is a prerequisite for this PR

oSumAtrIX:
A ?dev query parameter does not exist and is not planned to be added. The TLDR is that the dev branch is not a controlled release channel like a "beta" channel would be.

Ushie:
The dev branch has nothing to do with this, the flag can be renamed to ?prereleases instead of ?dev because it simply enables returning prereleases

oSumAtrIX:
The name does not change that pre releases are not a controlled production release channel. Any further discussion should happen in the API repo and not in this PR

kitadai31:
Yeah, I've heard that ReVanced API v2 was deprecated.
And v3 doesn't have toggles for pre-releases.

So, I think it's enough to revert this PR to the beginning. (Pre-release requires alternative sources to be enabled, https://github.com/ReVanced/revanced-manager/pull/1785#issue-2205094068)

PR #1834, opened 2024-04-03
Fix typos regarding cancelling and cancelled

HiFIi:


oSumAtrIX (approved):

validcube:
Canceled is correct, they're commonly use in American version of English, while Cancelled is commonly use in British version of English.

Issue #1854, opened 2024-04-05
bug: crash on Android 14 when try to start FGS


Howard20181:
### Bug description

Patch any app on Android 14 can reproduce

### Version of ReVanced Manager and version & name of app you are patching

1.20.0-dev.1
YouTube 19.09.37

### Installation method

Regular

### ReVanced Manager logs

```shell
Generated by AppErrorsTracking 1.3(6)
Project URL: https://github.com/KitsunePie/AppErrorsTracking
===============
[Device Brand]: Xiaomi
[Device Model]: 23127PN0CC
[Display]: UKQ1.230804.001 release-keys
[Android Version]: 14
[Android API Level]: 34
[System Locale]: zh_CN
[Process ID]: 24983
[User ID]: 0
[CPU ABI]: arm64-v8a
[Package Name]: app.revanced.manager.flutter
[Version Name]: 1.20.0-dev.1
[Version Code]: 101800014
[Target SDK]: 34
[Min SDK]: 26
[Error Type]: JVM
[Crash Time]: 2024-04-05T09:53:16.462
[Stack Trace]:
java.lang.RuntimeException: Unable to start service de.julianassmann.flutter_background.IsolateHolderService@243de with Intent { act=START cmp=app.revanced.manager.flutter/de.julianassmann.flutter_background.IsolateHolderService }: android.app.MissingForegroundServiceTypeException: Starting FGS without a type  callerApp=ProcessRecord{45d88a8 24983:app.revanced.manager.flutter/u0a483} targetSDK=34
	at android.app.ActivityThread.handleServiceArgs(ActivityThread.java:5046)
	at android.app.ActivityThread.-$$Nest$mhandleServiceArgs(Unknown Source:0)
	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2418)
	at android.os.Handler.dispatchMessage(Handler.java:106)
	at android.os.Looper.loopOnce(Looper.java:224)
	at android.os.Looper.loop(Looper.java:318)
	at android.app.ActivityThread.main(ActivityThread.java:8677)
	at java.lang.reflect.Method.invoke(Native Method)
	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:561)
	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1013)
[...]

Issue #1841, opened 2024-04-04
bug: Crashes upon pressing the "patch" button on Android 14


dwitterer:
### Bug description

I can launch ReVanced Manager, navigate to any of the screens, and get all the way to the point of selecting which patches to apply, but then whenever I press 'Patch', the app immediately crashes and I get a pop-up that 'ReVanced Manager keeps crashing'. Rebooting, re-installing the Manager and patches, and switching to ReVanced GMSCore did not fix the issue. The Manager dashboard also always says that my patcher was last updated 26 days ago, which is not true - unsure of this is relevant. Logs attached below.

### Version of ReVanced Manager and version & name of app you are patching

ReVanced Manager 1.19.4-dev.5, release build (but earlier versions such as 1.19.3 also do the same thing)
Android 14, OneUI 6.1 (Samsung S22 Ultra)
Patches 4.6.0
Trying to patch YouTube 19.09.37

### Installation method

None

### ReVanced Manager logs

```shell
[...]
04-04 21:58:38.969 12718 12744 W vulkan  : CreateSwapchainKHR(VkSwapchainCreateInfoKHR.imageColorSpace = 1000104013) failed: Unsupported color space, but assume as Success for ANGLE.
04-04 21:58:38.971 12718 12718 I SurfaceView: 162898386 surfaceChanged -- format=4 w=1440 h=3088
04-04 21:58:38.971 12718 12718 I SurfaceView@9b5a1d2: surfaceChanged (1440,3088) 1 #8 io.flutter.embedding.android.m{9b5a1d2 V.E...... ......ID 0,0-1440,3088}
04-04 21:58:38.971 12718 12718 I SurfaceView: 162898386 surfaceRedrawNeeded
04-04 21:58:38.971 12718 12718 I SurfaceView: 162898386 finishedDrawing
04-04 21:58:38.971 12718 12718 V SurfaceView@9b5a1d2: Layout: x=0 y=0 w=1440 h=3088, frame=Rect(0, 0 - 1440, 3088)
04-04 21:58:39.307 12718 12718 D InsetsSourceConsumer: applyRequestedVisibilityToControl: visible=true, type=1
04-04 21:58:39.308 12718 12718 D InsetsSourceConsumer: applyRequestedVisibilityToControl: visible=true, type=2
04-04 21:58:39.647 12718 12743 I flutter : dynamic_color: Core palette detected.
[...]

Issue #1839, opened 2024-04-04
In android 14, apps are automatically closed


la8896:
### Bug description

In android 14, apps are automatically closed when running the patch in v1.19.4-dev.x, but they can be used normally in v1.19.3

### Version of ReVanced Manager and version & name of app you are patching

All v1.19.4-dev.x version 

### Installation method

Regular

### ReVanced Manager logs

```shell
Nil
android 14 prompt close apps
```

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The bug is only related to ReVanced Manager

PR #1856, opened 2024-04-05
fix: SecurityException when patching application

validcube:
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://raw.githubusercontent.com/validcube/validcube/main/assets/ReVancedManager/dynamic-dark.webp">
  
</picture>

## 😊 Issue
Multiple users told us about ReVanced Manager v1.19.4/v1.12.0-dev.X getting killed in Android 14 when attempting to patch. We've found the problem to be originating from one of the dependency that we happened to be using, [flutter_background](https://github.com/JulianAssmann/flutter_background).

Starting In Android 14, all foreground services must list at least one foreground service type or else system will raise `SecurityException`.
* https://developer.android.com/about/versions/14/changes/fgs-types-required

This issue happen on https://github.com/JulianAssmann/flutter_background/commit/65cf0ec93486b3a6c74339aadaeb384cf2c9a24b and newer, so a PR to fix this have been opened
* https://github.com/JulianAssmann/flutter_background/pull/83

The relevant code was line 7 in [Manifest](https://github.com/ReVanced/revanced-manager/blob/main/lib/ui/views/settings/settingsFragment/settings_manage_sources.dart#L30-L40) and line 128 in [IsolateHolderService](https://github.com/JulianAssmann/flutter_background/blob/65cf0ec93486b3a6c74339aadaeb384cf2c9a24b/android/src/main/kotlin/de/julianassmann/flutter_background/IsolateHolderService.kt#L128)

https://github.com/JulianAssmann/flutter_background/blob/65cf0ec93486b3a6c74339aadaeb384cf2c9a24b/android/src/main/AndroidManifest.xml#L6-L8

https://github.com/JulianAssmann/flutter_background/blob/65cf0ec93486b3a6c74339aadaeb384cf2c9a24b/android/src/main/kotlin/de/julianassmann/flutter_background/IsolateHolderService.kt#L128

## 🔧 Testing note
Testing was done in [Firebase Device Streaming](https://developer.android.com/studio/preview/android-device-streaming), which means everything happened remotely in Google's servers... alternatively a physical device farm.

| 📱 Device                           | ⚙️ SDK                         | 📝 Human Verdict |
| ---------------------------- | ------------------------- | --------------------- |
| Samsung Galaxy Z Fold 5 | Upside Down Cake (34) | ✅ Passed               |

https://github.com/ReVanced/revanced-manager/assets/93124920/5518b5c9-815c-40e5-986e-002d586605d0

fix: #1854, #1841, #1839


Domenic-MZS:
It's just me (fast check not totally sure) but Why are we missing the android permission for the foreground tho?

```xml
android:name="android.permission.FOREGROUND_SERVICE"/>
```

Which is required for the [flutter_background](https://github.com/JulianAssmann/flutter_background) package

validcube:
> (fast check not totally sure) why are we missing the android permission for the foreground tho?
> 
> ```
> <uses-permission android:name="android.permission.FOREGROUND_SERVICE"/>
> ```
> 
> Which is required by the [flutter_background](https://github.com/JulianAssmann/flutter_background) package

That's a good question, I'm finding a reason to git blaming someone in team now :D - aside from that, yeah we should add it.

PR #1859, opened 2024-04-07
fix: Increase dashboard RefreshIndicator edge offset

### LINKED ISSUE #1568: feat(Refresh Animation): Improve the refreshing animation
### Feature description

When refreshing `Dashboard`, the refresh animation that shows up is an incomplete circle animation instead of the traditional circling cycle animation on  every other apps. (Don't judge me for phrasing it this way.)

What ReVanced Manager does:

https://github.com/ReVanced/revanced-manager/assets/117499019/b3601f3e-d398-4cf9-96e0-965141a7fbd8

What it should be doing:

https://github.com/ReVanced/revanced-manager/assets/117499019/d8a8207f-e3e0-4345-a464-f486c29fe7bc

Also, the refresh icon starts showing up all the way from the notification shade instead of showing up within the manager. It should start from the point in the `Dashboard` where the padding is when scrolling down.

Without scrolling:
Screenshot_20231213_114911_ReVanced Manager

When scrolled down a bit:
Screenshot_20231213_114730_ReVanced Manager

The refresh icon should start after the padded area of `Dashboard`, like it's coming from underneath the padded area.

### Motivation

I checked out the examples provided in the repository and [`MaterialClassicHeader`](https://github.com/peng8350/flutter_pulltorefresh/blob/master/lib/src/indicator/material_indicator.dart) seems to be the one which was implemented in ReVanced Manager (or looks like what was intended). Correct me if I'm wrong.
P.S: Attaching the gif for further clarity
material_classic

### Additional context

_No response_

### Acknowledgements

Domenic-MZS:
# 🔧🎨 UI Fix: Increase dashboard RefreshIndicator edge offset

This change adjusts the edge offset for the RefreshIndicator in the home view dashboard UI with the CustomSliverAppBar maxExtent size, ensuring a better user experience by moving the RefreshIndicator under the SliverAppbar.

> ##### Fixes #1568

### Overview
The HomeView / Dashboard `[CustomScrollView]` refresh functionality has been enhanced by matching the starting refresh position with the `[SliverAppBar]` height, and by reducing the displacement range to settle the refresh.

- **Issue:** The RefreshIndicator starts the animation on top of the `[SliverAppBar]`, disturbing the user experience (?).
- **Context:** The current functionality lacks an increased starting point (`[edgeOffset]`) or a better layout.
- **Related Solutions:** (None).
- **Impact:** Improves the user experience by making the refresh indicator more concise and easy to use by reducing the needed displacement to work and by correct positioning the indicator on the affected area (the body and not the header).

### 📓 Description
This PR sets the Dashboard `[edgeOffset]` (indicator starting point) to match the `[CustomSliverAppBar]` (title) height and changes the default `[displacement]` to execute the refresh callback.

### ✍🏼 Changes Made
- Added property (`[edgeOffset]`)[edge-offset].
- Reduces the refresh indicator needed `[displacement]` to settle.

<details>
<summary>Technical Overview</summary>
> (For the RefreshIndicator widgtet wrapping the content)
- Add the `edgeOffset` property with the `CustomSliverAppBar` hardcoded `extendedHeight`. (~magic number)
- Add the `displacement` property to 50. (reduces the needed scroll to settle the "refresh" action)
</details>

### 🦯 Testing
- [x] ~Update Tests (ToDo - Not Found nor needed for now...)~
- [x] Compile/Build code
  - 🐞 Tested with Stable Flutter v3.19.3 and dart 3.3.1.
  - **System:** Linux archlinux 6.7.9-arch1-1 GNU/Linux
 
https://github.com/ReVanced/revanced-manager/assets/75044136/c028c996-d467-442b-ad24-020dd439524e

### 📋 Notes & References
[RefreshIndicator - edgeOffset](https://api.flutter.dev/flutter/material/RefreshIndicator-class.html#:~:text=The%20offset%20where%20RefreshProgressIndicator%20starts%20to%20appear%20on%20drag%20start)
Current [Custom Sliver App Bar - ExpandedHeight](https://github.com/ReVanced/revanced-manager/blob/a23f032fd24017286391a94dfa924e7256ff506c/lib/ui/widgets/shared/custom_sliver_app_bar.dart#L20-L23)

### 🔬 Reviewers
N/A (any mod is ok)

---

👾 Feedback, changes, or suggestions are welcome! ✌🏼 🏖️ this time an 🥑 emoji...

### DIFF
--- a/lib/ui/views/home/home_viewmodel.dart
+++ b/lib/ui/views/home/home_viewmodel.dart
@@ -492,8 +492,8 @@ class HomeViewModel extends BaseViewModel {
   }
 
   Future<void> forceRefresh(BuildContext context) async {
-    _managerAPI.clearAllData();
+    await _managerAPI.clearAllData();
+    await initialize(context);
     _toast.showBottom(t.homeView.refreshSuccess);
-    initialize(context);
   }
 }
--- a/lib/ui/views/home/home_view.dart
+++ b/lib/ui/views/home/home_view.dart
@@ -22,7 +22,7 @@ class HomeView extends StatelessWidget {
         body: RefreshIndicator(
           edgeOffset: 110.0,
           displacement: 10.0,
-          onRefresh: () => model.forceRefresh(context),
+          onRefresh: () async => await model.forceRefresh(context),
           child: CustomScrollView(
             slivers: <Widget>[
               CustomSliverAppBar(
--- a/lib/ui/views/home/home_view.dart
+++ b/lib/ui/views/home/home_view.dart
@@ -20,6 +20,8 @@ class HomeView extends StatelessWidget {
       viewModelBuilder: () => locator<HomeViewModel>(),
       builder: (context, model, child) => Scaffold(
         body: RefreshIndicator(
+          edgeOffset: 110.0,
+          displacement: 10.0,
           onRefresh: () => model.forceRefresh(context),
           child: CustomScrollView(
             slivers: <Widget>[


Domenic-MZS:
Opened as draft as I don't know if using "magic numbers" (actually a static number - related to another static magic number) is right or not...

ILoveOpenSourceApplications:
Can the refresh icon animation be also improved or is it because the manager refreshes too quickly that the animation is getting canceled out?

Domenic-MZS:
> Can the refresh icon animation be also improved or is it because the manager refreshes too quickly that the animation is getting canceled out?

The animation is not cancelled; it's just fast because the completion of the promise for refresh happens almost instantly. This occurs because the code doesn't wait for the promises.

oSumAtrIX:
Is this ready for review?

Domenic-MZS:
The issue is fixed, but there's a related issue that could be fixed in this batch (the refresh animation goes away too fast because the callback does not wait to complete).

> Is this ready for review?

lib/ui/views/home/home_viewmodel.dart
```diff
@@ -492,8 +492,8 @@ class HomeViewModel extends BaseViewModel {
   }
 
   Future<void> forceRefresh(BuildContext context) async {
-    _managerAPI.clearAllData();
+    await _managerAPI.clearAllData();
+    await initialize(context);
```
oSumAtrIX, line 496: Whats the reason this line is changed?

Domenic-MZS: That's a good question, and the answer is quite simple. 

As @ILoveOpenSourceApplications pointed out, the refresh animation appears to cancel because it automatically resolves after being started. 

The `RefreshIndicator` plays the refreshing animation until the `onRefresh Future<void>` resolves. Since the method we use in the code, `forceRefresh`, doesn't properly await its asynchronous callbacks, the function immediately stops/resolves milliseconds after being called. It is for this reason that I changed the function to:

1. Wait for the inner promises to resolve 
2. Changed the order to prevent the "refresh done" before actually doing the work 
3. Complete the promise, stopping the animation properly with patience (waiting) instead of rushing (no-waiting)

> Flutter Docs 
> [onRefresh](https://api.flutter.dev/flutter/material/RefreshIndicator/onRefresh.html) → [RefreshCallback](https://api.flutter.dev/flutter/material/RefreshCallback.html)
>
> A function that's called when the user has dragged the refresh indicator far enough to demonstrate that they want the app to refresh. The returned [Future](https://api.flutter.dev/flutter/dart-async/Future-class.html) must complete when the refresh operation is finished.

Aunali321:
@Domenic-MZS Can you try to remove magic numbers if possible?

Domenic-MZS:
> @Domenic-MZS Can you try to remove the magic numbers if possible?

While it's technically feasible, the trade-offs involved wouldn't add significant value. Incorporating this functionality while maintaining the current user interface would unnecessarily increase complexity for this particular case.

I can think one of three possible solutions:
1. Make the `extendedHeight` of the `CustomSliverAppBar` parametrizable to match the offset.
2. Encapsulate the entire component tree within a `LayoutBuilder` to obtain metrics and sizes after rendering, which would allow us to compute and adjust the offset during a second redraw.
3. Utilize a `GlobalKey` to retrieve the expanded height of the `SliverAppBar` after it renders and then update the offset on a subsequent redraw.

However, none of these solutions are particularly appealing, as the requirement is specific to only one file, and the drawbacks outweigh the benefits.

oSumAtrIX (approved):

Issue #1817, opened 2024-04-01
bug: Memory overflow error while fixing the TikTok app


wangyi1573:
### Bug description

Memory overflow error while fixing the TikTok app

### Version of ReVanced Manager and version & name of app you are patching

com.zhiliaoapp.musically apk version 32.5.3

### Installation method

None

### ReVanced Manager logs

```shell
- Device Info
ReVanced Manager: 1.19.4-dev.2
Model: 23078RKD5C
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: Yes

- Patch Info
App: com.zhiliaoapp.musically v32.5.3 (Suggested: 32.5.3)
Patches version: v4.5.0
Patches added: SIM spoof
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: true
Version compatibility check: false
Show universal patches: false
Patches source: revanced/revanced-patches
Integration source: revanced/revanced-integrations

- Logs
Reading APK
Decoding app manifest
Loading patches
Merging integrations
Executing patches
Applied 9 patches
Compiling patched dex files
Compiled 22 dex files
An error occurred:
java.lang.OutOfMemoryError: Failed to allocate a 16 byte allocation with 4692160 free bytes and 4582KB until OOM, target footprint 536870912, growth limit 536870912; giving up on allocation because <1% of heap free after GC.
	at com.android.tools.smali.dexlib2.dexbacked.util.DebugInfo$DebugInfoImpl$3.readNextItem(SourceFile:1)
	at com.android.tools.smali.dexlib2.dexbacked.util.DebugInfo$DebugInfoImpl$3.readNextItem(SourceFile:2)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeLookaheadIterator.computeNext(Unknown Source:2)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeLookaheadIterator.tryToComputeNext(Unknown Source:3)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeLookaheadIterator.hasNext(Unknown Source:8)
	at com.android.tools.smali.dexlib2.writer.DexWriter.writeDebugItem(Unknown Source:136)
	at com.android.tools.smali.dexlib2.writer.DexWriter.writeDebugAndCodeItems(Unknown Source:198)
	at com.android.tools.smali.dexlib2.writer.DexWriter.writeTo(SourceFile:2)
[...]

Issue #1802, opened 2024-03-28
bug: java.lang.OutOfMemoryError while patching TikTok


Ondrejtra:
### Bug description

Attempting to patch TikTok from an apk file fails with java.lang.OutOfMemoryError. This issue occurs with default patches as well as a custom patch selection. Patches in logs below are Downloads, Feed filter, Playback speed, Remember clear display, Settings and Show seekbar. Patching the recommended version 32.5.3.

### Version of ReVanced Manager and version & name of app you are patching

ReVanced Manager: 1.19.3
App: TikTok (com.zhiliaoapp.musically v32.5.3)

### Installation method

Regular

### ReVanced Manager logs

```shell
--------- beginning of kernel
03-28 21:57:44.874 12253 12253 I binder  : 12253:12253 ioctl 40046210 7fed472a80 returned -22
--------- beginning of system
03-28 21:57:45.051 12253 12253 D NtViewRootImpl: mPopUpViewOffsets: offset=(0.0, 0.0), scale=(1.0, 1.0)
03-28 21:57:45.100 12292 12292 W audit   : audit_lost=50 audit_rate_limit=5 audit_backlog_limit=64
03-28 21:57:45.100 12292 12292 E audit   : rate limit exceeded
03-28 21:57:45.735 12280 12280 I binder  : 12253:12280 ioctl 40046210 7c1e985dd4 returned -22
03-28 21:57:49.682 12279 12279 W         : process 12830 (DartWorker) no longer affine to cpu7
03-28 21:57:50.412 12279 12279 W         : process 12829 (DartWorker) no longer affine to cpu7
03-28 21:57:50.633 12279 12279 W         : process 12828 (DartWorker) no longer affine to cpu7
03-28 21:57:50.652 12279 12279 W         : process 12557 (DartWorker) no longer affine to cpu7
03-28 21:57:56.374 12253 12253 W PackageParser: Unknown element under <manifest>: queries at /data/user/0/app.revanced.manager.flutter/cache/com.zhiliaoapp.musically_32.5.3-2023205030_minAPI21(arm64-v8a,armeabi-v7a)(nodpi)_apkmirror.com.apk Binary XML file line #5
[...]

Issue #1759, opened 2024-03-09
bug: unable to patch tik tok


LaganYT:
### Bug description

It won't let me patch the app.

### Version of ReVanced Manager and version & name of app you are patching

Tik Tok, 32.5.3

### Installation method

None

### ReVanced Manager logs

```shell
- Device Info
ReVanced Manager: 1.19.3
Model: 2201117TG
Android version: 13
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.ss.android.ugc.trill v32.5.3 (Suggested: 32.5.3)
Patches version: v4.3.0
Patches added: Default
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: false
Version compatibility check: true
Show universal patches: true
Patches source: revanced/revanced-patches
Integration source: revanced/revanced-integrations

- Logs
Reading APK
Decoding app manifest
Loading patches
Merging integrations
An error occurred:
java.lang.OutOfMemoryError: Failed to allocate a 24 byte allocation with 1213648 free bytes and 1185KB until OOM, target footprint 268435456, growth limit 268435456; giving up on allocation because <1% of heap free after GC.
	at com.android.tools.smali.dexlib2.dexbacked.instruction.DexBackedInstruction.buildInstruction(Unknown Source:237)
	at com.android.tools.smali.dexlib2.dexbacked.instruction.DexBackedInstruction.readFrom(Unknown Source:43)
	at com.android.tools.smali.dexlib2.dexbacked.DexBackedMethodImplementation$1$1.readNextItem(SourceFile:1)
	at com.android.tools.smali.dexlib2.dexbacked.DexBackedMethodImplementation$1$1.readNextItem(SourceFile:2)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeLookaheadIterator.computeNext(Unknown Source:2)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeLookaheadIterator.tryToComputeNext(Unknown Source:3)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeLookaheadIterator.hasNext(Unknown Source:8)
	at app.revanced.patcher.fingerprint.LookupMap$Maps.initializeLookupMaps$revanced_patcher(Unknown Source:172)
	at app.revanced.patcher.Patcher$apply$1.invokeSuspend(Unknown Source:187)
[...]

Issue #1684, opened 2024-02-08
bug: Compiling DEX (java out of memory)


yoimgllala:
### Bug description

In title

### Version of ReVanced Manager and version & name of application you tried to patch

- Patch Info
App: com.zhiliaoapp.musically v32.5.3 (Suggested: v32.5.3)
Patches version: v4.2.0
Patches added: Default
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: false
Version compatibility check: false
Show universal patches: false
Patches source: revanced/revanced-patches
Integration source: revanced/revanced-integrations

### Installation type

None

### Device logs

```shell
- Device Info
ReVanced Manager: 1.18.0
Model: 23078PND5G
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No
```

### Patcher logs

```shell
- Logs
Copying APK
Reading APK
Decoding app manifest
Loading patches
Merging integrations
Executing patches
Applied 8 patches
Compiling patched dex files
Compiled 22 dex files
An error occurred:
java.lang.OutOfMemoryError: Failed to allocate a 24 byte allocation with 623552 free bytes and 608KB until OOM, target footprint 536870912, growth limit 536870912; giving up on allocation because <1% of heap free after GC.
	at com.android.tools.smali.dexlib2.dexbacked.DexBackedAnnotation$1.readNextItem(DexBackedAnnotation.java:69)
	at com.android.tools.smali.dexlib2.dexbacked.DexBackedAnnotation$1.readNextItem(DexBackedAnnotation.java:65)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeSet$1.readNextItem(VariableSizeSet.java:58)
	at com.android.tools.smali.dexlib2.dexbacked.util.VariableSizeIterator.next(VariableSizeIterator.java:79)
	at java.util.AbstractSet.hashCode(AbstractSet.java:122)
	at com.android.tools.smali.dexlib2.base.BaseAnnotation.hashCode(BaseAnnotation.java:44)
	at java.util.HashMap.hash(HashMap.java:336)
	at java.util.HashMap.getNode(HashMap.java:566)
	at java.util.LinkedHashMap.get(LinkedHashMap.java:464)
	at com.android.tools.smali.dexlib2.writer.pool.BaseOffsetPool.getItemOffset(BaseOffsetPool.java:52)
[...]

PR #1869, opened 2024-04-16
Default to non-multithreaded until UI toggle support

taylorkline:
Patching large applications (e.g. TikTok) on Android with multithreading simply doesn't work on multicore devices where the amount of memory given to the manager (regardless of the system RAM) is not sufficient.

It's more important that patching succeeds for all available use cases than for only some patching to succeed with faster performance. A UI option can be added later, but the multithreading option should be opt-out by default rather than opt-in.

Closes ReVanced/revanced-documentation#35
Closes ReVanced/revanced-manager#1454
Closes ReVanced/revanced-manager#1571
Closes ReVanced/revanced-manager#1595
Closes ReVanced/revanced-manager#1659
Closes ReVanced/revanced-manager#1661
Closes ReVanced/revanced-manager#1684
Closes ReVanced/revanced-manager#1759
Closes ReVanced/revanced-manager#1802
Closes ReVanced/revanced-manager#1817
Closes ReVanced/revanced-patches#2885
Closes ReVanced/revanced-manager#592
Closes ReVanced/revanced-patcher#193
Closes ReVanced/revanced-patches#1533
Closes ReVanced/revanced-patches#1608
Closes ReVanced/revanced-patches#1613
Closes ReVanced/revanced-patches#1630
Closes ReVanced/revanced-patches#190
Closes ReVanced/revanced-patches#2511
Closes ReVanced/revanced-patches#525


android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
```diff
@@ -279,7 +279,7 @@ class MainActivity : FlutterActivity() {
                         tmpDir,
                         Aapt.binary(applicationContext).absolutePath,
                         tmpDir.path,
-                        true // TODO: Add option to disable this
+                        false // TODO: Add option to toggle multithreading in UI
```
taylorkline, line 282: I'm happy to create an Issue on GitHub for this to keep track of this, if at all desired.

taylorkline:
@oSumAtrIX This should save you some headaches. I see you replying to all of the OOM issues opened on the assorted ReVanced GitHubs.

Tagging some recent committers for potential review: @Aunali321 @Ushie @TheAabedKhan @lamnhan066 @Domenic-MZS @Axelen123 @validcube 

Thank you!

oSumAtrIX:
I think it's better to add a toggle if at all. Slowing down patching by the amount of cores worsens the experience for the majority of people whereas only fixes a problem for a handful of people. Single threading is still not a guarantee patching is gonna work though. The issue originates in how dex are compiled and that they consume lots of memory. From what I know this issue is a regression caused by a past migration from the smali library to Google's fork, whereas upstream had a fix involved for this particular problem.

taylorkline:
I'm of the opinion that maximum compatibility is preferable to speed. 

> ...worsens the experience for the majority of people...

On my OnePlus 9 Pro, I'm timing about 60 seconds for a patch of YouTube, which is already long enough where I set my phone down and go on to do something else while I'm waiting.

Without multithreading, I'm getting 120 seconds. When I've already set my phone aside because it was already taking a minute, I don't know that it really makes a difference for it to take another minute.

> ...only fixes a problem for a handful of people.

Not sure I agree with that given the number of reports, both on here and on Reddit.

> Single threading is still not a guarantee patching is gonna work though.

While true, the majority of reports on this issue are about TikTok, and TikTok patches fine without multithreading.

Ushie:
The complainers are loud, ReVanced Manager is used by hundreds of thousands of people, during your tests this change brings a %100 decrease in patching speed which could be even worse for different devices, I believe the UI toggle is not much of a complicated feat, making it a bit unpreferable to merge this

Aunali321:
I believe the better approach is how @Axelen123 has done it in compose manager (feat/compose-dev). It allows manager to execute patcher in another process and allows to change maximum alloted memory. It doesn't slow down patching rather speeds it up. I couldn't patch tiktok on my S21FE so i increased memory limit to 900mb (default is 512) and now i can patch it. It's better because it does not slow down patching process and because it allows the user the specify the max memory limit (in future maybe it can automated to use all available memory.)
SmartSelect_20240417_165004_ReVanced Manager

vmoffset:
> I believe the better approach is how @Axelen123 has done it in compose manager (feat/compose-dev). It allows manager to execute patcher in another process and allows to change maximum alloted memory. It doesn't slow down patching rather speeds it up. I couldn't patch tiktok on my S21FE so i increased memory limit to 900mb (default is 512) and now i can patch it. It's better because it does not slow down patching process and because it allows the user the specify the max memory limit (in future maybe it can automated to use all available memory.) SmartSelect_20240417_165004_ReVanced Manager

Can somebody implement this feature into main branch and release it, I genuinely hate going on pc and using revanced cli for bigger apps because of this error. don't see why it should stay this way if theres a fix for it @Axelen123

Axelen123:
> Can somebody implement this feature into main branch and release it, I genuinely hate going on pc and using revanced cli for bigger apps because of this error. don't see why it should stay this way if theres a fix for it @Axelen123 , this error has been a thing for so long too...

The implementation I wrote for compose manager will work in flutter manager with some changes. You can compile the compose-dev branch from source or try using the CLI in termux if you don't want to wait for a port or compose manager's release. I am not going to port the feature myself because I am not familiar with dart/flutter.

Ushie:
Closing this, there is no ETA towards the release of ReVanced Manager v2 but if the implementation is simple enough then it may be backported soon

Domenic-MZS:
That should be fairly easy. Can I take care of it?

oSumAtrIX:
@Domenic-MZS 

Sounds good. We could try to fast-track it into the main if it is merged.
Before copy-pasting the code, talk with @Axelen123 regarding the implementation, as he labeled it experimental. Make sure the code quality and style match the Flutter code base.

Domenic-MZS:
Yeah, I'll write to him later @oSumAtrIX. Also, I have one question. Do we really need to backport the compose-manager functionality? We could just increase the spawned thread with the given custom stack size without many changes, just by changing Thread { } to Thread(null, 'anyname', fun, stackSize).

oSumAtrIX:
I am not sure what you mean. The issue is that the Dalvik VM heap is being limited by Android.

Domenic-MZS:
Are you sure that the issue is on the heap? I did some tests on my phone with the above changes (increased stack size), and it worked flawlessly (without the changes it failed with the no enough memory issue).

oSumAtrIX:
Android reports an issue with heap not stack,  but perhaps increasing stack size increases heap size.

Domenic-MZS:
That makes more sense, but technically both are unrelated.

oSumAtrIX:
Feel free to propose the changes in a PR

Axelen123:
> Yeah, I'll write to him later @oSumAtrIX. Also, I have one question. Do we really need to backport the compose-manager functionality? We could just increase the spawned thread with the given custom stack size without many changes, just by replacing the `Thread { <patch-work-fun> }` to `Thread(null, 'anyname', <patch-work-fun>, stackSize)`.

Are you sure the stack size increase helps? What apps and sizes did you test with?

Domenic-MZS:
> > ...
>
> Are you sure the stack size increase helps? What apps and sizes did you test with?

I would say no. I tried increasing the stack size with `TikTok v32.5.3`, as mentioned in issue ReVanced/revanced-manager#1571, but it did not compile successfully. However, that was back in early April. I used the same device, but there have been many system updates since then, so I wouldn't put much trust in increasing the stack size alone. Nevertheless, it compiled successfully now.

I think this could be a 'false positive' at this point.

Domenic-MZS:
Anyway @Axelen123, could you further explain the necessary adjustments or parts needed for porting the `compose-dev` so i can start looking into for ?

Ushie:
We currently plan to release ReVanced Manager Compose in the following week or two, maybe we could use your contributions there instead, what do you think?

Domenic-MZS:
> We currently plan to release ReVanced Manager Compose in the following week or two, maybe we could use your contributions there instead, what do you think?

Yeah, I definitely would take a look over there to help when I can.

Axelen123:
> Anyway @Axelen123, could you further explain the necessary adjustments or parts needed for porting the `compose-dev` so i can start looking into for ?

Like Ushie said, there is a major push going on to release Compose Manager. If you still want to do it, start by looking here: https://github.com/ReVanced/revanced-manager/tree/compose-dev/app/src/main/java/app/revanced/manager/patcher/runtime. The folder can be copied to the Kotlin part of Flutter Manager, but it will need some adjustments:
- The `Session` class needs to be ported or replaced.
- Some of the progress tracking code needs to be replaced with the corresponding methods used by Flutter Manager.
- Code that uses classes injected using koin's `inject()` has to be changed since it does not exist in Flutter Manager.
- You will have to setup NDK compilation to compile the shared library used to spoof properties

Issue #1776, opened 2024-03-21
bug: GitHub ratelimit unhandled


ILoveOpenSourceApplications:
### Bug description

If alternative source is toggled on/off once, the manager goes into this unresponsive state where none of the patchable apps load, and if a patchable app is selected, none of the patches load either. And the changeling also stops working as it showing a loading circle instead.

https://github.com/ReVanced/revanced-manager/assets/117499019/75d2a702-58b3-4fb8-919c-8c619b838cdf

![Screenshot_20240322_025615_ReVanced Manager](https://github.com/ReVanced/revanced-manager/assets/117499019/e1878afa-f851-4969-b550-b25c44051d15)

PS: #1760 can be clubbed with this issue.

### Version of ReVanced Manager and version & name of app you are patching

v1.19.3

### Installation method

None

### ReVanced Manager logs

```shell
Nil
```

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The bug is only related to ReVanced Manager

oSumAtrIX:
Are you rate limited on GitHub? https://api.github.com/rate_limit

ILoveOpenSourceApplications:
```
{"resources":{"core":{"limit":60,"remaining":2,"reset":1711061079,"used":58,"resource":"core"},"graphql":{"limit":0,"remaining":0,"reset":1711061595,"used":0,"resource":"graphql"},"integration_manifest":{"limit":5000,"remaining":5000,"reset":1711061595,"used":0,"resource":"integration_manifest"},"search":{"limit":10,"remaining":10,"reset":1711058055,"used":0,"resource":"search"}},"rate":{"limit":60,"remaining":2,"reset":1711061079,"used":58,"resource":"core"}}
```

oSumAtrIX:
You seem rate-limited, but still, this is something ReVanced Manager should take care of.

Maccer1:
I have the same issue I believe. I don't have many technical details but I am unable to patch any app and everything looks pink like in the video above.

I am using the default config and API. Tried reinstalling the last 2 versions with no luck.

oSumAtrIX:
For support move to the links on https://revanced.app

validcube:
👋 Hello, @ILoveOpenSourceApplications!

[...]

PR #1911, opened 2024-05-18
fix: Cache external API calls 

kitadai31:
https://github.com/ReVanced/revanced-manager/issues/1776#issuecomment-2115827977

Make github_api to use Lock (synchronized) like revanced_api
This reduces GitHub API rate limit exceedance.


BenjaminHalko (approved):
Code LGTM!

Issue #1783, opened 2024-03-24
bug: App Not Drawing Behind Navigation Bar on Patching Screen


Dan-Banfield:
### Bug description

On the patching screen, it seems the app is not in edge-to-edge mode because it does not draw behind the system navigation bar.

![1000020943](https://github.com/ReVanced/revanced-manager/assets/73112854/8c41a347-1b1b-413a-bc89-82e75b0e9462)

### Version of ReVanced Manager and version & name of app you are patching

Version 1.19.3
App being patched: any

### Installation method

Regular

### ReVanced Manager logs

```shell
Not relevant.
```

### Patch logs

```shell
Not relevant.
```

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The bug is only related to ReVanced Manager

snehithssv:
So, while on patching screen, the bottom nav-bar should be hidden?? @Dan-Banfield

Dan-Banfield:
> So, while on patching screen, the bottom nav-bar should be hidden?? @Dan-Banfield

No, not hidden, just the app should draw behind it in-line for the upcoming mandate in Android 15.

snehithssv:
So, app is to be forced to be in edgeToEdge instead of leanBack/immersive.?
More like this...??
![Screenshot_1718089769](https://github.com/ReVanced/revanced-manager/assets/112557710/c173bf34-1a32-4999-95d9-17bc96a60539)

 And if yes, should it be this way only in patch screen on for the whole application??

Dan-Banfield:
In the screenshot you provided, the navigation bar at the bottom is still black. Instead, it should be drawn behind. Please see: https://developer.android.com/about/versions/15/behavior-changes-15#:~:text=Edge-to-edge%20enforcement,because%20most%20components%20apply%20insets.

snehithssv:
Thanks a lot for helping me out... @Dan-Banfield 
[projc.webm](https://github.com/ReVanced/revanced-manager/assets/112557710/86d118d1-4fbe-40db-8920-1df1e7399665)
So, is this the desired functionality?

Dan-Banfield:
> Thanks a lot for helping me out... @Dan-Banfield 
> [projc.webm](https://github.com/ReVanced/revanced-manager/assets/112557710/86d118d1-4fbe-40db-8920-1df1e7399665)
[...]

Issue #1475, opened 2023-11-09
Progress output gets obstructed by the navigation bar


lifehackerhansol:
### Bug description

During the patching progress, after the text has flowed to the bottom of the screen, the text is partially covered by the navigation bar. Scrolling down does nothing.

Steps to reproduce:
1. Install as many patches as possible to fill up the output area
2. Watch the text flow down

![Screenshot_20231109-042106](https://github.com/ReVanced/revanced-manager/assets/37358975/18e95de2-99ee-4c88-8f54-f69eefc276d1)

### Version of ReVanced Manager and version & name of application you tried to patch

Version: 1.15.1
Model: Pixel 6a
Android Version: 14
Supported Archs: arm64-v8a, armeabi-v7a, armeabi

YouTube 18.38.44, but this is a UI bug and the app shouldn't matter.

### Installation type

Non-root

### Device logs

```shell
--------- beginning of crash
11-09 03:21:22.797 22874 22874 E AndroidRuntime: FATAL EXCEPTION: main
11-09 03:21:22.797 22874 22874 E AndroidRuntime: Process: app.revanced.manager.flutter, PID: 22874
11-09 03:21:22.797 22874 22874 E AndroidRuntime: java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=1024, result=-1, data=Intent { (has extras) }} to activity {app.revanced.manager.flutter/app.revanced.manager.flutter.MainActivity}: java.lang.IllegalStateException: Reply already submitted
11-09 03:21:22.797 22874 22874 E AndroidRuntime: 	at android.app.ActivityThread.deliverResults(ActivityThread.java:5527)
11-09 03:21:22.797 22874 22874 E AndroidRuntime: 	at android.app.ActivityThread.handleSendResult(ActivityThread.java:5566)
11-09 03:21:22.797 22874 22874 E AndroidRuntime: 	at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:67)
11-09 03:21:22.797 22874 22874 E AndroidRuntime: 	at android.app.servertransaction.ActivityTransactionItem.execute(ActivityTransactionItem.java:45)
11-09 03:21:22.797 22874 22874 E AndroidRuntime: 	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:139)
[...]

PR #1945, opened 2024-06-12
fix: Patching Screen draw-behind Navigation Bar

snehithssv:
fixes: #1783


TheAabedKhan:
Not everyone uses the same navigation mode. Refer to #1475

snehithssv:
Thanks @TheAabedKhan for the insight. Fixed for other nav. mode too

PR #1947, opened 2024-06-13
feat: Add ability to set `null` in patch options

TheAabedKhan:
Some screenshots
null-patch-option


validcube (approved):
🥞 Approved, thanks! 

Compiled using Flutter 3.23.0-0.1.pre (Beta) and tested on physical devices running Android 14 (Samsung One UI 6.1 - 1st May)

## Test result (🥞 Passed)
* Application: YouTube v19.16.39 - Custom Branding
  * Null value provided to "App name" option: ✅ (Should stay "YouTube" since we don't change anything)
  * "YT ReVanced" provided to "App name" option: ✅ (The application should be "YT ReVanced" as provided)

PR #1951, opened 2024-06-17
feat: allow bundles to use classes from other bundles

Axelen123:


Axelen123:
Note to self: This needs more testing, since I have basically overhauled the entire state management part of the bundles repository and have not sufficiently tested it yet.

PR #2011, opened 2024-06-30
fix: unsupported patch toast says "patchItem.unsupportedPatchVersion"

kitadai31:
When I tried selecting unsupported patches, wrong text is shown.

Screenshot_20240630-134240_ReVanced_Manager_Debug

This is a forgotten migration from flutter_i18n.
I confirmed that there were no other missed migrations.


validcube (approved):

validcube:
🥞 Approved, thanks!

PR #2016, opened 2024-06-30
chore: Remove NDK constraint

BenjaminHalko:
A plugin says it requires NDK version `26.1.10909125`, so the NDK version has been changed from `flutter.ndkVersion` to `26.1.10909125`

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -24,7 +24,6 @@ if (flutterVersionName == null) {
 
 android {
     compileSdk 34
-    ndkVersion "26.1.10909125"
 
     compileOptions {
         coreLibraryDesugaringEnabled true
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -24,7 +24,7 @@ if (flutterVersionName == null) {
 
 android {
     compileSdk 34
-    ndkVersion flutter.ndkVersion
+    ndkVersion "26.1.10909125"
 
     compileOptions {
         coreLibraryDesugaringEnabled true
--- a/lib/ui/widgets/patchesSelectorView/patch_options_fields.dart
+++ b/lib/ui/widgets/patchesSelectorView/patch_options_fields.dart
@@ -6,6 +6,7 @@ import 'package:revanced_manager/app/app.locator.dart';
 import 'package:revanced_manager/gen/strings.g.dart';
 import 'package:revanced_manager/models/patch.dart';
 import 'package:revanced_manager/services/manager_api.dart';
+import 'package:revanced_manager/ui/views/patch_options/patch_options_viewmodel.dart';
 import 'package:revanced_manager/ui/widgets/shared/custom_card.dart';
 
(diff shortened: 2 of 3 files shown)


oSumAtrIX:
I don't know if this is right. From what I saw building still worked

BenjaminHalko:
Yep, building still works, but you get this error:
image

BenjaminHalko:
What does prevent building is the fact that `Patch options fields` was missing the `patch_options_viewmodel` import, which prevented building. So that was fix here as well. (Maybe could have made a separate PR or pushed that directly to dev though)

BenjaminHalko:
If this PR gets approved quickly, I'll leave that extra fix in there, otherwise I should probably push the `patch_options_viewmodel` fix to dev so that Manager can build properly.

oSumAtrIX:
> Yep, building still works, but you get this error:

Which plugin causes this. This wasn't an issue before.

BenjaminHalko:
It might have happened after upgrading `intl` to `0.19.0` which is necessary to use `flutter_localizations`, but it is hard to say for certain. It has been happening for at least a month judging from the build logs, but any further and the logs have been deleted.

oSumAtrIX:
A git bisect should be helpful finding the commit, or it may be that plugin, even though it's actually a dependency. I don't know if it's correct practice to simply just pin the ndk version. It is set to flutter.ndkVersion by default for a reason

BenjaminHalko:
The other option is to simply get rid of `ndkVersion` all together and it will instead use the NDK version provided by the `AGP`. This gets rid of the error.

oSumAtrIX:
What's the reason for it being set to flutter.ndkVersion

BenjaminHalko:
OK, so `flutter.ndkVersion` is an NDK version that is supposedly the "most compatible" with the flutter framework. I think currently it is `23.something`

BenjaminHalko:
But you are fully allowed to change the NDK version if there is a version mismatch.

oSumAtrIX:
Why would the intl dependency rely on an NDK not most compatible with Flutter. I think downgrading the library makes more sense

BenjaminHalko:
Unfortunately, `flutter_localizations` depends on intl `0.19.0` and `flutter_localizations` is from the flutter SDK. That said other plugins that have been recently updated/added are `skeletons` and `file_picker`. So it might be one of those...

BenjaminHalko:
It could have been a dependency of a package that got upgraded...

android/app/build.gradle
```diff
@@ -24,7 +24,7 @@ if (flutterVersionName == null) {
 
 android {
     compileSdk 34
-    ndkVersion flutter.ndkVersion
+    ndkVersion "26.1.10909125"
```
validcube, line 27: This should be safe to remove, see

https://github.com/flutter/flutter/issues/139427#issuecomment-1989024985

validcube (approved):
🥞 Approved!

PR #2059, opened 2024-07-14
build: Enable Gradle Configuration Cache

validcube:
Speed up build:tm:

Compilation Trials were done using Android Studio Koala Feature Drop | 2024.1.2 Canary 8

VM: JetBrains OpenJDK 64-Bit 17.0.11 amd64
Windows 11.0


validcube:
Upon checking, there are seems to be no incompatibility that could take place in enabling configuration cache

PR #2074, opened 2024-07-22
feat: Support Android 15

validcube:
# ✨ Support and utilise features from SDK 35
This PR adds support for Vanilla Ice Cream and uses features from it like `<nobreak>` (https://developer.android.com/about/versions/15/features#line-breaks)

This PR hopes to support and utilise features from Android 15 (hopefully) before the final release of Android.


app/src/main/res/values/strings.xml
```diff
@@ -43,17 +43,17 @@
     <string name="legacy_import_failed">Could not import legacy settings</string>
 
     <string name="auto_updates_dialog_title">Configure updates</string>
-    <string name="auto_updates_dialog_description">Do you want ReVanced Manager to periodically check for updates for the following components?</string>
-    <string name="auto_updates_dialog_manager">ReVanced Manager</string>
-    <string name="auto_updates_dialog_patches">ReVanced Patches</string>
+    <string name="auto_updates_dialog_description">Do you want <nobreak>ReVanced Manager</nobreak> to periodically check for updates for the following components?</string>
+    <string name="auto_updates_dialog_manager"><nobreak>ReVanced Manager</nobreak></string>
+    <string name="auto_updates_dialog_patches"><nobreak>ReVanced Patches</nobreak></string>
```
Axelen123, line 48: Are you sure `nobreak` works with `stringResource`?

validcube: I.. forgot

Axelen123: It should work if the string is read as an `AnnotatedString`. Compose does not have this currently, but there are snippets online for it.

PR #2076, opened 2024-07-23
feat: Add reset button to custom API

### LINKED ISSUE #2051: feat: Add reset API to default button
### Feature description

If a custom API is set, theres no way to reset it back to default:

image

### Motivation

You could forget the original API url.

### Acknowledgements

- [X] This issue is not a duplicate of an existing feature request.
- [X] The feature request is only related to ReVanced Manager

Ushie:
Closes #2051

### DIFF
--- a/app/src/main/java/app/revanced/manager/domain/manager/base/BasePreferencesManager.kt
+++ b/app/src/main/java/app/revanced/manager/domain/manager/base/BasePreferencesManager.kt
@@ -56,7 +56,7 @@ class EditorContext(private val prefs: MutablePreferences) {
 
 abstract class Preference<T>(
     private val dataStore: DataStore<Preferences>,
-    protected val default: T
+    val default: T
 ) {
     internal abstract fun Preferences.read(): T
     internal abstract fun MutablePreferences.write(value: T)
@@ -64,7 +64,6 @@ abstract class Preference<T>(
     val flow = dataStore.data.map { with(it) { read() } ?: default }.distinctUntilChanged()
 
     suspend fun get() = flow.first()
-    fun defaultValue() = default
     fun getBlocking() = runBlocking { get() }
     @Composable
     fun getAsState() = flow.collectAsStateWithLifecycle(initialValue = remember {
--- a/app/src/main/java/app/revanced/manager/domain/manager/base/BasePreferencesManager.kt
+++ b/app/src/main/java/app/revanced/manager/domain/manager/base/BasePreferencesManager.kt
@@ -64,6 +64,7 @@ abstract class Preference<T>(
     val flow = dataStore.data.map { with(it) { read() } ?: default }.distinctUntilChanged()
 
     suspend fun get() = flow.first()
+    fun defaultValue() = default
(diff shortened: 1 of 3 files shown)


app/src/main/java/app/revanced/manager/domain/manager/base/BasePreferencesManager.kt
```diff
@@ -64,6 +64,7 @@ abstract class Preference<T>(
     val flow = dataStore.data.map { with(it) { read() } ?: default }.distinctUntilChanged()
 
     suspend fun get() = flow.first()
+    fun defaultValue() = default
```
Ushie, line 67: This was named defaultValue because of this:
```
app/src/main/java/app/revanced/manager/domain/manager/base/BasePreferencesManager.kt:59:15 Platform declaration clash: The following declarations have the same JVM signature (getDefault()Ljava/lang/Object;):
    fun <get-default>(): T defined in app.revanced.manager.domain.manager.base.Preference
    fun getDefault(): T defined in app.revanced.manager.domain.manager.base.Preference
```

Axelen123: `default` is declared as a `protected val` in the `Preference` constructor. You can make it public to get rid of the function

app/src/main/java/app/revanced/manager/ui/screen/settings/AdvancedSettingsScreen.kt
```diff
@@ -209,7 +203,12 @@ private fun APIUrlDialog(currentUrl: String, onSubmit: (String?) -> Unit) {
                 OutlinedTextField(
                     value = url,
                     onValueChange = { url = it },
-                    label = { Text(stringResource(R.string.api_url)) }
+                    label = { Text(stringResource(R.string.api_url)) },
+                    trailingIcon = {
+                        IconButton(onClick = { url = defaultUrl }) {
+                            Icon(Icons.Outlined.Restore, null)
```
Axelen123, line 209: A `contentDescription` is required here because there is no other text

Axelen123 (changes requested):

validcube (approved):
🪐 I would suggest making the default value or current value as the TextField's placeholder, anyway, still LGTM.

app/src/main/res/values/strings.xml
```diff
@@ -367,4 +367,5 @@
     <string name="add_patch_bundle">Add patch bundle</string>
     <string name="bundle_url">Bundle URL</string>
     <string name="auto_update">Auto update</string>
+    <string name="reset_api_url">Reset API URL</string>
```
Axelen123, line 370: Move this to the other api url dialog strings and rename the resource accordingly

Axelen123 (changes requested):

Ushie:
> 🪐 I would suggest making the default value or current value as the TextField's placeholder, anyway, still LGTM.

That's not possible, that would be the label not the placeholder and we use label for "API URL"

Axelen123 (approved):

PR #2080, opened 2024-07-24
fix: `ExtendedFloatingActionButton` not accessible by screen readers

TheAabedKhan:


app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -532,7 +538,10 @@ fun OptionsDialog(
                     if (values == null || !values.contains(key)) option.default else values[key]
 
                 @Suppress("UNCHECKED_CAST")
-                OptionItem(option = option as Option<Any>, value = value, setValue = { set(key, it) })
+                OptionItem(
+                    option = option as Option<Any>,
+                    value = value,
+                    setValue = { set(key, it) })
```
Axelen123, line 544: Add a newline

Axelen123 (changes requested):
Each `ExtendedFloatingActionButton` already has text. Won't this result in duplicates when the buttons are expanded?

optimatrix0:
Now the button is accessible, I really appreciate that.

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -532,7 +538,12 @@ fun OptionsDialog(
                     if (values == null || !values.contains(key)) option.default else values[key]
 
                 @Suppress("UNCHECKED_CAST")
-                OptionItem(option = option as Option<Any>, value = value, setValue = { set(key, it) })
+                OptionItem(
+                    option = option as Option<Any>,
+                    value = value,
+                    setValue = {
+                        set(key, it)
+                    })
```
Axelen123, line 546: Expand it so it looks like this:
```kt
                OptionItem(
                    option = option as Option<Any>,
                    value = value,
                    setValue = {
                        set(key, it)
                    }
		)
```

Axelen123 (approved):

PR #2084, opened 2024-07-25
fix: Support patching on ARMv7 by updating AAPT2 

oSumAtrIX:
Binary taken from run https://github.com/ReVanced/aapt2/actions/runs/9977958347

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -11,7 +11,7 @@
     <string name="select_app">Select an app</string>
     <string name="select_patches">Select patches</string>
 
-    <string name="unsupported_architecture_warning">Patching on ARMv7 devices is not yet supported and will most likely fail.</string>
+    <string name="unsupported_architecture_warning">Patching on this device architecture is unsupported and will most likely fail.</string>
 
     <string name="import_">Import</string>
     <string name="import_bundle">Import patch bundle</string>
--- a/app/src/main/java/app/revanced/manager/patcher/aapt/Aapt.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/aapt/Aapt.kt
@@ -4,7 +4,7 @@ import android.content.Context
 import app.revanced.manager.patcher.LibraryResolver
 import android.os.Build.SUPPORTED_ABIS as DEVICE_ABIS
 object Aapt : LibraryResolver() {
-    private val WORKING_ABIS = setOf("arm64-v8a", "x86", "x86_64")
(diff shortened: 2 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -11,7 +11,7 @@
     <string name="select_app">Select an app</string>
     <string name="select_patches">Select patches</string>
 
-    <string name="unsupported_architecture_warning">Patching on ARMv7 devices is not yet supported and will most likely fail.</string>
+    <string name="unsupported_architecture_warning">Patching on this device architecture is unsupported and will most likely fail.</string>
```
Ushie, line 14: If this PR solves patching on ARMv7 then what is this used for? Does it still check if the device arch isnt in the aapt `WORKING_ABIS`?

oSumAtrIX: Yes. ARMv7 is added to the array https://github.com/ReVanced/revanced-manager/pull/2084/files#diff-5becc4ed7b2026d3a471cc120219ed7ccfdfdd19a89fbd948f0ff9ef34f65e42R7

Ushie: That's not what I asked, I asked whether or not the ABI check on patch checks if the device ABI is in the array or not, or if it's hardcoded to check for ARMv7

Axelen123 (approved):

kitadai31:
Now that all architectures are supported, I think the entire architecture warning can be removed

https://github.com/ReVanced/revanced-manager/blob/52a89b16389ac39ff559ed52b1196915316cc6c8/app/src/main/java/app/revanced/manager/ui/screen/DashboardScreen.kt#L210-L219

https://github.com/ReVanced/revanced-manager/blob/52a89b16389ac39ff559ed52b1196915316cc6c8/app/src/main/java/app/revanced/manager/patcher/aapt/Aapt.kt#L7-L9

oSumAtrIX:
That's not all architectures

kitadai31:
If you are talking about `armeabi` `mips` and `mips64`, there is no need to consider them
If the device only supports those architectures, the installation will be rejected with a compatibility error because the package installer checks the apk archs

In addition, wasn't this warning necessary because armv7 aapt2 was unstable but still existed?
Previously, some devices were compiling resources successfully, and others were failing with 135 error
I believe this is the reason of the word "_will most likely fail._"
OTOH `armeabi` `mips` and `mips64` of aapt2 are not present in manager
This is the difference

kitadai31:
> I believe this is the reason of the word "will most likely fail."

nvm, I forgot about if users patched non-resource patches only

Ushie (approved):

Issue #2120, opened 2024-08-06
bug: tmp directory should not be inside the cache directory


kitadai31:
### Bug description

When the storage space is not enough, the Android system deletes apps cache automatically even if the app is running.
Currently, ReVanced Manager (for both Compose and Flutter) places the tmp directory inside the cache folder.
Therefore, if the cache deletion occurs during the patching process, the patching will fail.

The following files are better to be moved out of the cache folder to ensure they are kept until the end of patching:

- Patcher temporary dir (It's called "ephemeral" in Compose Manager)
- Input apk (Only when selected the apk from storage)
- Patches jar (Only on Flutter Manager. This has been resolved in Compose Manager.)

This automatic deletion is only occurs for the cache directory.
The cacheDir is unreliable for working directory due to deletion by Android system.
So, changing the tempDir place from `app.cacheDir.resolve("ephemeral")` to like `app.getDir("ephemeral", Context.MODE_PRIVATE)` solves this issue.

This puts the responsibility on the ReVanced Manager to ensure that the ephemeral folder is deleted (as users cannot manage the data directory), but this is not a problem since it is already deleted every time the app is launched and the patching is finished.

cf. https://discord.com/channels/952946952348270622/1135563848586379264/1269377385669857282 (Discussion of this issue)

I can make a PR for this.

### Version of ReVanced Manager and version & name of app you are patching

Flutter Manager v1.21.1
Compose Manager 32f2710 (compose-dev)

### Patch logs

**Examples of cache deletion issues:**

```
[Case 1]
The input apk was deleted immediately after start patching.

- Logs
Copying APK
Reading APK
Decoding app manifest
Loading patches
Merging integrations
An error occurred:
java.io.FileNotFoundException: /data/user/0/app.revanced.manager.flutter/cache/patcher/tmp-MOERDE/base.apk


[Case 2]
The resources folder was deleted while executing patches.

- Logs
Copying APK
Reading APK
Decoding app manifest
Loading patches
Merging integrations
Deleting existing resource cache directory
Decoding resources
Executing patches
Applied 55 patches
SponsorBlock failed: app.revanced.patcher.patch.PatchException: 'SponsorBlock' depends on 'SponsorBlockResourcePatch' that raised an exception:
app.revanced.patcher.patch.PatchException: /data/user/0/app.revanced.manager.flutter/cache/patcher/tmp-ILRKZL/cache/res/layout/revanced_sb_inline_sponsor_overlay.xml
[...]

PR #2122, opened 2024-08-07
fix: Move temporary files outside of the cache directory

kitadai31:
fix for #2120

The "ephemeral" directory is moved from cacheDir to filesDir.
The input.apk (exists if selected from storage) is moved to the ephemeral directory.

I'm not sure if the changes in `Session.kt` are right.
This is intended to delete input.apk as soon as possible when it is no longer needed.
If the apk is a downloaded apk or selected from installed apps, nothing happens on that line.

### DIFF
--- a/app/src/main/java/app/revanced/manager/data/platform/Filesystem.kt
+++ b/app/src/main/java/app/revanced/manager/data/platform/Filesystem.kt
@@ -2,6 +2,7 @@ package app.revanced.manager.data.platform
 
 import android.Manifest
 import android.app.Application
+import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Environment
@@ -16,7 +17,7 @@ class Filesystem(private val app: Application) {
      * A directory that gets cleared when the app restarts.
      * Do not store paths to this directory in a parcel.
      */
-    val tempDir = app.filesDir.resolve("ephemeral").apply {
+    val tempDir = app.getDir("ephemeral", Context.MODE_PRIVATE).apply {
         deleteRecursively()
         mkdirs()
     }
--- a/app/src/main/java/app/revanced/manager/data/platform/Filesystem.kt
+++ b/app/src/main/java/app/revanced/manager/data/platform/Filesystem.kt
@@ -1,10 +1,10 @@
 package app.revanced.manager.data.platform
 
+import android.Manifest
 import android.app.Application
+import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Environment
-import android.Manifest
-import android.content.pm.PackageManager
 import androidx.activity.result.contract.ActivityResultContract
 import androidx.activity.result.contract.ActivityResultContracts
 import app.revanced.manager.util.RequestManageStorageContract
@@ -16,7 +16,7 @@ class Filesystem(private val app: Application) {
      * A directory that gets cleared when the app restarts.
      * Do not store paths to this directory in a parcel.
      */
-    val tempDir = app.cacheDir.resolve("ephemeral").apply {
+    val tempDir = app.filesDir.resolve("ephemeral").apply {
         deleteRecursively()
         mkdirs()
     }
--- a/app/src/main/java/app/revanced/manager/patcher/Session.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/Session.kt
@@ -19,7 +19,7 @@ import java.nio.file.StandardCopyOption
 internal typealias PatchList = List<Patch<*>>
 
 class Session(
-    cacheDir: String,
+    private val cacheDir: String,
     frameworkDir: String,
     aaptPath: String,
     multithreadingDexFileWriter: Boolean,
@@ -129,6 +129,7 @@ class Session(
 
     override fun close() {
         tempDir.deleteRecursively()
+        File(cacheDir, "input.apk").delete()
         patcher.close()
     }
 
--- a/app/src/main/java/app/revanced/manager/patcher/Session.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/Session.kt
@@ -19,7 +19,7 @@ import java.nio.file.StandardCopyOption
 internal typealias PatchList = List<Patch<*>>
 
 class Session(
-    private val cacheDir: String,
+    cacheDir: String,
     frameworkDir: String,
     aaptPath: String,
     multithreadingDexFileWriter: Boolean,
@@ -129,7 +129,6 @@ class Session(
 
     override fun close() {
         tempDir.deleteRecursively()
-        File(cacheDir, "input.apk").delete()
         patcher.close()
     }
 
--- a/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
@@ -193,6 +193,9 @@ class PatcherWorker(
             Result.failure()
         } finally {
             patchedApk.delete()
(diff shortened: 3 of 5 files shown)


oSumAtrIX:
Does this PR also delete the temporal files after they are not necessary anymore?

kitadai31:
Yes
After patching is finished, only the output patched apk will be in the temporary folder
This apk will also be deleted when I back to the dashboard

oSumAtrIX:
Which line of code deletes the temp files?

kitadai31:
`File(cacheDir, "input.apk").delete()` removes a input apk which is selected from storage

Other temporary files are already deleted after patching since before this PR

app/src/main/java/app/revanced/manager/data/platform/Filesystem.kt
```diff
@@ -16,7 +16,7 @@ class Filesystem(private val app: Application) {
      * A directory that gets cleared when the app restarts.
      * Do not store paths to this directory in a parcel.
      */
-    val tempDir = app.cacheDir.resolve("ephemeral").apply {
+    val tempDir = app.filesDir.resolve("ephemeral").apply {
```
Axelen123, line 19: Use getDir instead

app/src/main/java/app/revanced/manager/patcher/Session.kt
```diff
@@ -129,6 +129,7 @@ class Session(
 
     override fun close() {
         tempDir.deleteRecursively()
+        File(cacheDir, "input.apk").delete()
```
Axelen123, line 132: PatcherViewModel already deletes it.

app/src/main/java/app/revanced/manager/ui/viewmodel/AppSelectorViewModel.kt
```diff
@@ -17,15 +18,19 @@ import kotlinx.coroutines.Dispatchers
...
 class AppSelectorViewModel(
     private val app: Application,
     private val pm: PM,
     private val patchBundleRepository: PatchBundleRepository
-) : ViewModel() {
-    private val inputFile = File(app.cacheDir, "input.apk").also {
+) : ViewModel(), KoinComponent {
+    private val fs: Filesystem by inject()
+
+    private val inputFile = File(fs.tempDir, "input.apk").also {
```
Axelen123, line 33: tempDir is only valid for the duration of the app process, but this needs to live longer than that. Use getDir instead.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/data/platform/Filesystem.kt
kitadai31: Since getDir creates the directory when it is called, I thought it was not suitable because I would have to create the directory myself.
I noticed that getDir and mkdirs() have different permissions on the folders created.
Should I still use getDir?

kitadai31:
Reflected the review, thank you.

Also, I moved the deleting selected apk code to PatcherWorker.

Currently, PatcherViewModel deletes the selected apk on `onCleared()` method.
But this may cause the selected apk (input.apk) to remain in internal storage because onCleared() won't be called if the users exit the task from recent apps.
If user exits the task after patching has finished without returning to the dashboard by accident, the input.apk remains in the app data.
Therefore, I think the input.apk should be deleted as soon as possible when patching is ended.
This is the reason I moved the code.

Is this OK?

app/src/main/java/app/revanced/manager/ui/viewmodel/AppSelectorViewModel.kt
```diff
@@ -25,7 +26,7 @@ class AppSelectorViewModel(
     private val pm: PM,
     private val patchBundleRepository: PatchBundleRepository
 ) : ViewModel() {
-    private val inputFile = File(app.cacheDir, "input.apk").also {
+    private val inputFile = File(app.getDir("ephemeral", Context.MODE_PRIVATE), "input.apk").also {
```
Axelen123, line 29: "ephemeral" is deleted by `Filesystem` every time the app process is restarted. Use another directory.

app/src/main/java/app/revanced/manager/data/platform/Filesystem.kt
Axelen123: Yes, because that is what the android docs recommends.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -193,6 +193,9 @@ class PatcherWorker(
             Result.failure()
         } finally {
             patchedApk.delete()
+            if (args.input is SelectedApp.Local && args.input.temporary) {
+                args.input.file.delete()
+            }
```
Axelen123, line 198: What is the reason for this change?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/AppSelectorViewModel.kt
kitadai31: input.apk should be deleted every time the app restarted, shouldn't it?
Is input.apk needed after restarted?

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
kitadai31: Please read the message above 🙇‍♂️

Axelen123: My bad, makes sense.

app/src/main/java/app/revanced/manager/ui/viewmodel/AppSelectorViewModel.kt
Axelen123: We cannot delete the input APK every time the app process is restarted because of [system-initiated process death](https://galex.dev/posts/process-death-is-the-rule-not-the-exception/#users-experience-system-initiated-process-death-a-lot). The process is killed when the app is in the background, but when the user navigates back a new process is launched and state is restored. System-initiated process death would definitely cause problems if the app is killed while the user is in a screen between `AppSelectorScreen` and `PatcherScreen`.

kitadai31: OK understood.
I now get what you mean by "process".

Since this is a single file, I think the destination will be `filesDir`. Is that correct?
( `File(app.filesDir, "input.apk")` )

Axelen123: Use something like `getDir("app_selector", Context.MODE_PRIVATE)`. I remember having problems when trying to use `filesDir` directly.

kitadai31: Isn't it [`dataDir`](https://developer.android.com/reference/android/content/Context#getDataDir()) ?

> Apps should not use this path directly; they should instead use getFilesDir(), getCacheDir(), getDir(java.lang.String, int), or other storage APIs on this class. 

`filesDir` is a standard location to place a file
https://developer.android.com/training/data-storage/app-specific#internal-access-store-files
I replaced it with `File(app.filesDir, "input.apk")` and there were no problems.

Axelen123: In that case `filesDir` is fine.

Axelen123 (approved):

Issue #2125, opened 2024-08-08
bug: Patched file deleted after patching if "Save last APK" in settings is toggled off


oSumAtrIX:
### Bug description

When the settings to save the last patched APK are off, the patched APK on the install screen can not be installed. The file seems to be missing.

### Version of ReVanced Manager and version & name of app you are patching

1.21.1

### Installation method

Regular

### Patch logs

```shell
An error occurred:
java.io.FileNotFoundException: /data/user/0/app.revanced.manager.flutter/cache/patcher/tmp-IHTONZ/in.apk: open failed: EACCES (Permission denied)
    at libcore.io.IoBridge.open(IoBridge.java:574)
    at java.io.RandomAccessFile.<init>(RandomAccessFile.java:289)
    at com.android.tools.build.apkzlib.zip.ZFile.reopenRw(Unknown Source:105)
    at com.android.tools.build.apkzlib.zip.ZFile.writeAllFilesToZip(Unknown Source:2)
    at com.android.tools.build.apkzlib.zip.ZFile.update(Unknown Source:21)
    at com.android.tools.build.apkzlib.zip.ZFile.close(Unknown Source:9)
    at kotlin.io.CloseableKt.closeFinally(Unknown Source:4)
    at app.revanced.library.ApkUtils.applyTo(Unknown Source:323)
    at app.revanced.manager.flutter.MainActivity.runPatcher$lambda$35(Unknown Source:484)
    at app.revanced.manager.flutter.MainActivity.$r8$lambda$t9VOe3aFJZkCMY59Ya4Pcinu-pA(SourceFile:0)
    at app.revanced.manager.flutter.MainActivity$$ExternalSyntheticLambda6.run(SourceFile:0)
    at java.lang.Thread.run(Thread.java:1012)
Caused by: android.system.ErrnoException: open failed: EACCES (Permission denied)
    at libcore.io.Linux.open(Native Method)
    at libcore.io.ForwardingOs.open(ForwardingOs.java:563)
    at libcore.io.BlockGuardOs.open(BlockGuardOs.java:274)
    at libcore.io.ForwardingOs.open(ForwardingOs.java:563)
    at android.app.ActivityThread$AndroidOs.open(ActivityThread.java:8478)
    at libcore.io.IoBridge.open(IoBridge.java:560)
    ... 11 more
Something went wrong:
[...]

PR #2128, opened 2024-08-09
fix: Don't crash installation when saving last APK is disabled

BenjaminHalko:
closes: #2125

### DIFF
--- a/lib/ui/widgets/settingsView/settings_last_patched_app.dart
+++ b/lib/ui/widgets/settingsView/settings_last_patched_app.dart
@@ -1,7 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:revanced_manager/gen/strings.g.dart';
 import 'package:revanced_manager/ui/views/settings/settings_viewmodel.dart';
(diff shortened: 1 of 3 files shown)


lib/ui/widgets/settingsView/settings_last_patched_app.dart
Axelen123, line 1: Why is this here?

Axelen123 (changes requested):

lib/ui/widgets/settingsView/settings_last_patched_app.dart
BenjaminHalko: The setting didn't have haptics applied, so I figured I'd fix it since I was fix a last Patched App issue anyways. Though maybe it is better to place it in its own PR?

BenjaminHalko: I've already created a separate PR for another last apk issue, so maybe I'll do the same for this.

Axelen123 (approved):
I don't know if the pubspec.lock changes belong here or not, but I don't have anything else to comment on.

PR #2135, opened 2024-08-11
fix: Resolve EACCESS error in special cases

BenCat07:
Sometimes, when there is an attempt to patch an App, that APK is only marked as `r-x` and not `rwx`.  

This can cause the patch to fail, as it cannot read the file after copying it to `patcher/tmp-.../in.apk` 

This PR Fixes that issue (as far as i can tell; tested with brand new install of revanced manager, with and without patch. To Verify)

### DIFF
--- a/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
+++ b/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
@@ -220,6 +220,7 @@ class MainActivity : FlutterActivity() {
         keystorePassword: String
     ) {
         val inFile = File(inFilePath)
+        // Necessary because the file is copied from a nonwriteable location.
         inFile.setWritable(true)
         inFile.setReadable(true)
         val outFile = File(outFilePath)
--- a/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
+++ b/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
@@ -220,6 +220,8 @@ class MainActivity : FlutterActivity() {
         keystorePassword: String
     ) {
         val inFile = File(inFilePath)
+        inFile.setWritable(true)
+        inFile.setReadable(true)
         val outFile = File(outFilePath)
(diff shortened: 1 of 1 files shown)


oSumAtrIX (changes requested):
The changes in this PR should not be necessary and also doesn't explain why the issue only occurs sometimes.

BenCat07:
Cause: Dart/Flutter had no built in permission management system, and the Copy operation of Dart copies the attributes of the original file, including read/write perms.

Why Missing write perms cause `File(path)` to fail, even when it could have been read-only, i do not know, and it would probably need to be checked in java/kotlin docs.

BenCat07:
FileReader instead of File is probably a better idea.

oSumAtrIX:
Why is Manager copying the file and trying to write to it in the first place?

BenCat07:
the "in.apk" is copied to "out.apk". No reading or writing in that case. Could replace the in.apk File with a FoleReader.

Would still need to fix the perms for "out.apk" after patching is complete though

BenCat07:
> Why is Manager copying the file and trying to write to it in the first place?

So to answer this, it isn't writing to it at all. Hut java.io.File seems to be unable to open a file unless that file as a write flag set

BenCat07:
Oh, actually, it seems the backtrace was not accurate due to release mode

```kt
patcherResult.applyTo(inFile)
```


This seems to be the culprit, it applies the patches to in.apk, then signs it and spits out out.apk

android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
```diff
@@ -220,6 +220,8 @@ class MainActivity : FlutterActivity() {
         keystorePassword: String
     ) {
         val inFile = File(inFilePath)
+        inFile.setWritable(true)
```
oSumAtrIX, line 223:
```suggestion
        // Necessary because the file is copied from a nonwriteable location.
        inFile.setWritable(true)
```

oSumAtrIX (approved):

BenCat07:
pog

Issue #1703, opened 2024-02-25
bug: Can't select file paths


szbinks:
### Bug description

On a Samsung phone:
1. Launch Revanced Manager 
2. Patcher 
3. Select an application
4. Storage
5. Recommended version of youtube apk
6. Selected Patches
7. Custom Branding
8. Add options
9. App icon
10. Custom value
11. 3 dots next to ReVanced*logo
12. Select folder
13. Go to the folder that contain required folders with files
14. Tap on "use this folder"
15. Tap on authorize
16. Start patching
17. There's an error which says "No such file or directory" when revanced tries to eccess the given folder

### Error logs

```shell
- Device Info
ReVanced Manager: 1.18.0
Model: SM-G991B
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.google.android.youtube v19.04.37 (Suggested: v19.04.37)
Patches version: v4.2.0
Patches added: Change header, Custom branding [App name: YT Revanced, App icon: content://com.android.externalstorage.documents/tree/primary%3ARevanced%2FRevanced%20Icon%20Pack]
Patches removed: None
Default patch options changed: None

- Settings
Allow changing patch selection: true
Version compatibility check: true
Show universal patches: false
Patches source: revanced/revanced-patches
Integration source: revanced/revanced-integrations

- Logs
Copying APK
Reading APK
Decoding app manifest
Loading patches
Merging integrations
Deleting existing resource cache directory
Decoding resources
Executing patches
Applied 62 patches
Custom branding failed: app.revanced.patcher.patch.PatchException: content:/com.android.externalstorage.documents/tree/primary%3ARevanced%2FRevanced%20Icon%20Pack/mipmap-xxxhdpi/adaptiveproduct_youtube_background_color_108.png: open failed: ENOENT (No such file or directory)
    at app.revanced.patcher.Patcher$apply$1.invokeSuspend$executePatch(Patcher.kt:185)
    at app.revanced.patcher.Patcher$apply$1.invokeSuspend(Patcher.kt:203)
    at app.revanced.patcher.Patcher$apply$1.invoke(Unknown Source:8)
    at app.revanced.patcher.Patcher$apply$1.invoke(Unknown Source:4)
    at kotlinx.coroutines.flow.SafeFlow.collectSafely(Builders.kt:61)
    at kotlinx.coroutines.flow.AbstractFlow.collect(Flow.kt:230)
[...]

PR #2144, opened 2024-08-16
fix: Handle selecting files and folders for patch options correctly

Francesco146:
closes https://github.com/ReVanced/revanced-manager/issues/1703

### DIFF
--- a/lib/ui/widgets/patchesSelectorView/patch_options_fields.dart
+++ b/lib/ui/widgets/patchesSelectorView/patch_options_fields.dart
@@ -573,7 +573,6 @@ class _TextFieldForPatchOptionState extends State<TextFieldForPatchOption> {
                       widget.onChanged(null);
                     },
                   };
-                  print('[patch option] Selected: $selection');
                   if (availableActions.containsKey(selection)) {
                     await availableActions[selection]!();
                   }
--- a/lib/ui/widgets/patchesSelectorView/patch_options_fields.dart
+++ b/lib/ui/widgets/patchesSelectorView/patch_options_fields.dart
@@ -528,19 +528,23 @@ class _TextFieldForPatchOptionState extends State<TextFieldForPatchOption> {
                   ];
                 },
                 onSelected: (String selection) async {
-                  // manageExternalStorage permission is required for file/folder selection
-                  // otherwise, the app will not complain, but the patches will error out
-                  // the same way as if the user selected an empty file/folder.
-                  // Android 11 and above requires the manageExternalStorage permission
-                  final Map<String, dynamic> availableActions = {
-                    'file': () async {
-                      if (_managerAPI.isScopedStorageAvailable) {
-                        final permission =
-                            await Permission.manageExternalStorage.request();
-                        if (!permission.isGranted) {
-                          return;
-                        }
-                      }
+                  Future<bool> gotExternalStoragePermission() async {
+                    // manageExternalStorage permission is required for folder selection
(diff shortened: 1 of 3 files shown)


pubspec.yaml
```diff
@@ -23,6 +23,7 @@ dependencies:
   dynamic_color: ^1.7.0
   dynamic_themes: ^1.1.0
   expandable: ^5.0.1
+  file_picker: ^8.0.5
```
Francesco146, line 26: to use the latest version we need to ubump the minimum Android SDK version to 21

oSumAtrIX: what is it right now/what android version does it correspond to

Francesco146: not sure, but i tried bumping it to the latest but in the next version they dropped support and the build fails

https://pub.dev/packages/file_picker/changelog


I'll take a deeper look soon enough

TheAabedKhan:
I am out of town right now. Wait for around 15 hours please.

pubspec.yaml
Francesco146: @oSumAtrIX appears to be Android SDK 19 (A4.4)

lib/ui/widgets/patchesSelectorView/patch_options_fields.dart
```diff
@@ -524,29 +528,53 @@ class _TextFieldForPatchOptionState extends State<TextFieldForPatchOption> {
...
-                    case 'file':
-                      final String? result = await FlutterFileDialog.pickFile();
-                      if (result != null) {
-                        controller.text = result;
-                        widget.onChanged(controller.text);
+                  // manageExternalStorage permission is required for file/folder selection
+                  // otherwise, the app will not complain, but the patches will error out
+                  // the same way as if the user selected an empty file/folder.
+                  // Android 11 and above requires the manageExternalStorage permission
+                  final Map<String, dynamic> availableActions = {
```
TheAabedKhan, line 535: Reduce code duplication
```suggestion
                  if (selection == 'file' || selection == 'folder') {
                    if (_managerAPI.isScopedStorageAvailable) {
                      final permission =
                          await Permission.manageExternalStorage.request();
                      if (!permission.isGranted) {
                        return;
                      }
                    }
                  }
```

TheAabedKhan (changes requested):
Use `switch-case` block instead of `Map` as it is  potentially faster when dealing with a small, fixed set of cases due to compile-time optimizations.

```dart
switch (selection) {
  case 'file':
    final FilePickerResult? result =
        await FilePicker.platform.pickFiles();
    if (result == null) {
      return;
    }
    controller.text = result.files.single.path!;
    widget.onChanged(controller.text);
    break;
  case 'folder':
    final String? result =
        await FilePicker.platform.getDirectoryPath();
    if (result == null) {
      return;
    }
    controller.text = result;
    widget.onChanged(controller.text);
    break;
  case 'remove':
    widget.removeValue!();
    break;
  case 'null':
    controller.text = '';
    widget.onChanged(null);
    break;
}
```

Francesco146:
will do within 10h

kitadai31:
When select a file with `FilePicker.platform.pickFiles();`, the selected file will be copied to the cache directory of the app.
So there is no need to request the manage all files permission if "Select file path" is pressed.

(Ideally, manager should get the file path directly without copying, though.)

Francesco146:
@kitadai31 I'll remove the check for all permissions in pickFiles, but idk if the second approach is doable rn

TheAabedKhan (approved):

PR #2147, opened 2024-08-18
fix: Screen turns off while patching due to wrong WakeLock

kitadai31:
Fix a issue that screen timeout is activated during patching.
This is caused by WakeLock not being set correctly.

The current code is passing `WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON` to `newWakeLock()`.
But this is wrong.

It is true that the document says that `PowerManager.FULL_WAKE_LOCK` is deprecated and use `WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON` instead.

> Most applications should use WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON instead of this type of wake lock
https://developer.android.com/reference/android/os/PowerManager#FULL_WAKE_LOCK

However, this does not mean passing the `FLAG_KEEP_SCREEN_ON` flag to `newWakeLock()`.
The flag have to be passed to `window.addFlags()`.
https://developer.android.com/develop/background-work/background-tasks/scheduling/wakelock#screen

Therefore, I removed WakeLock and add window.addFlags().

Since the window requires Activity context, it can't be handled from PatcherWorker, so I added it to the UI code.

I've checked this sets KEEP_SCREEN_ON correctly at the start of patching, and clears the flag correctly when succeeded or failed to patching, or patching was canceled.


Axelen123:
The incorrect use of `FLAG_KEEP_SCREEN_ON` for the `PowerManager` seems to originate from the old compose manager. The reason for the wakelock is to try to discourage Android from killing Manager while patching in the background, so just using `FLAG_KEEP_SCREEN_ON` is not sufficient. I believe the correct thing to do is to use `FULL_WAKE_LOCK` and suppress the deprecation, which is fine because it does not seem to be deprecated for our use case.

kitadai31:
> The reason for the wakelock is to try to discourage Android from killing Manager while patching in the background

Isn't this purpose already achieved by foreground service?

Axelen123:
That was what I heard from the previous maintainer, but it was a while ago so I might be wrong. Also, it appears [`PARTIAL_WAKE_LOCK`](https://developer.android.com/reference/android/os/PowerManager#PARTIAL_WAKE_LOCK) would definitely be better than `FULL_WAKE_LOCK` since we only need the CPU, not the screen and it doesn't get silently released if the user presses the power button.

kitadai31:
I could not find any information in the official documentation that foreground services are also limited without WakeLock, but I did find such information on StackOverflow.
https://stackoverflow.com/questions/55170819/android-slows-down-foreground-service-when-device-sleeps
https://stackoverflow.com/questions/52002533/does-service-startforeground-imply-wakelock
According to this information, WakeLock may affect the CPU power when the screen is off.

But WakeLock is not related to background task killing by Android.
Also, this PR fixes the problem that screen turns off during patching.
Therefore I don't think WakeLock is needed since the screen won't be turned off while patching.

What's your opinion?
If you think it's needed, I'll restore the WakeLock and change it to PARTIAL_WAKE_LOCK.

Axelen123:
Keeping the CPU on while patching might be desirable. I am not sure if we should use `FLAG_KEEP_SCREEN_ON`, `PARTIAL_WAKE_LOCK` or both.

@Ushie or @validcube might have an opinion on this.

oSumAtrIX:
What's the progress of this PR?

Axelen123:
I apologize for taking forever to respond to this. Neither of the team members I pinged has any opinions regarding this.

app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
```diff
@@ -85,6 +88,16 @@ fun PatcherScreen(
         }
     }
 
+    if (patcherSucceeded == null) {
+        DisposableEffect(Unit) {
+            val window = (context as Activity).window
+            window.addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
```
Axelen123, line 94: Most examples I have seen for applying flags seem to do it with a lifecycle observer

kitadai31:
Sorry for no responding.
After thinking about it for a few weeks, I have come to the conclusion that it is better to set both `PARTIAL_WAKE_LOCK` and `FLAG_KEEP_SCREEN_ON`.
Because the user may manually turn off the screen during patching.
In this case, the behavior most users would expect is for the app to continue patching even when the screen is off.

app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
kitadai31: Really?
I searched for how to add `FLAG_KEEP_SCREEN_ON` flag in Jetpack Compose and this came up.

Additionally, the flag needs to be cleared when the patching succeeds or fails, so it should have nothing to do with lifecycle.

Axelen123: Hmm, some examples I've seen for `FLAG_SECURE` seem to use lifecycle observer.  Like the second comment [here](https://stackoverflow.com/questions/78694809/is-there-a-way-to-apply-layoutparams-flag-secure-to-compose-screen-level-to-prev). Does the flag stick around if you exit the app and then go back in?

kitadai31: > Does the flag stick around if you exit the app and then go back in?

Yes.
When I leave the app while patching, screen timeout will be activated, and when I back to the app, screen timeout will be disabled again.

This is also described here:
> If an app with the `FLAG_KEEP_SCREEN_ON` flag goes into the background, the system allows the screen to turn off normally. You don't need to explicitly clear the flag in this case.
https://developer.android.com/develop/background-work/background-tasks/awake/screen-on

Axelen123 (approved):
Code seems good then. Can you fix the merge conflict?

kitadai31:
Thanks.
I fixed the conflict.

PR #2148, opened 2024-08-18
feat: Open the app-specific Manage all files permission dialog

kitadai31:
Open the app-specific Manage all files permission dialog directly instead of apps list

Before PR:
image

After PR:
image


validcube (approved):
🥞 Look good to me!

Issue #2140, opened 2024-08-15
feat(Flutter/Compose): Improve the message of incompatible installation


kitadai31:
### Feature description

It says "Contact the developer of the app and ask for support."
However, in most cases, this is caused by the user has used the wrong APK.

I think this message can be like "Check if you are using a correct APK for your device and try again."
Or just only "The app is incompatible with this device." would be fine.

![installation_incompatible](https://github.com/user-attachments/assets/215fd7d7-b3e0-430c-a8a9-6d6d28d7380e)

This is related to compose manager for #1473 

### Motivation

In most cases, this is caused by the user has used the wrong architecture APK.
In this case, asking developer for support is wrong solution.
Therefore, this one sentence is no point.
I think the dialog should only contain instructions that actually lead to a solution.

(I know this may also happen when the device hardware does not meet \<uses-feature\>, but this is a rare case.)

### Acknowledgements

- [X] I have checked all open and closed feature requests and this is not a duplicate.
- [X] I have chosen an appropriate title.
- [X] The feature request is only related to ReVanced Manager.

KobeW50:
What a lovely world :)
I cannot be happier

![image](https://github.com/user-attachments/assets/c22036fb-0c87-48a3-be0f-9aa60a14d39f)

Ushie:
There's no restriction to how many repos you can fork, you just had to rename the repo when forking

KobeW50:
inotia00's rvx manager repository is a fork of this repository, so GitHub considers my fork of inotia00's repository as a fork of this repository.

Here is a screenshot from the fork creation page.
![Screenshot_20240827_015507_Firefox](https://github.com/user-attachments/assets/e78cbbde-4774-49a2-86df-e8d42b8f497c)

kitadai31:
@KobeW50 
> (Made a new GH account to PR for this, since my main account has a fork for inotia00's rvx-manager, which prevents me from creating a fork of this repo)

You don't have to that.
Both your RVX Manager fork and inotia00's RVX Manager belong to the "fork network" of ReVanced/revanced-manager.
So you can open a PR to ReVanced/revanced-manager directly from a fork of another fork.
(That's why GitHub restricts multiple creation of fork.)
Checkout the dev branch of ReVanced/revanced-manager, then create a new branch from the official dev branch, add changes, push the new branch to your fork, and open a PR.

Look this, I opened a PR to compose-dev branch from ReVanced Manager for Android 5-7 repository.
[...]

PR #2164, opened 2024-08-27
feat: Improve "Installation incompatible" dialog message

KobeW50:
Fixes https://github.com/ReVanced/revanced-manager/issues/2140

### DIFF
--- a/assets/i18n/strings.i18n.json
+++ b/assets/i18n/strings.i18n.json
@@ -283,7 +283,7 @@
     "status_failure_timeout_description": "The installation took too long to finish.\n\nWould you like to try again?",
     "status_failure_storage_description": "The installation failed due to insufficient storage.\n\nFree up some space and try again.",
     "status_failure_invalid_description": "The installation failed due to the patched app being invalid.\n\nUninstall the app and try again?",
-    "status_failure_incompatible_description": "The app is incompatible with this device.\n\nEnsure that the APK you are using is supported by this device and try again.",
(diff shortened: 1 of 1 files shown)


assets/i18n/strings.i18n.json
```diff
@@ -283,7 +283,7 @@
     "status_failure_timeout_description": "The installation took too long to finish.\n\nWould you like to try again?",
     "status_failure_storage_description": "The installation failed due to insufficient storage.\n\nFree up some space and try again.",
     "status_failure_invalid_description": "The installation failed due to the patched app being invalid.\n\nUninstall the app and try again?",
-    "status_failure_incompatible_description": "The app is incompatible with this device.\n\nContact the developer of the app and ask for support.",
+    "status_failure_incompatible_description": "The app is incompatible with this device.\n\nEnsure that the APK you are using is supported by this device and try again.",
```
oSumAtrIX, line 286: The APK is incompatible, ensuring it is compatible or retrying when it is incompatible doesn't make sense

oSumAtrIX: You just changed "you are using" to "you use"

KobeW50: "You use" refers to the future, meaning that the user should use an APK that has been ensured to be supported.

It can instead say something such as, 1. `Try again with an APK that is supported by this device.` or 2., `Use an APK that is supported by this device and try again.`

KobeW50: I deleted it because I thought it had the same issue as the current string 🫠

`Use an APK that is supported by this device and try again.`

PR #2193, opened 2024-09-06
fix: Move temporary files outside of the cache directory

kitadai31:
fix for #2120 (Flutter)

But this PR doesn't resolve the issue that the patches jar is located in cacheDir.
However, this PR will reduce the likelihood of problems with patches jars being deleted during patching.
When the storage space is low, Android system will first start deleting the cache of apps that have exceeded their cache quota.
https://developer.android.com/reference/android/content/Context#getCacheDir()
This PR reduces the cache usage of ReVanced Manager during patching, so that the ReVanced Manager will no longer be a priority for Android System to delete the cache.

Also, clear the selected APK file after copied to the patcher temporary directory.
(Because the file picker copies the picked file to the cacheDir when selected APK from storage.)

In addition, clear the patcher temporary directory immediately after finished patching in case of user close the Manager without going back to Dashboard.
Since temporary directory is no longer in the cacheDir, Manager need to clear the temporary directory more reliably than ever.

### DIFF
--- a/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
+++ b/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
@@ -365,6 +365,9 @@ class MainActivity : FlutterActivity() {
                         "An error occurred:\n$stack"
                     )
                 }
+            } finally {
+                inFile.delete()
+                tmpDir.deleteRecursively()
             }
 
             handler.post { result.success(null) }
--- a/lib/services/patcher_api.dart
+++ b/lib/services/patcher_api.dart
@@ -34,7 +34,7 @@ class PatcherAPI {
   Future<void> initialize() async {
     await loadPatches();
     await _managerAPI.downloadIntegrations();
-    final Directory appCache = await getTemporaryDirectory();
+    final Directory appCache = await getApplicationSupportDirectory(); // context.getFilesDir()
     _dataDir = await getExternalStorageDirectory() ?? appCache;
     _tmpDir = Directory('${appCache.path}/patcher');
     _keyStoreFile = File('${_dataDir.path}/revanced-manager.keystore');
@@ -151,6 +151,7 @@ class PatcherAPI {
     String packageName,
     String apkFilePath,
     List<Patch> selectedPatches,
+    bool isFromStorage,
   ) async {
     final File? integrationsFile = await _managerAPI.downloadIntegrations();
     final Map<String, Map<String, dynamic>> options = {};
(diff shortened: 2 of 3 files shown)


oSumAtrIX:
Why was patches jar not moved out of cache?

kitadai31:
Fixing this would require a significant rewrite of the Manager's network stack and management of patches files.

oSumAtrIX:
Hm, can you elaborate? You should need to change the save location I would assume

kitadai31:
Currently, the Manager relies on Flutter dio's http cache to manage patches and integrations.
Everything relies on dio's cache, from patching an app offline to handling patch updates.
This is the reason.

I thought about changing the dio caches place to outside of cacheDir, but dio caches other than patches should still be placed in cacheDir and be clearable from the app info.
Also, in the first place, it is highly likely that the location of the dio cache cannot be changed. (I did a little research.)

Anyway, with this change, the Android system will no longer prioritize deleting the ReVanced Manager cache.
The patches jar is very small in size, so I think it is less likely to exceed the "quota" and get deleted.
So this should work as a quick fix until Compose Rewrite is completed.

android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
```diff
@@ -365,6 +365,9 @@ class MainActivity : FlutterActivity() {
                         "An error occurred:\n$stack"
                     )
                 }
+            } finally {
+                inFile.delete()
+                tmpDir.deleteRecursively()
```
oSumAtrIX, line 370: Isn't this done on the flutter side already?

kitadai31: Certainly, pressing the back button to return to the dashboard, or installing the app and pressing the "Open" button delete entire patcher temporary directory in flutter side.
However, if the user closes the app directly, all the temp files will remain.
Therefore, I chose to delete everything except `out.apk` from the patcher temporary directory immediately after patch ended.

(Since the temporary directories are no longer inside the cacheDir, I believe the Manager is now more responsible than ever for ensuring that the temporary directories are cleared.)

lib/services/patcher_api.dart
```diff
@@ -34,7 +34,7 @@ class PatcherAPI {
   Future<void> initialize() async {
     await loadPatches();
     await _managerAPI.downloadIntegrations();
-    final Directory appCache = await getTemporaryDirectory();
+    final Directory appCache = await getApplicationSupportDirectory(); // context.getFilesDir()
```
validcube, line 37: What's the commented code for?

```dart
context.getFilesDir()
```

validcube (approved):
🪐 Looks fine to me!

lib/services/patcher_api.dart
kitadai31: I added it because I thought it was unclear what "ApplicationSupportDirectory" referred to.
but if it is not necessary, please remove it.

validcube: Sure,

```suggestion
    final Directory appCache = await getApplicationSupportDirectory();
```

validcube (approved):
🥞 LGTM!

PR #2194, opened 2024-09-06
fix: Empty “tmp-XXXXXX” directory keeps growing in cacheDir

kitadai31:
Every time Manager starts, an empty directory is created in `/data/data/app.revanced.manager.flutter/cache/cache/` .

many_empty_dirs

Use a native [`getCodeCacheDir()`](https://developer.android.com/reference/android/content/Context#getCodeCacheDir()) for the optimizedDexDirectory.
This is a right place to use for optimizedDexDirectory.

### DIFF
--- a/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
+++ b/android/app/src/main/kotlin/app/revanced/manager/flutter/MainActivity.kt
@@ -109,14 +109,13 @@ class MainActivity : FlutterActivity() {
 
                 "getPatches" -> {
                     val patchBundleFilePath = call.argument<String>("patchBundleFilePath")!!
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
But no dex is stored there? Why does it matter?

kitadai31:
I noticed this little problem while I was working on the previous PR, so I thought I'd fix it while I was at it.

Also, iirc, optimizedDirectory will be used for Android 8.0.
The optimized dex acrtually remains available even when you restart the app, so optimizing a dex every time Manager is started is an overhead.

oSumAtrIX (approved):

validcube (approved):
🥞 LGTM!

validcube:
🥞 Merged, thank!

PR #2230, opened 2024-09-29
fix: Don't show toasts when export cancelled

TheAabedKhan:


lib/ui/views/settings/settings_viewmodel.dart
```diff
@@ -233,8 +233,11 @@ class SettingsViewModel extends BaseViewModel {
             sourceFilePath: outFile.path,
             fileName: 'selected_patches_$dateTime.json',
           ),
-        );
-        _toast.showBottom(t.settingsView.exportedPatches);
+        ).then((status) {
```
oSumAtrIX, line 236: Instead of then, simply do final status = await FlutterFileDialog.saveFile...

oSumAtrIX (approved):
Apart from the one review lgtm

oSumAtrIX (approved):

PR #2231, opened 2024-09-30
fix: Match "Installation incompatible" dialog message with Flutter Manager

kitadai31:
#2140 has been solved on both Flutter and Compose, but Flutter Manager has a better message that was reviewed by oSumAtrIX. (#2164)
I think it is better to reflect it to Compose Manager also.


oSumAtrIX (approved):

validcube (approved):
🥞 LGTM!

PR #2255, opened 2024-10-11
Create codeql.yml

Selemela07:


oSumAtrIX:
For now, code analysis tools like this are not necessary. We can consider them later with dedicated attention and a project-specific setup. Thanks anyways.

PR #2263, opened 2024-10-17
ci: Use `semantic-release-pub` for updating build number

CnC-Robert:


.releaserc
```diff
@@ -17,15 +17,11 @@
...
+      "semantic-release-pub",
       {
-        "files": [
-          {
-            "path": ["pubspec.yaml"],
-            "type": "flutter",
-            "branches": ["main", "dev"]
-          }
-        ]
+        "cli": "flutter",
```
oSumAtrIX, line 22: I don't think this needs to be set. The publishPub field would prevent the use of the cli I would assume.

validcube (approved):
🥞🪄 Looks good to me

ReVanced Manager uses build number (versionCode) that's proportionate to the build version (versionName) here

https://github.com/ReVanced/revanced-manager/blob/dev/pubspec.yaml#L7

(it's broken)

The dependency only increments by one, but I'll let it pass because it doesn't make sense to have versionCode the same as versionName.

PR #2268, opened 2024-10-22
feat: Import and export manager settings

TheAabedKhan:
Export Manager Settings (1)

> [!NOTE] 
> Patch options and patched apps info are also included.

Let me know if those values shouldn't be included.


indrastorms:
Instead of multiple import/export options show a dialog to select what(settings, patch selection) user wants to export would be clean way to do import/export

TheAabedKhan:
> Instead of multiple import/export options show a dialog to select what(settings, patch selection) user wants to export would be clean way to do import/export

Thank you for the suggestion. However, this is irrelevant to this PR. You can open a separate issue regarding it so that everyone can discuss it properly, and we can track the enhancement separately. This will also help keep the scope of this PR focused and ensure smoother reviews.

assets/i18n/strings.i18n.json
```diff
@@ -198,6 +198,12 @@
     "deleteTempDirLabel": "Delete temporary files",
     "deleteTempDirHint": "Delete unused temporary files",
     "deletedTempDir": "Temporary files deleted",
+    "exportManagerSettingsLabel": "Export manager settings",
+    "exportManagerSettingsHint": "Export manager settings to a JSON file",
+    "exportedManagerSettings": "Manager settings exported",
+    "importManagerSettingsLabel": "Import manager settings",
+    "importManagerSettingsHint": "Import manager settings from a JSON file",
+    "importedManagerSettings": "Manager settings imported",
```
oSumAtrIX, line 206: Just saying "Export settings" is shorter. Adjust labels too.

oSumAtrIX (changes requested):
Also add these settings to the docs under docs/ in this PR.Apart from the naming suggestion lgtm.

PR #2315, opened 2024-11-13
ci: Generate release artifact provenance

validcube:
Support GitHub Attestation for attesting the binary, matched SLSA requirements *level 2* which provides secure provenance for verification of authenticity of thg APK before installing onto devices.


.github/workflows/release.yml
```diff
@@ -63,3 +65,14 @@ jobs:
           KEYSTORE_ENTRY_PASSWORD: ${{ secrets.KEYSTORE_ENTRY_PASSWORD }}
         run: |
           npx semantic-release
+          
+      - name: Find APK file
+        id: find_apk
```
oSumAtrIX, line 70: Why is the step given an ID?

.github/workflows/release.yml
```diff
@@ -63,3 +65,14 @@ jobs:
...
+        run: |
+          APK_LOCATION=$(find build/app/outputs/apk/release/ -name "revanced-manager-*.apk")
+          echo "APK_LOCATION=$APK_LOCATION" >> $GITHUB_ENV
+          
+      - name: 🛰️ Generate artifact attestation
+        uses: actions/attest-build-provenance@v1
+        with:
+          subject-path: ${{ env.APK_LOCATION }}
```
oSumAtrIX, line 78: Can these both steps be merged into one artifact attestation step?

validcube:
👋 It can, but I've trouble getting the wildcard to work. Will try again
later.

oSumAtrIX:
Wouldn't *.apk work?

validcube:
I've tried that in
https://github.com/ReVanced/revanced-manager/pull/2315/commits/9024539e08252be3a6f9883db8cd55862f99685f
but that doesn't seem to work so I add additional steps to finding the APK.

I might be doing something wrong, if it is, lmk!

.github/workflows/release.yml
validcube: This is a placeholder from GitHub Copliot inline code suggestions, can be remove.

.github/workflows/release.yml
```diff
@@ -63,3 +65,8 @@ jobs:
           KEYSTORE_ENTRY_PASSWORD: ${{ secrets.KEYSTORE_ENTRY_PASSWORD }}
         run: |
           npx semantic-release
+          
+      - name: 🛰️ Generate artifact attestation
```
validcube, line 69: Clean up

```suggestion
      - name: Generate artifact attestation
```

validcube (commented):
Clean-up, restore ci step naming consistency (remove emoji)

validcube:
ReVanced Manager Compose variant of this PR is available here: 
* #2324

validcube:
Update, the PR support wildcard, it's just that during testing semantic release will always don't build APK if it isn't configured to publish.

Issue #2224, opened 2024-09-23
bug: Manager sometimes immediately cancels installation and Export APK button stops working


KobeW50:
### Bug description

Sometimes when trying to install a patched app, the Manager will immediately cancel the installation. The Export APK button will also not work. The issue only occurs when the `Save patched app` toggle in the Manager settings is enabled.

It is difficult to reproduce this bug. Below is the simplest and most consistent method I found.

**Steps to reproduce**

1. Prepare the replication environment by uninstalling YouTube ReVanced and clearing the ReVanced Manager data
2. Open the Manager settings and enable the `Allow changing patch selection` option
3. Select the suggested YouTube APK from storage
4. Patch and install YouTube with only the `GmsCore support` patch applied (the bug should not occur at this point)
5. Without closing the Manager, return to the Patcher screen and repeat step 3 and 4

The bug should occur when repeating step 4 for the second time. (You may need to try again)

Demonstration:

https://github.com/user-attachments/assets/5a5ca50d-dd6f-42f1-9236-34bacc239999

### Version of ReVanced Manager and version & name of app you are patching

1.22.0
YouTube 

### Installation method

Regular

### ReVanced Manager logs

[revanced-manager_logcat_20240923114749497367.log.txt](https://github.com/user-attachments/files/17100715/revanced-manager_logcat_20240923114749497367.log.txt)

### Patch logs

```shell
- Device Info
ReVanced Manager: 1.22.0
Model: SM-S906U1
Android version: 14
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Root permissions: No

- Patch Info
App: com.google.android.youtube v19.16.39 (Suggested: 19.16.39)
Patches version: v4.15.0
Patches added: Default
[...]

PR #2332, opened 2024-11-30
fix: Use cache directory for the last patched app

BenjaminHalko:
This should hopefully solve the issue where it can't write the patched app for some reason

### DIFF
--- a/lib/services/manager_api.dart
+++ b/lib/services/manager_api.dart
@@ -367,7 +367,7 @@ class ManagerAPI {
     File outFile,
   ) async {
     deleteLastPatchedApp();
-    final Directory appCache = await getApplicationSupportDirectory();
+    final Directory appCache = await getApplicationCacheDirectory();
     app.patchedFilePath =
         outFile.copySync('${appCache.path}/lastPatchedApp.apk').path;
     app.fileSize = outFile.lengthSync();


BenjaminHalko:
This needs to be tested before merging. I can't get the original issue to appear.

oSumAtrIX:
According to the doc, this change should not make a difference:

https://stackoverflow.com/questions/71823734/what-is-the-difference-between-gettemporarydirectory-and-getapplicationsupport

KobeW50:
> This needs to be tested before merging. I can't get the original issue to appear.

Does this fix https://github.com/ReVanced/revanced-manager/issues/2224? If yes, you can send me an APK with the changes and I'll test it

ILoveOpenSourceApplications:
This PR was marked as ready for review yet there's no review. Maybe some one can look into this?

ILoveOpenSourceApplications:
@validcube, This PR was marked as ready for review yet there's no review. Maybe you can look into this?

kitadai31:
I tested this PR and found two issues

1. "Installation cancelled" is fixed, but instead, the patched app is not saved to the last patched app section once every two times

|              | Before PR              | After PR                                                                                 | 
| ------------ | ---------------------- | ---------------------------------------------------------------------------------------- | 
| 1st patching | OK                     | OK                                                                                       | 
| 2nd patching | Installation cancelled | The installation is successful, but the patched app is not saved to the last patched app | 
| 3rd patching | OK                     | OK                                                                                       | 
| 4th patching | Installation cancelled | The installation is successful, but the patched app is not saved to the last patched app | 

2. When the above issue (1) occurs, the `tmp-XXXXXX.apk` file (last patched app) is left in the app data and become an orphan file.
Since the file names are randomized, it will never be overwritten.
There is no way to remove it other than clearing the app data.

I think randomization doesn't really solve the problem.

I found another simple solution, so please check a PR that I will open.

lib/services/patcher_api.dart
```diff
@@ -182,7 +182,15 @@ class PatcherAPI {
...
-    outFile = File('${workDir.path}/out.apk');
+    if (_managerAPI.isLastPatchedAppEnabled()) {
+      // Get a random name from temp dir
+      // This is to avoid overwriting the last patched app
+      final String randomName = workDir.path.split('/')[workDir.path.split('/').length - 1];
+      final Directory cacheDir = await getApplicationSupportDirectory();
+      outFile = File('${cacheDir.path}/$randomName.apk');
+    } else {
+      outFile = File('${workDir.path}/out.apk');
+    }
```
oSumAtrIX, line 193: This change may be reverted according to the comment in #2490. The change above can be kept.

lib/services/manager_api.dart
```diff
@@ -367,9 +367,7 @@ class ManagerAPI {
     File outFile,
   ) async {
     deleteLastPatchedApp();
-    final Directory appCache = await getApplicationSupportDirectory();
-    app.patchedFilePath =
-        outFile.copySync('${appCache.path}/lastPatchedApp.apk').path;
+    app.patchedFilePath = outFile.path;
```
oSumAtrIX, line 370: But outfile is a temporary file that is eventually deleted no?

lib/services/patcher_api.dart
BenjaminHalko: @oSumAtrIX Which change above can be kept?

lib/services/manager_api.dart
BenjaminHalko: When last patched app is enabled, outfile points to application support directory.

lib/services/patcher_api.dart
```diff
@@ -175,7 +175,12 @@ class PatcherAPI {
     final File inApkFile = File('${workDir.path}/in.apk');
     await File(apkFilePath).copy(inApkFile.path);
 
-    outFile = File('${workDir.path}/out.apk');
+    if (_managerAPI.isLastPatchedAppEnabled()) {
+      final Directory cacheDir = await getApplicationSupportDirectory();
+      outFile = File('${cacheDir.path}/out.apk');
```
kitadai31, line 180: `getApplicationSupportDirectory()` is equivalent to Android's `context.getFilesDir()`, not cacheDir
So the naming of `cacheDir` is confusing
I think `filesDir` would be a more appropriate name for this variable

ILoveOpenSourceApplications:
@validcube, apart from merge conflicts, is this ready?

validcube:
Hello!

This PR will be closed as we have transitioned from Flutter to Compose.

🥞 Thanks.

PR #2362, opened 2024-12-23
feat: switch to androidx.navigation

Axelen123:
Some benefits include predictive back gestures support and nicer nested navigation.


Axelen123:
I am not sure what you mean by that.

PR #2363, opened 2024-12-23
Revert "feat: switch to androidx.navigation (#2362)"

Monday1908:
This reverts commit 5d3a81f4b9bba11ac342be0543e2c5d61c69f5c8.


validcube:
Please provide more detail as to why you're reverting this commit, maybe there is an issue(s) caused by the changes or is there another reason why adding the commit _might_ be a bad idea for the project, because if there's please let us know.

PR #2367, opened 2024-12-26
Updated Polish translation

Patrxgt:
I've updated Polish translation to make it more coherent and fix an issue with the "Patcher" tab when using bigger fonts than system's default.


MarcaDian:
Use the official crowdin project to translate into Polish
https://crowdin.com/project/revanced/pl

PR #2381, opened 2025-01-06
feat: Redesign the patches screen

X1nto:
This redesigns the patcher screen to reduce clutter. The search bar is now expandable, the patch count is displayed in a bottom bar and the reset button floats neatly above the save button. I added a clear button to the search bar to clear the queries. The filter chips now have a checkmark when selected and the filters have been reworded.

| Before | After |
|--------|-------|
|image|image|

New patcher screen in action:

[Screen_recording_20250109_003652.webm](https://github.com/user-attachments/assets/ffbef204-2118-486a-a162-ff7c37c31b77)

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
@@ -12,6 +12,8 @@ import androidx.compose.animation.scaleIn
 import androidx.compose.animation.scaleOut
 import androidx.compose.animation.shrinkOut
 import androidx.compose.animation.togetherWith
+import androidx.compose.foundation.ExperimentalFoundationApi
+import androidx.compose.foundation.background
 import androidx.compose.foundation.clickable
 import androidx.compose.foundation.layout.*
 import androidx.compose.foundation.lazy.LazyListScope
@@ -37,7 +39,13 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.draw.alpha
+import androidx.compose.ui.draw.drawWithContent
 import androidx.compose.ui.draw.rotate
+import androidx.compose.ui.geometry.Offset
+import androidx.compose.ui.graphics.BlendMode
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.CompositingStrategy
+import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.dp
@@ -64,7 +72,9 @@ import app.revanced.manager.util.isScrollingUp
 import app.revanced.manager.util.transparentListItemColors
 import kotlinx.coroutines.launch
 
-@OptIn(ExperimentalMaterial3Api::class, ExperimentalLayoutApi::class)
+@OptIn(ExperimentalMaterial3Api::class, ExperimentalLayoutApi::class,
+    ExperimentalFoundationApi::class
+)
 @Composable
 fun PatchesSelectorScreen(
     onSave: (PatchSelection?, Options) -> Unit,
@@ -186,6 +196,20 @@ fun PatchesSelectorScreen(
         )
     }
 
+    fun LazyListScope.patchCount(background: @Composable () -> Color) {
+        stickyHeader(key = "count", contentType = 2) {
+            Text(
+                text = stringResource(R.string.patches_selected, selectedPatchCount, availablePatchCount),
+                modifier = Modifier
+                    .fillMaxWidth()
+                    .background(background())
+                    .padding(top = 8.dp, end = 16.dp),
+                style = MaterialTheme.typography.titleMedium,
+                textAlign = TextAlign.End
+            )
+        }
+    }
+
     fun LazyListScope.patchList(
         uid: Int,
         patches: List<PatchInfo>,
@@ -195,14 +219,15 @@ fun PatchesSelectorScreen(
     ) {
         if (patches.isNotEmpty() && visible) {
             header?.let {
-                item {
+                item(contentType = 0) {
                     it()
                 }
             }
 
             items(
                 items = patches,
-                key = { it.name }
+                key = { it.name },
+                contentType = { 1 }
             ) { patch ->
                 PatchItem(
                     patch = patch,
@@ -306,6 +331,8 @@ fun PatchesSelectorScreen(
                         it.name.contains(query, true)
                     }
 
+                    patchCount(background = { MaterialTheme.colorScheme.surfaceContainerHigh })
+
                     patchList(
                         uid = bundle.uid,
                         patches = bundle.supported.searched(),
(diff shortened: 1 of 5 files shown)


oSumAtrIX (commented):
I'll leave a review purely on what I saw in the PR body and leave the rest to the other reviewers: 

- I think in the patching screen the progress bar is at the top, however in the patch selector screen it is at the bottom, perhaps this is a consistency issue we want to avoid
- The progress bar in the patch selector screen makes it look like you wanna have all patches. "Reaching 100%". However this probably isn't what we want to imply via the UI

LGTM apart from that

X1nto:
I can remove the progress bar from the bottom and keep the `x/y selected`, or move that to the top too

oSumAtrIX:
Satisfies the issues I mentioned 👍

X1nto:
Updated the screenshot and the video showcase according to requested changes

MarcaDian:
Is it a big problem to make the number of selected patches as it is done in flutter in the Save button?
It seems to me that UI will be neater.

<details>

oSumAtrIX:
Works for me and avoids this label at the top aligned to the right 👍 LGTM apart from that based on the visuals.

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -222,102 +261,146 @@ fun PatchesSelectorScreen(
...
-            val bundle = bundles[pagerState.currentPage]
-
-            LazyColumnWithScrollbar(
-                modifier = Modifier.fillMaxSize()
-            ) {
-                fun List<PatchInfo>.searched() = filter {
-                    it.name.contains(query, true)
+    Scaffold(
+        topBar = {
+            app.revanced.manager.ui.component.SearchBar(
```
Axelen123, line 266: It should be possible to import this directly if you get rid of the `compose.material3.*` import.

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -222,102 +261,146 @@ fun PatchesSelectorScreen(
...
+                leadingIcon = {
+                    val rotation = remember { Animatable(0f) }
+                    LaunchedEffect(searchExpanded) {
+                        val direction = if (searchExpanded) 360f else 0f
+
+                        rotation.animateTo(
+                            targetValue = direction,
+                            animationSpec = tween(durationMillis = 400, easing = EaseInOut)
+                        )
+                    }
```
Axelen123, line 283: Wouldn't `animateFloatAsState` work for this?

```kt
val rotation by animateFloatAsState(
	if (searchExpanded) 360f else 0f,
	remember { tween(durationMillis = 400, easing = EaseInOut) }
)
```

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -222,102 +261,146 @@ fun PatchesSelectorScreen(
...
+                    IconButton(onClick = {
+                        if (searchExpanded) {
+                            setSearchExpanded(false)
+                        } else {
+                            onBackClick()
+                        }
+                    }) {
```
Axelen123, line 290:
```suggestion
                    IconButton(
                        onClick = {
                            if (searchExpanded) {
                                setSearchExpanded(false)
                            } else {
                                onBackClick()
                            }
                    	}
                    )  {
```

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
X1nto: You're right, this is just a leftover from a different type of animation I was testing with. Why remember the animation spec? First time seeing that, not gonna lie

X1nto:
How do we handle the FAB patch count if the FAB is not expanded?

Ushie:
Just hide it, as it is in the current version

sexstarvd:
Hey mate what am I meant to be hiding I'm not very good with computers I may make downloaded this for me

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
Monday1908: 266

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
Axelen123: The source code for the `animate*AsState` functions use `remember` or global `private val`s for the default animation spec. Probably doesn't make a difference though.

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -375,7 +444,7 @@ fun PatchesSelectorScreen(
                         patchList(
                             uid = bundle.uid,
                             patches = bundle.unsupported,
-                            filterFlag = SHOW_UNSUPPORTED,
+                            visible = vm.filter and SHOW_SUPPORTED == 0,
```
Axelen123, line 447: I think `HIDE_UNSUPPORTED` would be more accurate, since supported patches are shown regardless of the flag.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
X1nto: How would you implement that? Unless I change the filter logic in the ViewModel, HIDE_UNSUPPORTED is derived from SHOW_UNSUPPORTED.

Axelen123: You can just change the name of the constant in the viewmodels companion object. The viewmodel doesn't assume anything about what the flag actually does.

X1nto: I'd also have to flip one the filter chips at line 150.

From 
```kt
CheckedFilterChip(
    selected = vm.filter and SHOW_SUPPORTED != 0,
    onClick = { vm.toggleFlag(SHOW_SUPPORTED) },
    label = { Text(stringResource(R.string.supported)) }
)
```

To

```kt
CheckedFilterChip(
    selected = vm.filter and HIDE_UNSUPPORTED == 0,
    onClick = { vm.toggleFlag(HIDE_UNSUPPORTED) },
    label = { Text(stringResource(R.string.supported)) }
)
```

Axelen123: Hmm, in that case I guess the flag can be named `SHOW_UNSUPPORTED`. The filter chip should match the behavior of the flag though. It seems weird to have a flag named `SHOW_SUPPORTED` controlling visibility of the unsupported patches or a chip called "supported" that only affects the unsupported patches.

its0x08 (commented):
Looks good

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
X1nto: Updated the constant name, see if the change makes sense

Axelen123 (approved):

Axelen123:
Thank you

PR #2398, opened 2025-01-27
feat(Update Screen): Use BottomAppBar instead of a Header

Ushie:


Ushie:
image
image



Should I move the ProgressIndicator to the top of the screen like the Patcher screen? Or should I move the PatcherScreen's down?

oSumAtrIX:
Top is the expected location

Ushie:
image

0xrxL:
> image

Move the total MBs downloaded on the top (top-right to be precise) as well. The current position don't have much sense.

validcube:
> image

Currently it feel awkward having the progress indicator at the top, and total download at the bottom.

I'd suggest removing the total download since progress indicator is enough or use @0xrxL's suggestion.

oSumAtrIX:
Agreeing to both comments above. The bottom bar is not really conventional, can be moved to the top as well.

federicogith:
Instead of completely removing the total downloaded, it would look better on the top right as suggested by 0xrxL

PR #2403, opened 2025-02-01
feat: Reorder Import & Export settings

Ushie:
Before:
image

After:
image


validcube (approved):
While thinking about it I slightly have a different concept in mind like 

Keystore
* keystore related stuff
Patch Selection
* Patch Selection related stuff
Patch Option
* Patch Option related stuff

But overall this is fine to me:tm:

Ushie:
It would be nice if we all just sat down and rethought all of settings, because settings should be done from the groundup, for now this will do until then

validcube:
> It would be nice if we all just sat down and rethought all of settings, because settings should be done from the groundup, for now this will do until then
> 
> 

We could start one right now, I can draft up initial design and page the team up later for discussion.

validcube:
Draft @ #2413

PR #2420, opened 2025-02-12
feat: Improve trust plugin dialog design

### LINKED ISSUE #2061: feat: Improve trust plugin dialog design
### Feature description

Right now the content is one block of continuous text:

image

Instead add spacing, titles and similar 

### Motivation

Right now its difficult to read what begins where

### Acknowledgements

- [X] This issue is not a duplicate of an existing feature request.
- [X] The feature request is only related to ReVanced Manager

Ushie:
Closes #2061

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/settings/DownloadsSettingsScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/settings/DownloadsSettingsScreen.kt
@@ -2,7 +2,9 @@ package app.revanced.manager.ui.screen.settings
 
 import androidx.annotation.StringRes
 import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
@@ -10,10 +12,13 @@ import androidx.compose.foundation.lazy.items
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.filled.Delete
 import androidx.compose.material3.AlertDialog
+import androidx.compose.material3.Card
+import androidx.compose.material3.CardDefaults
 import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.MaterialTheme
+import androidx.compose.material3.OutlinedCard
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Text
 import androidx.compose.material3.TextButton
@@ -31,8 +36,10 @@ import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.input.nestedscroll.nestedScroll
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.style.TextAlign
+import androidx.compose.ui.tooling.preview.Preview
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.zIndex
 import androidx.lifecycle.compose.collectAsStateWithLifecycle
@@ -59,6 +66,7 @@ fun DownloadsSettingsScreen(
     val downloadedApps by viewModel.downloadedApps.collectAsStateWithLifecycle(emptyList())
     val pluginStates by viewModel.downloaderPluginStates.collectAsStateWithLifecycle()
     val scrollBehavior = TopAppBarDefaults.pinnedScrollBehavior(rememberTopAppBarState())
+    val context = LocalContext.current
 
     Scaffold(
         topBar = {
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/revanced/manager/ui/screen/settings/DownloadsSettingsScreen.kt
```diff
@@ -229,10 +241,27 @@ fun DownloadsSettingsScreen(
...
+private fun PreviewTrustDialog() {
+    TrustDialog(
+        title = R.string.downloader_plugin_trust_dialog_title,
+        body = stringResource(
+            R.string.downloader_plugin_trust_dialog_body,
+        ),
+        onDismiss = { },
+        onConfirm = { },
+        pluginName = "app.revanced.manager.apkmirror",
+        signature = "23 01 84 F6 0B AE 2F EA F2 44 F1 0A 8B AC 05 3C 8F F3 3A 18 3B CC 36 5B 4D 8B 87 6D 2B 7F 48 09"
```
validcube, line 255: While not stated in my RVMC/Weberie variant, you can use semi-colon ( : ) as separator like

2F:EA:F2:44 ... Like MAC Address

validcube:
Please ignore this is a test. I'm using your PR as playground for invalid PR

app/src/main/java/app/revanced/manager/ui/screen/settings/DownloadsSettingsScreen.kt
Ushie: That's not really typical display of signatures

oSumAtrIX:
Offtopic, but why does it say "ReVanced Manager: ..."?

oSumAtrIX:
I think the importance of this screen and dialog is enabling and disabling plugins. Trusting them is a secondary task to achieve that main objective. Strings can be renamed to say "Enable" or "Disable" plugin. And the dialog can mention that enabling the plugin can run arbitrary code and need to be trusted as a "side notice"

Ushie:
> Offtopic, but why does it say "ReVanced Manager: ..."?

Placeholder representation of the app name

oSumAtrIX:
I am not sure what you mean by placeholder. What would be written there if others develop plugins?

Ushie:
The app name of the plugin — for example, the APKMirror plugin would be named ReVanced Manager: APKMirror. This accurately represents what the app is and its purpose. If it were just APKMirror, you wouldn't know it is for ReVanced Manager in the app list, and that it isn't actually by APKMirror.

oSumAtrIX:
It should probably say "ReVanced: APKMirror downloader" or "Ushie: Some other downloader"

app/src/main/java/app/revanced/manager/ui/screen/settings/DownloadsSettingsScreen.kt
```diff
@@ -129,15 +136,18 @@ fun DownloadsSettingsScreen(
                                     .digest(androidSignature.toByteArray())
                                 hash.toHexString(format = HexFormat.UpperCase)
                             }
+                        val appName = packageInfo.applicationInfo?.loadLabel(context.packageManager)
+                            ?.toString()
+                            ?: packageName
```
Axelen123, line 141: Please `remember` this.

Axelen123 (changes requested):

Axelen123 (approved):

app/src/main/java/app/revanced/manager/ui/screen/settings/DownloadsSettingsScreen.kt
```diff
@@ -242,10 +256,27 @@ fun DownloadsSettingsScreen(
...
+        title = R.string.downloader_plugin_trust_dialog_title,
+        body = stringResource(
+            R.string.downloader_plugin_trust_dialog_body,
+        ),
+        onDismiss = { },
+        onConfirm = { },
+        pluginName = "app.revanced.manager.apkmirror",
+        signature = "23 01 84 F6 0B AE 2F EA F2 44 F1 0A 8B AC 05 3C 8F F3 3A 18 3B CC 36 5B 4D 8B 87 6D 2B 7F 48 09"
+    )
+}
```
brosssh: Should this stay? We have no preview components anywhere else.

Ushie: No, it shouldn't stay, I was using that for my own convenience during development

Although I wouldn't mind if we later down the line made it standard to use Previews where applicable

brosssh:
Can this be merged?

Axelen123:
If the designers are ok with it then yes

Ushie:
What do you think about this?

https://github.com/ReVanced/revanced-manager/pull/2420#discussion_r2073494466

app/src/main/res/values/strings.xml
```diff
@@ -137,7 +137,10 @@
     <string name="downloader_plugin_state_untrusted">Untrusted</string>
     <string name="downloader_plugin_trust_dialog_title">Trust plugin?</string>
     <string name="downloader_plugin_revoke_trust_dialog_title">Revoke trust?</string>
-    <string name="downloader_plugin_trust_dialog_body">Package name: %1$s\nSignature (SHA-256): %2$s</string>
+    <string name="downloader_plugin_trust_dialog_body">By continuing you\'ve agreed to running external plugins.\n\nDo not allow any suspicious plugin(s) to be installed as they can run arbitrary code.</string>
```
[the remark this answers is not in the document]
KobeW50: Also, "plugin(s)" can just be "plugins".

KobeW50:
> I think the importance of this screen and dialog is enabling and disabling plugins. Trusting them is a secondary task to achieve that main objective. Strings can be renamed to say "Enable" or "Disable" plugin. And the dialog can mention that enabling the plugin can run arbitrary code and need to be trusted as a "side notice"

I agree. In addition to changing the button to "Enable"/"Disable", I think the other button should be "Cancel" instead of "Dismiss". "Cancel" is already used in other similar dialogs (see the Advanced section in the settings).

A new _issue_ is that when using the words "Enable"/"Disable", there is no longer any purpose for the dialog when disabling it. Disabling a plugin inherently means that it will no longer function or run its code. Therefore, I think it would be fitting for the plugins to have a toggle, and the dialog would only appear when the toggle is being enabled.

Here is a proposal for the dialog when enabling a plugin: 
```
Enabling this plugin will allow it to run arbitrary code. Only enable this plugin if you trust it.
```

If the goal is to sound somewhat scary, "run arbitrary code" can be substituted with "execute arbitrary code".

PR #2422, opened 2025-02-12
fix: Use `compatible` rather than `support` when referring to patch compatibility

### LINKED ISSUE #2067: fix: Use `compatible` rather than `support` when referring to patch version compatibility
### Bug description

Currently ReVanced Manager uses the word support to refer to patch compatibility across the app, it might be better to change it to compatible to be consistent with the term

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] The bug is only related to ReVanced Manager

Ushie:
Closes #2067


validcube (approved):
🥞 LGTM!

PR #2485, opened 2025-04-17
feat: Add toggle to use pre-releases

TheAabedKhan:
Screenshots
pre-release


lib/services/github_api.dart
```diff
@@ -33,12 +33,12 @@ class GithubAPI {
...
-  ) async {
+  Future<Map<String, dynamic>?> getLatestRelease(String repoName) async {
+    final String prerelease =
+        _managerAPI.usePrereleases() ? '?per_page=1' : '/latest';
     try {
       final response = await _dioGetSynchronously(
-        '/repos/$repoName/releases/latest',
+        '/repos/$repoName/releases$prerelease',
       );
       return response.data;
```
kitadai31, line 43: Alternative sources are not working.

The `/releases/latest` endpoint returns a single release object, whereas the `/releases` endpoint returns an array of releases.
(even if you specified `?per_page=1`)
So, this have to be `return response.data[0];` for pre-releases.

```
E/flutter (22529): [ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: type 'List<dynamic>' is not a subtype of type 'FutureOr<Map<String, dynamic>?>'
E/flutter (22529): #0      GithubAPI.getLatestRelease (package:revanced_manager/services/github_api.dart:43:23)
```

kitadai31 (commented):
This enables Manager pre-releases, but currently, Manager update check is disabled for dev versions.

https://github.com/ReVanced/revanced-manager/blob/a8e192b85f433210bf7df7049d653119c8da1f16/lib/ui/views/home/home_viewmodel.dart#L137-L140

(This `_managerAPI.releaseBuild` variable will only be true if the app is release build and the version name doesn't contain `-dev`.)

This needs to be addressed.

lib/ui/widgets/settingsView/settings_advanced_section.dart
```diff
@@ -20,6 +21,7 @@ class SAdvancedSection extends StatelessWidget {
       title: t.settingsView.advancedSectionTitle,
       children: const <Widget>[
         SAutoUpdatePatches(),
+        SUsePrereleases(),
```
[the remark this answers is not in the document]
TheAabedKhan: @oSumAtrIX

assets/i18n/en.i18n.json
```diff
@@ -180,6 +180,8 @@
     "disablePatchesSelectionWarningText": "You are about to disable changing the selection of patches.\nThe default selection of patches will be restored.\n\nDisable anyways?",
     "autoUpdatePatchesLabel": "Auto update patches",
     "autoUpdatePatchesHint": "Automatically update patches to the latest version",
+    "usePrereleasesLabel": "Use pre-releases",
+    "usePrereleasesHint": "Use pre-releases of ReVanced Manager and ReVanced Patches",
```
oSumAtrIX, line 184:
```suggestion
    "usePrereleasesHint": "Use pre-release versions of ReVanced Manager and ReVanced Patches",
```

lib/services/github_api.dart
```diff
@@ -33,13 +33,16 @@ class GithubAPI {
...
+        _managerAPI.usePrereleases() ? '?per_page=1' : '/latest';
     try {
       final response = await _dioGetSynchronously(
-        '/repos/$repoName/releases/latest',
+        '/repos/$repoName/releases$target',
       );
+      if (_managerAPI.usePrereleases()) {
+        return response.data.first;
+      }
       return response.data;
```
oSumAtrIX, line 46: @kitadai31 would this work:
```suggestion
        return response.data.first;
```

lib/ui/widgets/settingsView/settings_advanced_section.dart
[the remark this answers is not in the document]
oSumAtrIX: Data source is fine. Manager updates is also provided by a data source (API)

oSumAtrIX (changes requested):
In addition, the toggle should warn the user and suggest not to use dev releases via a dialog. On toggle, the user should be prompted to restart to consume the new data.

lib/services/github_api.dart
TheAabedKhan: No that won't work. `/latest` provides Map, the other one provides List. So, need to add the condition.

oSumAtrIX (approved):

Issue #2526, opened 2025-05-06
bug(Compose): Using Dark mode, some labels are still black, making them unreadable


Brosssh:
### Bug description

![Image](https://github.com/user-attachments/assets/9ada77ae-eab6-4b0b-a661-70f98cc1be58)
![Image](https://github.com/user-attachments/assets/056cbd1b-1212-42b4-894e-2bdda9b338d6)
![Image](https://github.com/user-attachments/assets/4bfa4195-537f-4284-8bbb-0b9a2167a400)

### Version of ReVanced Manager and version & name of app you are patching

Latest version of `compose-dev` branch, just pulled it, `.\gradlew build`, `adb install .\app\build\outputs\apk\debug\app-debug.apk`

### Installation method

Regular

### ReVanced Manager logs

```shell
Are logs needed for this?
```

### Patch logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

oSumAtrIX:
I can not reproduce. Are you forcing dark mode with your OS?

Brosssh:
System default is light mode, so yes I'm forcing manager to be dark mode.

Brosssh:
![Screenshot_2025-05-06-13-31-54-07_761fe8f2c59677c7476dd46708cfb34f.jpg](https://github.com/user-attachments/assets/bbadb8fc-916b-44cb-9d86-814d838e28d8)

Phone specs

oSumAtrIX:
The same applies when using light mode in manager when phone is in dark mode:

![Image](https://github.com/user-attachments/assets/4f2324fc-c639-42e1-b52e-a05f7a34f652)

Brosssh:
Correction, after I restart the app, the error is gone and I can clearly see the text the intended way. So this is probably not related to changing the system theme, but just the cached theme (or something similar, at least).

PR #2527, opened 2025-05-06
fix(Compose): Reset cached theme on theme change

Brosssh:
Fix of https://github.com/ReVanced/revanced-manager/issues/2526

Most probably not done in the best way, sorry :)


Axelen123 (changes requested):
The fix seems fine, but I think the name `resetListItemColorsCached()` makes more sense.

Issue #2423, opened 2025-02-12
build: Code analysis, styling enforcement


validcube:
Currently in Compose variant, styles are inconsistent with almost every files.

A viewmodel might be defined as `vm`, while some files might defined literally as `viewModel`

https://github.com/ReVanced/revanced-manager/blob/compose-dev/app/src/main/java/app/revanced/manager/ui/screen/settings/AdvancedSettingsScreen.kt#L64

https://github.com/ReVanced/revanced-manager/blob/compose-dev/app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt#L45

Some of the files are importing using star import (`import *`) while at the same time importing each every entities. Some code editor by default enforce changing star import to individual import like as seen on:

https://github.com/ReVanced/revanced-manager/pull/2421/files#diff-4316f1a9d3abbb8ff718f8eaf53828f708cf56c11c2c28924b4addce1e3f848eR10-R15

https://github.com/ReVanced/revanced-manager/pull/2422/files#diff-5d8eef944c2d224a454e6414c1a0a90b057c4a1336b19fcde0632bd0856a8feaR33-R38

...

And there might be more inconsistencies not listed in this issue.

Ushie:
Readability should be the priority, for example vm should be expanded to viewModel instead of being shortened.

PR #2529, opened 2025-05-07
feat(Compose): Add confirmation dialog on multiple operations 

brosssh:
Added confirmation dialog when:
- Deleting patch bundles (single and multiple selection)
- Deleting download apps
- Stopping the patching process

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
@@ -17,7 +17,7 @@ import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.outlined.OpenInNew
-import androidx.compose.material.icons.outlined.Delete
+import androidx.compose.material.icons.outlined.Cancel
 import androidx.compose.material.icons.outlined.FileDownload
 import androidx.compose.material.icons.outlined.PostAdd
 import androidx.compose.material.icons.outlined.Save
@@ -105,7 +105,7 @@ fun PatcherScreen(
             },
             title = stringResource(R.string.patcher_stop_confirm_title),
             description = stringResource(R.string.patcher_stop_confirm_description),
-            icon = Icons.Outlined.Delete
+            icon = Icons.Outlined.Cancel
         )
     }
 


app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
```diff
@@ -0,0 +1,45 @@
...
+import androidx.compose.material3.AlertDialog
+import androidx.compose.material3.Icon
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import app.revanced.manager.R
+
+@Composable
+fun BundleDeleteDialog(
```
Brosssh, line 14: A single Dialog for both single and multiple bundle deletion. Not sure if it's the correct approach

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
```diff
@@ -0,0 +1,45 @@
...
+    onDismiss: () -> Unit,
+    onConfirm: () -> Unit,
+    bundleName: String? = null
```
Brosssh, line 17: Using `bundleName` to undestand if the deletion if from a single selection. This is most probably not the best way, but i'm not sure what that would be

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
```diff
@@ -0,0 +1,45 @@
...
+                Text(stringResource(R.string.confirm))
+            }
+        },
+        title = { Text(stringResource(R.string.delete_bundle_dialog_title)) },
+        icon = { Icon(Icons.Outlined.Delete, null) },
+        text = { Text(
+            if (bundleName != null)
+                stringResource(R.string.delete_bundle_single_dialog_description, bundleName)
+            else
+                stringResource(R.string.delete_bundle_multiple_dialog_description)
```
Brosssh, line 42: This will show "the selected bundles" (plural) even if only a bundle is selected. Not sure if it's ok

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleItem.kt
```diff
@@ -42,6 +42,7 @@ fun BundleItem(
     toggleSelection: (Boolean) -> Unit,
 ) {
     var viewBundleDialogPage by rememberSaveable { mutableStateOf(false) }
+    var showDeleteConfirmationDialog by remember { mutableStateOf(false) }
```
Brosssh, line 45: `remember` or `rememberSaveable`?

Axelen123: `rememberSaveable` makes more sense in order to preserve the value on screen rotation.

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
```diff
@@ -0,0 +1,45 @@
...
+                    onConfirm()
+                    onDismiss()
+                }
+            ) {
+                Text(stringResource(R.string.confirm))
+            }
+        },
+        title = { Text(stringResource(R.string.delete_bundle_dialog_title)) },
+        icon = { Icon(Icons.Outlined.Delete, null) },
+        text = { Text(
```
Axelen123, line 38: Can you change the formatting so it is consistent with the rest of the file?

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
Axelen123: Using a single dialog seems fine here.

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
Axelen123: The UID would be better since it is guaranteed to be unique unlike the name.

app/src/main/java/app/revanced/manager/ui/screen/DashboardScreen.kt
```diff
@@ -154,6 +155,17 @@ fun DashboardScreen(
         }
     )
 
+    var showDeleteConfirmationDialog by remember { mutableStateOf(false) }
```
Axelen123, line 158:
```suggestion
    var showDeleteConfirmationDialog by rememberSaveable { mutableStateOf(false) }
```

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
Axelen123: Isn't the if statement supposed to handle that?

app/src/main/res/values/strings.xml
```diff
@@ -425,4 +425,8 @@
     <string name="show_manager_update_dialog_on_launch_description">Shows a popup notification whenever there is a new update available on launch.</string>
     <string name="failed_to_import_keystore">Failed to import keystore</string>
     <string name="export">Export</string>
+    <string name="delete_bundle_dialog_title">Delete bundle</string>
+    <string name="delete_bundle_single_dialog_description">Are you sure you want to delete the bundle %s?</string>
+    <string name="delete_bundle_multiple_dialog_description">Are you sure you want to delete the selected bundles?</string>
```
Axelen123, line 430: Move these to somewhere around here: https://github.com/ReVanced/revanced-manager/blob/8a463be02081fe1bf3e01deaa962b0ecd864ed5f/app/src/main/res/values/strings.xml#L338

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
Brosssh: Not sure I understand, I'm using the `bundleName` just to display the name of the bundle it's going to be deleted. Why would i need the UID here?

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
Brosssh: If I delete a bundle from inside its view, it's going to set the `bundleName`, hence showing `Are you sure you want to delete the bundle My Bundle Name?`.

If I use the multi select feature, but I only select a bundle, the `bundleName` will be null, making it display the default message `Are you sure you want to delete the selected bundles?`.

Currently, if the user multi select a single bundle, the plural message will be shown. Ideally I think it would be better to show the first message if a single bundle is selected via multi select. But it's a very minor improvement and I couldn't find a easy way to do it.

Maybe check my video in the PR description, it may help understand.

Brosssh:
Please also check my new comments https://github.com/ReVanced/revanced-manager/pull/2529#discussion_r2079105018 https://github.com/ReVanced/revanced-manager/pull/2529#discussion_r2079091173.

Brosssh:
Added a fix for https://github.com/ReVanced/revanced-manager/issues/2534 since it's basically the same issue, abstracted the DeleteDialog even more since I used only that dialog for every action.

app/src/main/java/app/revanced/manager/ui/component/GenericDeleteDialog.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.material3.AlertDialog
+import androidx.compose.material3.Icon
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.res.stringResource
+import app.revanced.manager.R
+
+@Composable
+fun GenericDeleteDialog(
```
oSumAtrIX, line 14: Isn't there another dialog component that can be reused? This dialog is only "delete" because of the icon. icon could be parametrized.

oSumAtrIX, line 14: DeleteDialog is enough for the name :)

Brosssh: I'm not sure if there is already a dialog like this. Else it can be created as ConfirmDialog (?).

Axelen123: There isn't. ConfirmDialog seems like a better name though.

app/src/main/java/app/revanced/manager/ui/component/GenericDeleteDialog.kt
```diff
@@ -0,0 +1,41 @@
...
+    onDismiss: () -> Unit,
+    onConfirm: () -> Unit,
+    title: @Composable () -> Unit,
+    description: @Composable () -> Unit
```
Axelen123, line 18: Maybe these should take a string resource instead. Dialogs that need custom composable code can use `AlertDialog` directly.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
Axelen123: Oh, I misinterpreted that. You are right that its unnecessary. Using the bundle name like that is fine, since the parameter name kinda implies that there will only be one bundle if it is not null.

app/src/main/java/app/revanced/manager/ui/component/bundle/BundleDeleteDialog.kt
Axelen123: Oh now I get it. Displaying the plural message when you have only selected one bundle doesn't seem like a big deal.

app/src/main/java/app/revanced/manager/ui/component/ConfirmDialog.kt
```diff
@@ -0,0 +1,41 @@
...
+import androidx.compose.material3.Icon
+import androidx.compose.material3.Text
+import androidx.compose.material3.TextButton
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.res.stringResource
+import app.revanced.manager.R
+
+@Composable
+fun ConfirmDialog(
```
oSumAtrIX, line 13: In the app various dialogs are shown. Can't they use this component?

app/src/main/res/values/strings.xml
```diff
@@ -336,6 +338,11 @@
     <string name="bundle_view_patches">View patches</string>
     <string name="bundle_view_patches_any_version">Any version</string>
     <string name="bundle_view_patches_any_package">Any package</string>
+    <string name="bundle_delete_single_dialog_title">Delete bundle</string>
+    <string name="bundle_delete_multiple_dialog_title">Delete bundles</string>
+    <string name="bundle_delete_single_dialog_description">Are you sure you want to delete the bundle %s?</string>
```
oSumAtrIX, line 343: Maybe quotes should be used here around the name.

app/src/main/java/app/revanced/manager/ui/component/ConfirmDialog.kt
Brosssh: Yes, but adjusting all existing dialogs with this may be a bit of work that can maybe be postponed after other issues are fixed.

Is it ok to postpone this in to 2nd phase? 
Future dialogs should of course use this component.

Axelen123: I don't think that has to be done in this PR, but it wouldn't be wrong either.

app/src/main/java/app/revanced/manager/ui/component/ConfirmDialog.kt
```diff
@@ -0,0 +1,41 @@
...
+    onDismiss: () -> Unit,
+    onConfirm: () -> Unit,
+    title: String,
+    description: String,
+    imageVector: ImageVector
```
Axelen123, line 18: `icon` makes more sense to me.

Axelen123 (changes requested):

Brosssh:
This is also going to close https://github.com/ReVanced/revanced-manager/issues/2540.

As per https://github.com/ReVanced/revanced-manager/issues/2423, `vm` has been renamed to `viewModel`.

oSumAtrIX:
In the downloads section in the settings, downloaded apps can be deleted, those have also no confirmation

Brosssh:
> In the downloads section in the settings, downloaded apps can be deleted, those have also no confirmation

image0 (1)

Already implemented in this PR.

Can you please add link to https://github.com/ReVanced/revanced-manager/issues/2540?

Axelen123 (approved):

Issue #2520, opened 2025-05-05
bug: Padding missing between search and patches list


oSumAtrIX:
### Bug description

See:

![Image](https://github.com/user-attachments/assets/08b8b8d9-b646-440b-b573-c3ce1d16538a)

### Version of ReVanced Manager and version & name of app you are patching

-

### Installation method

None

### ReVanced Manager logs

```shell
-
```

### Patch logs

```shell
-
```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

PR #2533, opened 2025-05-08
Fix(Compose): Add padding in patch selection screen

Brosssh:
Closes https://github.com/ReVanced/revanced-manager/issues/2520

Enough? Too much? 
Screenshot_2025-05-08-10-25-45-49_761fe8f2c59677c7476dd46708cfb34f


Axelen123 (approved):
Seems good to me. @validcube or @Ushie can answer your question regarding the padding.

validcube (approved):
🥞 Padding looks good to me.

Ushie:
Can the padding be made more consistent with other screens?

I believe I use verticalArrangement for Columns across the app, that can be used for the dialog padding

As for the patches page padding, I don't exactly remember what it's called but I don't think I outright add padding, I think it might be called innerPadding in lazycolumns or something, also, why is there a lack of padding to begin with? there's supposed to be a default amount of padding, that should be looked into

brosssh:
@Ushie 
I checked but `verticalArrangement` is not adding the padding. I believe that is used to space items in a `Column` component, while I need to add it to the top of the component..

Ushie:
The verticalArrangement thing is for the add bundle dialog

contentPadding is for the patches screen, I'm not sure if contentPadding will work for landscape but it's worth a try

brosssh:
@Ushie 
Again, `verticalArrangement` is not adding the padding (unless I'm missing something)

image0

image.

For the patches screen, `contentPadding` is not a valid parameter for a `Column` attribute, that only exists for `LazyColumn`.

PR #2541, opened 2025-05-10
feat: Rename "Patch bundle" to "Patches"

oSumAtrIX:
Optionally "Patches file" can be used.

For reference: https://discord.com/channels/952946952348270622/952987428786941952/1370824632156033185
For now, only strings are named. Code and internal changes can be made after Compose Manager release


app/build.gradle.kts
Axelen123, line 1: I don't the project will build with these changes.

app/src/main/java/app/revanced/manager/ui/screen/DashboardScreen.kt
```diff
@@ -77,7 +77,7 @@ enum class DashboardPage(
     val icon: ImageVector
 ) {
     DASHBOARD(R.string.tab_apps, Icons.Outlined.Apps),
-    BUNDLES(R.string.tab_bundles, Icons.Outlined.Source),
+    BUNDLES(R.string.tab_patches, Icons.Outlined.Source),
```
Axelen123, line 80: The name of the enum variant wasn't changed.

app/src/main/res/values/strings.xml
```diff
@@ -132,8 +130,8 @@
     <string name="reset_patch_selection_success">Patch selection has been reset</string>
     <string name="patch_options_reset_package">Reset patch options for app</string>
     <string name="patch_options_reset_package_description">Resets patch options for a single app</string>
-    <string name="patch_options_reset_bundle">Resets patch options for bundle</string>
-    <string name="patch_options_reset_bundle_description">Resets patch options for all patches in a bundle</string>
+    <string name="patch_options_reset_patches">Resets patch options for this patches</string>
```
Axelen123, line 133: "for this patches" doesn't sound right.

app/src/main/res/values/strings.xml
```diff
@@ -185,8 +183,8 @@
     <string name="debug_logs_export_success">Exported logs</string>
     <string name="api_url">API URL</string>
     <string name="api_url_description">The API used to download necessary files.</string>
-    <string name="api_url_dialog_title">Set custom API URL</string>
-    <string name="api_url_dialog_description">Set the API URL of ReVanced Manager. ReVanced Manager uses the API to download patches and updates.</string>
+    <string name="api_url_dialog_title">Change API URL</string>
+    <string name="api_url_dialog_description">Change the API URL of ReVanced Manager. ReVanced Manager uses the API to download patches and updates.</string>
```
Axelen123, line 187: Are these changes within the scope of this PR?

Axelen123 (changes requested):

app/build.gradle.kts
oSumAtrIX: i built and ran it successfully

app/src/main/java/app/revanced/manager/ui/screen/DashboardScreen.kt
oSumAtrIX: Code changes need separate dedicated time to handle. The code is currently written like Bundles being containers of patches: Bundle(setOf<Patch>()).

However the code should've been: Bundle : Set<Patch>.
This semantical issue makes it difficult to "just" rename the code as we'd have semantical issues like:

Patches(setOf<Patch>()) which would mean: patches.patches

app/src/main/res/values/strings.xml
oSumAtrIX: Reset <bundle name>'s patch options would work. What do you say

app/src/main/res/values/strings.xml
oSumAtrIX: If we are exact no, but it was simpler to do here than switch branches, should be okay since its minor

app/build.gradle.kts
Axelen123: Those changes are gone now, so I will mark this as resolved.

app/src/main/java/app/revanced/manager/ui/screen/DashboardScreen.kt
Axelen123: I'll mark this as resolved. I forgot this PR was only about strings.

app/src/main/res/values/strings.xml
Axelen123: That sounds better.

app/src/main/res/values/strings.xml
Axelen123: I guess it is fine. I've sometimes been doing that myself.

app/src/main/res/values/strings.xml
oSumAtrIX: see my latest force push.

brosssh: "this patches" is still there

Axelen123: Yeah

Axelen123:
Seems like this was partially merged earlier, but now I will fully merge it to resolve the review comments.

brosssh:
I was wondering what I did wrong in the past 2 rebases I attempted, turns out they were correct and this was just already partially merged 🙂.

PenuasDaGod (commented):
Testing ph.

PR #2543, opened 2025-05-12
ci: Run Crowdin Cron task on dev branch

LisoUseInAIKyrios:


oSumAtrIX (approved):

LisoUseInAIKyrios:
If dev branch isn't going to merge to main soon, then this commit should be picked and merged to main.

Because the Cron task runs off the main branch, and currently Crowdin keeps adding and deleting the same strings because Cron is deleting strings that are missing from main but any commit to dev is pushing the new dev strings.

Crowdin users keep getting messages of "new strings added" when they're not new but the same strings that were previously added then deleted 5+ times.

Issue #2538, opened 2025-05-09
bug: Cannot distinguish between different versions of patch bundles in patches selector


kitadai31:
### Bug description

If different versions of the same patch bundle are added, they are not distinguishable in the patch selection screen.

<img width="480" src="https://github.com/user-attachments/assets/b548d114-1ac6-4aee-a36a-34b811f08ec9">

<br>

This is actually v5.22.0 on the left and v5.21.0 on the right.

<img width="360" src="https://github.com/user-attachments/assets/9145b340-26d5-4b6b-9d49-1333c5d9f823">

For example, the current latest version of ReVanced Patches is v5.22.0.
If I downloaded ReVanced Patches v5.21.0 from https://github.com/ReVanced/revanced-patches/releases/tag/v5.21.0 and added to Manager, I will run face this issue.

#### Possible solution

Add patch bundle versions to the patches selector.

Maybe make the tabs two lines and show versions in it?

<img width="360" src="https://github.com/user-attachments/assets/562f72fe-d244-4e08-a280-3bec9fa8723e">

### Version of ReVanced Manager and version & name of app you are patching

compose-dev 4d522664b61d6bf34cff7d7080ee452a433d6507 `fix: Reset cached theme on theme change to avoid broken colors (#2527)`

### Installation method

None

### ReVanced Manager logs

```shell
.
```

### Patch logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

PR #2545, opened 2025-05-14
feat(Compose): Improve patches selector tab by adding the bundle version

Brosssh:
image0 (1)

Closes https://github.com/ReVanced/revanced-manager/issues/2538

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
@@ -405,7 +405,19 @@ fun PatchesSelectorScreen(
                                     )
                                 }
                             },
-                            text = { Text(bundle.name) },
+                            text = {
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -405,7 +405,19 @@ fun PatchesSelectorScreen(
...
                             },
-                            text = { Text(bundle.name) },
+                            text = {
+                                Column(horizontalAlignment = Alignment.CenterHorizontally) {
+                                    Text(
+                                        text = bundle.name,
+                                        style = MaterialTheme.typography.bodyMedium
+                                    )
+                                    Text(
+                                        text = bundle.version!!,
```
Brosssh, line 415: The `!!` should be safe since if a patch is not fetched, it's not displayed in the tab. I don't see how version can be null otherwise.

Axelen123 (approved):

Issue #2056, opened 2024-07-13
bug: Universal patches safeguard warning dialog can simply be dismissed


oSumAtrIX:
### Bug description

This should require a toggle in the settings,  instead it allows simply clicking continue:

![image](https://github.com/user-attachments/assets/6bd9ce2a-3c4a-496f-8036-a2a0e5423915)

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The bug is only related to ReVanced Manager

Brosssh:
This can be closed as completed

![Screenshot_2025-05-15-08-10-05-55_761fe8f2c59677c7476dd46708cfb34f.jpg](https://github.com/user-attachments/assets/530abf61-4bb2-4074-86ed-d9a189607633)

Brosssh:
Nevermind, this still applies for universal patches selection, my bad.

Axelen123:
This was fixed by the pull request.

PR #2550, opened 2025-05-15
fix(Compose): Adjusted universal patches safeguard and warnings

Brosssh:
- Made the toggle title and description consistent with "Allow changing patch selection". 
- Replaced `SafeguardConfirmationDialog` in favour of the common `ConfirmDialog`. 
- Removed the "Continue" option from the warning message when selecting an universal patch.
- More replacements vm -> viewModel.

Closes https://github.com/ReVanced/revanced-manager/issues/2056


app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -192,10 +191,9 @@ fun PatchesSelectorScreen(
     if (showSelectionWarning) {
         SelectionWarningDialog(onDismiss = { showSelectionWarning = false })
     }
-    vm.pendingUniversalPatchAction?.let {
+    viewModel.pendingUniversalPatchAction?.let {
```
Axelen123, line 194: Storing a lambda is probably unnecessary now since continuing is no longer possible.

Axelen123 (changes requested):

Axelen123 (approved):

Issue #2557, opened 2025-05-20
bug: Patch selection is not initially shown


brosssh:
### Bug description

Shows 4 even if only a patch is selected.

![Image](https://github.com/user-attachments/assets/50f51a71-12f1-4545-8256-72da4eb0e8f0)
![Image](https://github.com/user-attachments/assets/15dc306f-2647-4d9c-961d-8bb0387b3097)

### Version of ReVanced Manager and version & name of app you are patching

Compose

### Installation method

None

### ReVanced Manager logs

```shell
.
```

### Patch logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

Axelen123:
Closing since a linked PR was merged

PR #2559, opened 2025-05-20
fix(Compose): Fix selected patch count

Brosssh:
This line is wrong since it's possible the user enable the "Allow changing patch" selection, change patches, turn it back off.

Closes https://github.com/ReVanced/revanced-manager/issues/2557.


Brosssh:
This last commit is unrelated to this PR but it's so easy it's probably not worth a separate PR.

app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
```diff
@@ -151,8 +151,6 @@ class SelectedAppInfoViewModel(
 
         // Try to get the previous selection if customization is enabled.
         viewModelScope.launch {
-            if (!prefs.disableSelectionWarning.get()) return@launch
```
Axelen123, line 154: Removing the early return will make manager use old selections, which shouldn't happen unless patch customization is enabled.

Axelen123 (changes requested):

Brosssh:
This took an embarassing amount of time for me to figure it out :D.

app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
```diff
@@ -151,7 +151,7 @@ class SelectedAppInfoViewModel(
...
+            if (!prefs.disableSelectionWarning.getBlocking()) return@launch
```
Axelen123, line 154: This fix feels really hacky assuming it works, since `getBlocking` is not made for coroutines. What happens if you add a `delay(1000L)` before this line? Will the fix break entirely or will it work after 1 second?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
Brosssh: > What happens if you add a delay(1000L) before this line? Will the fix break entirely or will it work after 1 second?

It indeed breaks the "fix". Honestly I'm not sure what the correct approach for fixing this would be..

Axelen123: I am guessing this is caused by the reference to the `MutableState` not pointing to the actual state being used when we set the value. Try changing `selection.value = (...)` into `selectionState = (...)`. Do the same for patch options.

brosssh: I think the problem is `selectionState.patches(bundles, allowIncompatible)`. This is called before the coroutine `viewModelScope.launch` in `selectionState` is over, so by default it's set to `SelectionState.Default`. The only way I found of "fixing" this is by replacing the `launch` with a `runBlocking`. I'm not sure if this is safe so please let me know.

Axelen123: `runBlocking` should ideally be avoided. Did you try the method I described?

brosssh: I tried but 
image

Axelen123: In that case manually set the type of `selectionState`: `private var selectionState: SelectionState by savedStateHandle.saveable {`

brosssh: It works, I'll study why this works because I don't understand it :D .

app/src/main/java/app/revanced/manager/ui/viewmodel/SelectedAppInfoViewModel.kt
```diff
@@ -155,7 +155,7 @@ class SelectedAppInfoViewModel(
 
             val previous = selectionRepository.getSelection(packageName)
             if (previous.values.sumOf { it.size } == 0) return@launch
-            selection.value = SelectionState.Customized(previous)
+            selectionState = SelectionState.Customized(previous)
```
Axelen123, line 158: Can you do this for the options as well? The code is similar so it might be experiencing the same issue.

Axelen123 (changes requested):

Axelen123 (approved):

Issue #2189, opened 2024-09-03
feat: Send push notifications when patch bundles are updated


KobeW50:
### Feature description

Add an option to the ReVanced Manager settings that allows users to choose how often to check for updates to the user's patch bundles. There should be an option to never check.

The following screenshot is from [Bunny Manager](https://github.com/pyoncord/BunnyManager):
![SmartSelect_20240903_161714_Bunny Manager](https://github.com/user-attachments/assets/3883e500-29e7-4130-9750-dee6d47af137)

When an update is found, a push notification should be delivered to the user, informing the user which patch bundle has been updated.

### Motivation

Currently, users need to open the Manager to see if patches have been updated. This can easily be forgotten by users.

Push notifications for updates is a useful utility for many users, and it will promote people to keep their patched apps up-to-date.

### Acknowledgements

- [X] I have checked all open and closed feature requests and this is not a duplicate.
- [X] I have chosen an appropriate title.
- [X] The feature request is only related to ReVanced Manager.

Ushie:
Why would users be interested in the patch bundles updating? Unless major fixes have been made, it's not intended for users to constantly check up on the patch bundles

KobeW50:
Many people prefer to keep their software up-to-date. Not just for major fixes, but also minor bug fixes, as well as newly implemented features.

KobeW50:
That's fine. Also it'sworth clarifying that the setting is only for checking in the background. There should always be a check when the user opens the app

Brosssh:
This can probably be closed as the bundle auto update feature exists. No notification is needed if the manager already does the refresh in the background (imo).

PalmDevs:
@Ushie If this were to be implemented, how do you plan on implementing it? I know that Android allows apps to periodically fetch data every set interval. Perhaps we can do that?

Ushie:
Yeah but I'm thinking about the necessity of this, there's no particular reason to update patches unnecessarily, besides you might be a user of ReVanced Twitter for example and not care about reddit, yet nothing says that ReVanced Twitter is going to be updated in the next 20 releases

PR #2561, opened 2025-05-21
feat: Add periodic background bundle updates

brosssh:
This is still WIP, but I open the PR to both get feedback and to keep track of what I'm doing.

### Feature:
Basically what is described here https://github.com/ReVanced/revanced-manager/issues/2189: A scheduled job will run each X time (choosable by the user); it will check if an update is available for every remote bundle. A bundle can be excluded by deselecting the "Search for updates" toggle in the bundle settings. If an update was found, send a push notification (if permission were granted, else do not run the job).

Edit:
I've removed the TODO list since it would contains too many item now, I've made too many changes :^). I'll provide a video explaining what I did.

### Open points:
- [ ] [My devices will force close the app when swiping the app away from the recent app menu](https://stackoverflow.com/a/52605503). Force closing the app will stop every background task. Pretty sure there is nothing that can be done, but if anyone knows a way for the job to run even of force close please let me know.
- [x] ~~Does it still make sense to have the auto update on app opening? Should that send a notification as well if it updates a bundle?~~ Edit: Yes. The bundles are going to be updated still on app opening, while the search will be in the background

Any feedback is appreciated :).

### DIFF
--- a/app/src/main/java/app/revanced/manager/domain/bundles/RemotePatchBundle.kt
+++ b/app/src/main/java/app/revanced/manager/domain/bundles/RemotePatchBundle.kt
@@ -10,9 +10,9 @@ import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.withContext
 import org.koin.core.component.inject
 import java.io.File
+import kotlinx.coroutines.flow.combine
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.first
-import kotlinx.coroutines.flow.flow
 
 
 class RemotePatchBundleFetchResponse(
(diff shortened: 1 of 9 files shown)


Ushie:
I feel like time options less than a day shouldn't be available, as it would encourage users to immediately update when updates drop

Wouldn't this be better if it were push notifications instead? The patch bundle source would be needed to host something but it would prevent the ReVanced Manager from having to poll for updates so frequently if we're going to include such little times

oSumAtrIX:
Push notifications requires some kind of web socket server from what I understand. Poll works fine thanks to Cloudflare caching as well as the /version response which is small

Brosssh:
Current behavior:

https://github.com/user-attachments/assets/7493074d-eacf-4cc9-93e0-a5dc12484e25

Brosssh:
> I feel like time options less than a day shouldn't be available, as it would encourage users to immediately update when updates drop

Time options and strings shown to the user can (and should) be adjusted. For now the time options are like this because it's easier for me to understand if the job is running correctly.

Ushie:
If the user doesn't wish to be notified of the bundle update then no background polling should happen, the bundles should be updated on app launch in that case

PR #2565, opened 2025-05-23
fix: "Save patched app" attempts to copy APK when patching fails

kitadai31:
Fix the issue that this error is always thrown when patching failed.

```
Something went wrong:
PathNotFoundException: Cannot copy file to '/data/user/0/app.revanced.manager.flutter/files/lastPatchedApp.apk', path = '/data/user/0/app.revanced.manager.flutter/files/patcher/tmp-ESURPI/out.apk' (OS Error: No such file or directory, errno = 2)
```

This was caused by a bug that "Save patched app" attempts to copy the patched APK even if patching failed.

This error was generated regardless of the actual reason why the patch failed, and was adding unnecessary noise to troubleshooting.

Since this PR conflicts with #2332, please merge it first.
After that I will fix the conflict.

### DIFF
--- a/lib/ui/views/installer/installer_viewmodel.dart
+++ b/lib/ui/views/installer/installer_viewmodel.dart
@@ -125,7 +125,7 @@ class InstallerViewModel extends BaseViewModel {
     });
     await WakelockPlus.enable();
     await handlePlatformChannelMethods();
-    await runPatcher(context);
+    await runPatcher();
   }
 
   Future<dynamic> handlePlatformChannelMethods() async {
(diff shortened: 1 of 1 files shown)


validcube:
🥞 Thanks

validcube:
Please ignore build failure notice, this is an problem when Flutter upgraded it dependency constraint. I'll push a fix soon:tm:

beerpormpang:
Something went wrong:
PathNotFoundException: Cannot copy file to '/data/user/0/app.revanced.manager.flutter/files/lastPatchedApp.apk', path = '/data/user/0/app.revanced.manager.flutter/files/patcher/tmp-ESURPI/out.apk' (OS Error: No such file or directory, errno = 2)

Issue #2578, opened 2025-05-30
bug: Manager should respect `Disable version compatibility check`


cyberboh:
### Bug description

Test to download `YT Music` version `8.21.51` (the supported one is `8.05.51`) error appeared as shown on the log below.
I am using [apkmirror downloader plugin 1.0.0](https://github.com/ReVanced/revanced-manager-apkmirror-downloader/releases/tag/v1.0.0)

![Image](https://github.com/user-attachments/assets/3fac5c20-2072-45ce-8d1d-3e66d49244b5)

![Image](https://github.com/user-attachments/assets/b4059175-6732-4617-8114-eaacb333e0ec)

### Version of ReVanced Manager and version & name of app you are patching

0.0.1 lastets commit https://github.com/ReVanced/revanced-manager/pull/1409/commits/7c5552f93fcd8747ebe58aec6b3debb4ede324a4

### Installation method

None

### ReVanced Manager logs

```shell
See image for the log
```

### Patch logs

```shell
none
```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

PR #2579, opened 2025-05-30
fix: Allow different app version when downloading via plugin if setting is off

brosssh:
Closes https://github.com/ReVanced/revanced-manager/issues/2578.

### DIFF
--- a/app/src/main/java/app/revanced/manager/domain/repository/DownloadedAppRepository.kt
+++ b/app/src/main/java/app/revanced/manager/domain/repository/DownloadedAppRepository.kt
@@ -96,7 +96,7 @@ class DownloadedAppRepository(
             val pkgInfo =
                 pm.getPackageInfo(targetFile.toFile()) ?: error("Downloaded APK file is invalid")
             if (pkgInfo.packageName != expectedPackageName) error("Downloaded APK has the wrong package name. Expected: $expectedPackageName, Actual: ${pkgInfo.packageName}")
-            if (expectedVersion != null && pkgInfo.versionName != expectedVersion) error("Downloaded APK has the wrong version. Expected: $expectedVersion, Actual: ${pkgInfo.versionName}")
+            if (expectedVersion != null && pkgInfo.versionName != expectedVersion) error("The version of the app you have selected does not match the suggested version.\nPlease use the suggested version: $expectedVersion\n\nTo continue anyway, disable \"Require suggested app version\" in the advanced settings.")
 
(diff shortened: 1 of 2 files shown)


brosssh:
Uhm I see. Since the user select the patches before downloading the APK, is there an "easy" way so that we can allow diffent version while still keeping the compatibility check?

Ushie:
Why is that needed? The version defined by ReVanced Manager by default is the latest that supports the patches, why would one want an older version? Nevertheless, if the user really wants to pick a different version then they can disable the check

oSumAtrIX:
I don't understand the issues mentioned here. But users can select patches, based on it, a suggested version is calculated, which is enforced as long as the switch to disable this enforcement is on.

Axelen123:
> Uhm I see. Since the user select the patches before downloading the APK, is there an "easy" way so that we can allow diffent version while still keeping the compatibility check?

The solution to that would be to check the setting for allowing incompatible patches as well.

app/src/main/java/app/revanced/manager/domain/repository/DownloadedAppRepository.kt
```diff
@@ -96,7 +98,12 @@ class DownloadedAppRepository(
             val pkgInfo =
                 pm.getPackageInfo(targetFile.toFile()) ?: error("Downloaded APK file is invalid")
             if (pkgInfo.packageName != expectedPackageName) error("Downloaded APK has the wrong package name. Expected: $expectedPackageName, Actual: ${pkgInfo.packageName}")
-            if (expectedVersion != null && pkgInfo.versionName != expectedVersion) error("Downloaded APK has the wrong version. Expected: $expectedVersion, Actual: ${pkgInfo.versionName}")
+            expectedVersion?.let {
+                if (
+                    pkgInfo.versionName != expectedVersion &&
+                    (appCompatibilityCheck || patchesCompatibilityCheck)
+                ) error("The selected app version doesn't match the suggested version. Please use the suggested version ($expectedVersion), or adjust your settings by disabling \"Require suggested app version\" and enabling \"Disable version compatibility check\".")
```
Axelen123:
```suggestion
                ) error("The selected app version ($pkgInfo.versionName) doesn't match the suggested version. Please use the suggested version ($expectedVersion), or adjust your settings by disabling \"Require suggested app version\" and enabling \"Disable version compatibility check\".")
```

Axelen123 (changes requested):

PR #2599, opened 2025-06-05
fix: Read bundle version from its manifest

TheAabedKhan:


Axelen123 (changes requested):
I don't think this should be merged currently since now the database version will never match the remote json version, which means manager will redownload the bundle every time it checks for updates. An alternative is to only use the database version for updates and use the manifest version everywhere else.

Axelen123:
Superseded by #2634

Issue #2567, opened 2025-05-24
feat(compose): Use ongoing patching app icon in notification


vippium:
### Feature description

Instead of using static android icon in notification pop-up, having an icon of app which is going to be patched. For eg: if user is patching YT, then YT icon should be there.

![Image](https://github.com/user-attachments/assets/d6f2d640-5039-4975-8e71-6d16745113f0)

### Motivation

visuals enhancements which'll help users identify the app being patched quickly.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] The feature request is only related to ReVanced Manager.

kitadai31:
In the first place, the "static android icon" is a problem that should be fixed.

But it can't be reproduced.
The broken icon is most likely because you are using an unofficial modded build of Compose Manager.
I can see the app name is "RVX Manager".

In the original Compose Manager, that icon is ReVanced Manager icon.

![Image](https://github.com/user-attachments/assets/7f65fa1d-0468-42a8-ad03-389850d2e2ca)

So the problem did not exist from the beginning.

By the way, there are several other issues with this icon.

1\. The icon doesn't look good in light mode.

![Image](https://github.com/user-attachments/assets/e284c7ea-f0e7-4e62-bb4f-9c931a0306e4)

2\. According to the developers document, this icon must not be the app icon.
So this icon should be removed.

https://developer.android.com/develop/ui/views/notifications#Templates

> Large icon: optional; usually used only for contact photos. Don't use it for your app icon.

I want to open a PR for this
I have other improvements for notification.

Also, regarding your original suggestion "Use ongoing patching app icon in notification", in my opinion, it is unnecessary.
This notification is only used to run the app in the background, and moreover, it is never displayed on Android 13 or later.
This notification just needs to be there and isn't worth the effort.

Ushie:
The icon can be removed altogether, Android already displays the app icon of the app sending the notification

vippium:
> The icon can be removed altogether, Android already displays the app icon of the app sending the notification

This is much better approach/

PR #2614, opened 2025-06-25
fix: Improve background running notification

kitadai31:
Several improvements for the notification which appears while patcher is running.

Changes:

### 1. Remove large icon 

The large icon was unnecessary.
Reason:

- According to the developers document, this icon must not be the app icon.
https://developer.android.com/develop/ui/views/notifications#Templates
- It didn't look good in light mode.
See https://github.com/ReVanced/revanced-manager/issues/2567#issuecomment-2999108092 for a image.

### 2. Use importance `IMPORTANCE_LOW` instead of `IMPORTANCE_HIGH`

`IMPORTANCE_HIGH` = Status bar icon + Sound + Heads-up banner.
`IMPORTANCE_LOW` = Status bar icon only.

Semantically, this notification is not important as it is merely necessary to run the foreground service.
So the latter is suitable for it.
This resolves the annoyance of a notification sound and Heads-up banner each time you start applying a patch.

(`IMPORTANCE_MIN` cannot be used for foreground services.)

### 3. ~~Remove unused intent~~ Fix intent

Currently, the notification responds to a tap but nothing happens.
Fixed the intent to resolve this issue.

### 4. Set category

Following best practices, set a notification category.

### 5. Improve text

Adjusted notification title and text.
Using the app title to notification title is not recommended.

Also, add a message "Tap to return to the patcher" like flutter manager.

## Screenshots

Before PR:
before

After PR:
after

For reference, the Flutter Manager's notification:
flutter

### DIFF
--- a/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
@@ -4,6 +4,7 @@ import android.app.Activity
 import android.app.Notification
 import android.app.NotificationChannel
 import android.app.NotificationManager
+import android.app.PendingIntent
 import android.content.Context
 import android.content.Intent
 import android.content.pm.ServiceInfo
@@ -15,6 +16,7 @@ import android.util.Log
 import androidx.activity.result.ActivityResult
 import androidx.work.ForegroundInfo
 import androidx.work.WorkerParameters
+import app.revanced.manager.MainActivity
 import app.revanced.manager.R
 import app.revanced.manager.data.platform.Filesystem
 import app.revanced.manager.data.room.apps.installed.InstallType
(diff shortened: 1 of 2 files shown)


brosssh:
Does the notification get automatically dismissed once the patching process is over? 
Ideally the notification should stay there until the process is done, and if user want to get back to the patching screen they can simply press the notification.

kitadai31:
> Does the notification get automatically dismissed once the patching process is over?

Yes

>Ideally ... if user want to get back to the patching screen they can simply press the notification.

Should I restore and fix the notification intent?

brosssh:
> Should I restore and fix the notification intent?

I would say yes, thank you.

Axelen123:
I agree that tapping should bring the user back to the patcher screen.

kitadai31:
Done.

I updated the PR description based on the latest commit.
Added the screenshots.
Please take a look.

Now I'm struggling with PR title.
Please update the title as needed.

Axelen123 (approved):

Issue #2553, opened 2025-05-16
bug: Manager v1.24.0 and later crash on launch with null pointer dereference


ZargEzey:
### Bug description

After updating revanced-manager to 1.24.0, it no longer launches, and instead instantly crashes. I tested the 1.25.x-dev releases and the same crash happens.

The last version that works for me is 1.23.3. 

I have attached the tombstone file of the crash, hopefully it can help.

[revanced-manager-1.24.0-tombstone.txt](https://github.com/user-attachments/files/20258599/revanced-manager-1.24.0-tombstone.txt)

### Version of ReVanced Manager and version & name of app you are patching

1.24.0

### Installation method

Regular

### ReVanced Manager logs

```shell
Cannot get these logs as app instantly crashes.
```

### Patch logs

```shell
Cannot get these logs as app instantly crashes.
```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

validcube:
Hello @ZargEzey 

The problem is currently out of our control, the problem originated within Flutter Impeller Engine (which is a renderer for the app) has crashed.

So far GPUs like Mali, Qualcomm and potentially others are affected.

I don't have the affected devices but I'll soon send you APKs for you to test with: 
* (1) Newer versions of Flutter, and 
* (2) Latest unstable version of Flutter, and
* (3) Google Skia Engine instead of Flutter Impeller Engine.

Please use ReVanced Manager v1.23.X in the meantime.

thepaleguy:
Im using the latest dev and i just patch spotify it works

martixy:
Very possibly I am experiencing the same crash.

If you need more data, I can provide logcat, but I don't know how to get the more detailed tombstone log OP gave.

I'll just revert to the older versions for now.

ZargEzey:
Hello @validcube, thank you for your efforts on this on v1.25.0

Unfortunately I test the new release and the same crash happens, flutter is still using the Impeller Engine (I see from the crash log). Despite you putting to disabled it in Android Manifest.
 
I believe the solution is talked about [here](https://github.com/flutter/flutter/issues/161070#issuecomment-2574319423)

The tag `<meta-data android:name="io.flutter.embedding.android.EnableImpeller"
[...]

PR #2631, opened 2025-07-02
fix: Disable Impeller Flutter render engine correctly to fix rendering issues

ZargEzey:
Part of commit [21ceadaf7f8e7138cf750a47142796208620cb59](https://github.com/ReVanced/revanced-manager/commit/21ceadaf7f8e7138cf750a47142796208620cb59) added option in AndroidManifest.xml to disable use of Impeller rendering backend of Flutter, to fix issues like [mine](https://github.com/ReVanced/revanced-manager/issues/2553).

I tested this in release 1.25.0 and Impeller was still used, causing same crashing as previous version. I found that the option is ineffectual when nested inside another tag, this pull request fixes it.

Move the option to disable Impeller out of the `<activity> ... </activity>` tag block. When nested inside there, the option is ineffectual.

Will close [2553](https://github.com/ReVanced/revanced-manager/issues/2553)

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -28,6 +28,9 @@
         android:largeHeap="true"
         android:requestLegacyExternalStorage="true"
         android:enableOnBackInvokedCallback="true">
+        <meta-data
+            android:name="io.flutter.embedding.android.EnableImpeller"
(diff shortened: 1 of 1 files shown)


prince863038-ux:
Youtube.com

singhsab202827-dot:
[](url)

Issue #2053, opened 2024-07-13
bug: Safeguard does not account for patch options


oSumAtrIX:
### Bug description

The warning and safeguard should not allow changing options:

![image](https://github.com/user-attachments/assets/4dd29d5b-1324-46b6-a216-864b9f94d5e2)

But it is still possible to enter and change options:

![image](https://github.com/user-attachments/assets/e1decdec-6958-48a0-be6c-41bcb477c28e)

### Version of ReVanced Manager and version & name of app you are patching

-

### Installation method

None

### ReVanced Manager logs

```shell
-
```

### Patch logs

```shell
-
```

### Acknowledgements

- [X] This issue is not a duplicate of an existing bug report.
- [X] I have chosen an appropriate title.
- [X] All requested information has been provided properly.
- [X] The bug is only related to ReVanced Manager

Axelen123:
I don't remember why the dialog mentions patch options at all. Why shouldn't the user be allowed to change patch options? Didn't you say that it should be allowed because changing patch options is intended design?

oSumAtrIX:
> Why shouldn't the user be allowed to change patch options

Lets say there's a patch which by default changes the logo. The logo option requires a path to a file. If the user now changes it to something random, the patch would fail. It is encouraged to use the default options unless explicitly confirmed with the user that they don't want that.

PR #2643, opened 2025-07-08
fix: Show selection warning also on patch option

brosssh:
Closes https://github.com/ReVanced/revanced-manager/issues/2053.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
@@ -216,7 +216,10 @@ fun PatchesSelectorScreen(
                 PatchItem(
                     patch = patch,
                     onOptionsDialog = {
-                        viewModel.optionsDialog = uid to patch
+                        if (viewModel.selectionWarningEnabled)
+                            showSelectionWarning = true
+                        else
+                            viewModel.optionsDialog = uid to patch
                     },
                     selected = compatible && viewModel.isSelected(
                         uid,


app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -216,7 +216,10 @@ fun PatchesSelectorScreen(
                 PatchItem(
                     patch = patch,
                     onOptionsDialog = {
-                        viewModel.optionsDialog = uid to patch
+                        if (viewModel.selectionWarningEnabled)
+                            showSelectionWarning = true
+                        else
+                            viewModel.optionsDialog = uid to patch
```
validcube: Is this intended? This looks off.

brosssh: Uhm should it be one line? Or with {}? Thought this was ok formatting.

validcube: > Uhm should it be one line? Or with {}? Thought this was ok formatting.

No, I'm referring to this, should this be a string?

brosssh: `to` is Kotlin operator used to create a Pair. Btw I didn't modify that, it was like that even before.

validcube: 😭 sorry, Kotlin is hard, anyway I've updated my review status.

Axelen123 (commented):
In my opinion, we shouldn't have this for options that are marked as `required` (like client id for reddit clients). Users are expected to change those options, so we shouldn't discourage them from doing that.

brosssh:
Hope I didn't overcomplicate something easy. Naming of onClick wrapper needs improvement but I'm not sure how to call it (and even if it's the best way to do this).

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -141,26 +156,37 @@ private inline fun <T : Any> WithOptionEditor(
     option: Option<T>,
     value: T?,
     noinline setValue: (T?) -> Unit,
+    selectionWarningEnabled: Boolean,
     crossinline onDismissDialog: @DisallowComposableCalls () -> Unit = {},
     block: OptionEditorScope<T>.() -> Unit
 ) {
     var showDialog by rememberSaveable { mutableStateOf(false) }
+    var showSelectionWarning by rememberSaveable { mutableStateOf(false) }
     val scope = remember(editor, option, value, setValue) {
```
Axelen123: You probably need to add `selectionWarningEnabled` to the remember keys (it's not strictly necessary in this case, but theres no reason not to do it).

app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
```diff
@@ -473,7 +472,7 @@ fun PatchesSelectorScreen(
 }
 
 @Composable
-private fun SelectionWarningDialog(
+internal fun SelectionWarningDialog(
```
Axelen123: It would be better to move this to `ui/components/patches`. `internal` does nothing since this project is not a library.

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -112,14 +120,21 @@ private interface OptionEditor<T : Any> {
...
 
     @Composable
-    fun ListItemTrailingContent(scope: OptionEditorScope<T>) {
-        IconButton(onClick = { clickAction(scope) }) {
+    fun Dialog(scope: OptionEditorScope<T>)
+
+    @Composable
+    fun ListItemTrailingContent(
+        scope: OptionEditorScope<T>,
+        onClick: (clickAction: () -> Unit) -> Unit
```
Axelen123: Handle clicks through the `OptionEditorScope` instead of adding a new parameter.

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -141,26 +156,37 @@ private inline fun <T : Any> WithOptionEditor(
...
         OptionEditorScope(
             editor,
             option,
             openDialog = { showDialog = true },
+            clickAction = { clickAction ->
+                if (!option.required && selectionWarningEnabled)
+                    showSelectionWarning = true
+                else
+                    clickAction()
+            },
```
Axelen123: `OptionEditorScope` has all the necessary information in order to make that decision. I think it would be more readable to move this logic inside `OptionEditorScope` and supply it with a callback to open the dialog instead of having a callback that accepts another callback. This is already done with `openDialog`.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -354,7 +376,9 @@ private object BooleanOptionEditor : OptionEditor<Boolean> {
 
     @Composable
     override fun ListItemTrailingContent(scope: OptionEditorScope<Boolean>) {
-        HapticSwitch(checked = scope.current, onCheckedChange = scope.setValue)
+        HapticSwitch(checked = scope.current, onCheckedChange = {
+            scope.showSelectionWarningDialogElse { scope.setValue }
+        })
```
Axelen123: Can you add newlines here to make the formatting more consistent with the rest of the project?

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -91,29 +88,45 @@ private class OptionEditorScope<T : Any>(
...
+    }
 
     @Composable
     fun ListItemTrailingContent() = editor.ListItemTrailingContent(this)
 
     @Composable
     fun Dialog() = editor.Dialog(this)
+
+    @Composable
+    fun SelectionWarningDialog(onDismissDialog: () -> Unit) = SelectionWarningDialog(onDismiss = onDismissDialog)
```
Axelen123: You can remove this function from `OptionEditorScope` and call it directly instead.

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
Axelen123: Also, I think `onCheckedChange` is broken now since it is not calling `scope.setValue`.

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -91,29 +88,45 @@ private class OptionEditorScope<T : Any>(
...
-    val setValue: (T?) -> Unit,
+    val setValue: (T?) -> Unit
 ) {
     fun submitDialog(value: T?) {
         setValue(value)
         dismissDialog()
     }
 
-    fun clickAction() = editor.clickAction(this)
+    fun showSelectionWarningDialogElse(block: () -> Unit) {
```
Axelen123: Consider renaming this to something like `checkSafeguard`

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -141,10 +154,14 @@ private inline fun <T : Any> WithOptionEditor(
     option: Option<T>,
     value: T?,
     noinline setValue: (T?) -> Unit,
+    selectionWarningEnabled: Boolean,
     crossinline onDismissDialog: @DisallowComposableCalls () -> Unit = {},
     block: OptionEditorScope<T>.() -> Unit
 ) {
     var showDialog by rememberSaveable { mutableStateOf(false) }
+    var selectionWarningEnabled by rememberSaveable { mutableStateOf(selectionWarningEnabled) }
```
Axelen123: This line is unnecessary.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
Axelen123: This is unresolved.

brosssh: I misunderstood what you meant, fixed now.

Axelen123 (approved):

PR #2663, opened 2025-07-18
build: Optimize ProGuard rules

kitadai31:
Optimized the ProGuard rules for Compose.
There were unnecessary or excessive or obsoleted rules.
The rules for kotlinx-serialization are removed since the same rules are bundled into the library starting from [v1.5.0-RC](https://github.com/Kotlin/kotlinx.serialization/releases/tag/v1.5.0-RC).
As with other ReVanced projects (Flutter Manager and Extensions), I did not add any comments.

One thing that needs to be discussed is whether to enable obfuscation.
If I remove `-dontobfuscate`, the APK size is reduced by 160 KB.
The total size of the unzipped classes.dex is reduced by 722 KB.

Since this app is open source, there is obviously no need for obfuscation.
But obfuscation would still optimize the app like runtime memory.
I have tested that all functions of the app work even with obfuscation.
However, the possibility still exists that unforeseen problems may arise in the future.
Also, some Exception names like `brut.common.BrutException` will be obfuscated so it will be hard to understand. (This may be addressed with proguard rules like `-keep class extends Exception`.)
I don't think the 160 KB of reduction is worth the obfuscation.

If you enable obfuscation, these 2 rules are needed additionally.
```
-keepnames class com.google.common.collect.*
-keepnames class android.util.** { *; }
```

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -1,57 +1,12 @@
-# Add project specific ProGuard rules here.
-# You can control the set of applied configuration files using the
-# proguardFiles setting in build.gradle.kts.kts.
-#
-# For more details, see
-#   http://developer.android.com/guide/developing/tools/proguard.html
-
 -dontobfuscate
 
-# Required for serialization to work properly
--if @kotlinx.serialization.Serializable class **
--keepclassmembers class <1> {
-    static <1>$Companion Companion;
-}
--if @kotlinx.serialization.Serializable class ** {
-    static **$* *;
-}
--keepclassmembers class <2>$<3> {
-    kotlinx.serialization.KSerializer serializer(...);
-}
--if @kotlinx.serialization.Serializable class ** {
-    public static ** INSTANCE;
-}
--keepclassmembers class <1> {
-    public static <1> INSTANCE;
-    kotlinx.serialization.KSerializer serializer(...);
-}
-
-# This required for the process runtime.
--keep class app.revanced.manager.patcher.runtime.process.* {
-  *;
-}
-# Required for the patcher to function correctly
--keep class app.revanced.patcher.** {
-  *;
-}
--keep class brut.** {
-  *;
-}
--keep class org.xmlpull.** {
-  *;
-}
--keep class kotlin.** {
-  *;
-}
--keep class org.jf.** {
-  *;
-}
--keep class com.android.** {
-  *;
-}
--keep class app.revanced.manager.plugin.** {
-  *;
-}
+-keep class app.revanced.patcher.** { *; }
+-keep class kotlin.** { *; }
+-keep class com.android.tools.smali.** { *; }
+-keep class app.revanced.manager.patcher.runtime.process.* { *; }
+-keep class app.revanced.manager.plugin.** { *; }
+-keepnames class com.android.apksig.internal.** { *; }
+-keepnames class org.xmlpull.** { *; }
 
 -dontwarn com.google.auto.value.**
 -dontwarn java.awt.**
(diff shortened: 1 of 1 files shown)


app/proguard-rules.pro
```diff
@@ -1,63 +1,16 @@
...
-  *;
-}
--keep class app.revanced.manager.plugin.** {
-  *;
-}
+-keep class app.revanced.patcher.** { *; }
+-keep class kotlin.** { *; }
+-keep class com.android.tools.smali.** { *; }
+-keep class app.revanced.manager.patcher.runtime.process.* { *; }
+-keep class app.revanced.manager.plugin.** { *; }
```
validcube: This is nitpick, in my opinion, I would prefer to sort revanced class at the top of the list.

```suggestion
-keep class app.revanced.patcher.** { *; }
-keep class app.revanced.manager.patcher.runtime.process.* { *; }
-keep class app.revanced.manager.plugin.** { *; }
-keep class kotlin.** { *; }
-keep class com.android.tools.smali.** { *; }
```

kitadai31:
I reflected the suggestion.
But I put the Manager's own class at the top because I thought that was more natural.

Also changed:

- Sorted `org.kotlin` alphabetical
- Removed `-dontwarn it.skrape.fetcher.*` (The dependency has been removed in last commit)
- Removed `-dontwarn com.google.auto.value.**` (Build succeeds without this rule)

app/proguard-rules.pro
```diff
@@ -1,63 +1,14 @@
...
-}
--keep class com.android.** {
-  *;
-}
--keep class app.revanced.manager.plugin.** {
-  *;
-}
+-keep class app.revanced.manager.patcher.runtime.process.* { *; }
+-keep class app.revanced.manager.plugin.** { *; }
+-keep class app.revanced.patcher.** { *; }
```
oSumAtrIX, line 5: might as well just
```suggestion
-keep class app.revanced.** { *; }
```

kitadai31: That is not good.
All of the Manager app's code is located in `app.revanced.manager`, so doing so would result in keeping the entire app code.

Source: https://developer.android.com/topic/performance/app-optimization/adopt-optimizations-incrementally?hl=en#:~:text=Note%3A%20Never,gradually%20over%20time.

> Note: Never write a keep rule that keeps all of your app, for example -keep ** or -keep com.myapp.**. These rules prevent optimization across such a wide portion of the app that you'll pay the build cost of R8 with effectively none of the benefit. Instead, always start with library- or package- based keep rules, and try to remove them gradually over time.

oSumAtrIX: I don't think we have anything of the app that we don't use apart from maybe one or two library Apis. Instead we just want to optimize other packages we use that transitively depend on many others.

kitadai31: That's partially true.
Your app has no unused classes or members except for one or two library apis, so code shrinking a.k.a. tree shaking will have no effect.

However, ProGuard does not only tree shaking, but also optimization.

The optimization rewrites code, for example inlining.
This may improve the performance of the app.
And `-keep` disables any optimizations, so it should not be used for the entire codes.
https://developer.android.com/topic/performance/app-optimization/add-keep-rules

oSumAtrIX: In this case enable this feature for all classes while globally ignoring shrinking or tree shaking for app.revanced

validcube: > In this case enable this feature for all classes while globally ignoring shrinking or tree shaking for app.revanced

I don't think it is possible to do that, you can't selectively choose which class to optimise or shrink in proguard.

oSumAtrIX: This will be problematic in the future. We will add a new package, proguard will remove it and we'll spend time to debug what the issue is

Issue #2521, opened 2025-05-05
feat: Move downloader API repository as a module here


oSumAtrIX:
### Feature description

Right now it is an independent project at https://github.com/ReVanced/revanced-manager-downloader-api. It should be moved into this repository as a Gradle module.

### Motivation

The reason it is a separate repository was because it was easier to version independently. However, https://github.com/dhoulb/multi-semantic-release can achieve this in a single repository.

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] The feature request is only related to ReVanced Manager.

Axelen123:
I don't think this issue blocks release, so I have removed it from the release milestone. If anyone has a motivation for why this needs to be done before release, feel free to share it.

Axelen123:
> When packages are released moving it to this repository might create breaking changes such as the package name in the maven registry.

I noticed earlier when I moved the package to the other repo that you can keep using the same package name, so no breaking changes are necessary.

> If that's not the case, then the other reason would be that moving shouldn't take more than a couple minutes. I could look into this myself in a couple of minutes

I am not familiar with semantic release, but if you know how to do this then go head 👍

Axelen123:
This was completed months ago.

PR #2675, opened 2025-07-22
ci: Adjust to release `app` and `api` module respectively

brosssh:
Updated CI flows as accorded.

What will happens when this is merged to `dev`?

For the `app`:
- It will fetch the latest tag (with format `vX.Y.Z`)
- Calculate the next version
- Create a new tag and release the APK with attestation
- Update `gradle.properties`
- Create `CHANGELOG.md`

For the `api`:
- It will fetch the latest tag (with format `api@X.Y.Z`)
- Since it won't find it, it will start from `api@1.0.0-dev.1`
- Create the new tag
- Publish the package to Maven
- Update `gradle.properties` to `api@1.0.0-dev.1`
- Create `CHANGELOG.md`


gradle/libs.versions.toml
```diff
@@ -78,9 +79,10 @@ room-runtime = { group = "androidx.room", name = "room-runtime", version.ref = "
 room-ktx = { group = "androidx.room", name = "room-ktx", version.ref = "room-version" }
 room-compiler = { group = "androidx.room", name = "room-compiler", version.ref = "room-version" }
 
-# Patcher
+# ReVanced
 revanced-patcher = { group = "app.revanced", name = "revanced-patcher", version.ref = "revanced-patcher" }
 revanced-library = { group = "app.revanced", name = "revanced-library", version.ref = "revanced-library" }
+revanced-manager-api = { group = "app.revanced", name = "revanced-manager-downloader-api", version.ref = "revanced-manager-api" }
```
brosssh: This is temporary. Once the API is released for the first time via the CI flow in this PR, `name` needs to be adjusted to `revanced-manager-api`. After that, the [old package](https://github.com/ReVanced/revanced-manager-downloader-api/packages/2485751) could be deleted to avoid confusion.

app/build.gradle.kts
```diff
@@ -60,7 +60,7 @@ dependencies {
     implementation(libs.revanced.library)
 
     // Downloader plugins
-    implementation(project(":api"))
+    implementation(libs.revanced.manager.api)
```
Axelen123: What is the reason for this change? Using `project` makes it slightly easier for developers working on the API.

gradle/libs.versions.toml
Axelen123: If you want to change where the package is published you'll have to update the existing downloaders and template as well. Is there a reason why we are changing it?

Axelen123 (changes requested):

app/build.gradle.kts
oSumAtrIX: It should definitely rely on the local Gradle project as this module is a part of manager and only a separate module to separate the interface from manager for other downloaders

brosssh: I did this because of https://github.com/ReVanced/revanced-manager/issues/2521. Did I misunderstand the pr?

Axelen123: I think you misunderstood the issue. That issue was about moving the downloader API to here (previously it was a separate repo). oSumAtrIX did that a while ago, so I closed the issue.

brosssh: Ok, I reverted the commit.

app/.releaserc
validcube, line 1: The file currently uses lenient JSON, try to format them in regular JSON.

api/.releaserc
validcube, line 1: The file currently uses lenient JSON, try to format them in regular JSON.

.github/workflows/release.yml
validcube, line 1: While we at it, update action dependency here.

api/build.gradle.kts
```diff
@@ -19,43 +17,6 @@ dependencies {
...
-val lastTag = "git describe --tags --abbrev=0".runCommand()
-val hasChangesInThisModule = "git diff --name-only $lastTag..HEAD".runCommand().lineSequence().any {
-    it.startsWith(projectPath)
-}
-
-tasks.matching { it.name.startsWith("publish") }.configureEach {
-    onlyIf {
-        hasChangesInThisModule
-    }
-}
```
validcube, line 57: Without this code we can use Gradle Configuration Cache again.

.github/workflows/release.yml
```diff
@@ -55,18 +51,38 @@ jobs:
...
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set latest App tag (before the new release)
+        id: before
+        run: |
+          git fetch --tags
+          TAG=$(git describe --tags --abbrev=0 --match 'v[0-9]*' $(git rev-parse HEAD))
+          echo "LATEST_TAG=$TAG" >> $GITHUB_OUTPUT
```
validcube: I'm not sure what the point of using two steps for setting app tags are, if I have to guess this is a workaround, This should be documented.

.github/workflows/release.yml
```diff
@@ -55,18 +51,38 @@ jobs:
...
           KEYSTORE_PASSWORD: ${{ secrets.KEYSTORE_PASSWORD }}
           KEYSTORE_ENTRY_ALIAS: ${{ secrets.KEYSTORE_ENTRY_ALIAS }}
           KEYSTORE_ENTRY_PASSWORD: ${{ secrets.KEYSTORE_ENTRY_PASSWORD }}
 
+      - name: Set latest App tag (after the new release)
+        id: after
+        run: |
+          git fetch --tags
+          TAG=$(git describe --tags --abbrev=0 --match 'v[0-9]*' $(git rev-parse HEAD))
+          echo "LATEST_TAG=$TAG" >> $GITHUB_OUTPUT
```
validcube: I'm not sure what the point of using two steps for setting app tags are, if I have to guess this is a workaround, This should be documented.

gradle.properties
```diff
@@ -6,23 +6,22 @@
...
-android.nonTransitiveRClass=true
-android.nonFinalResIds=false
+android.nonTransitiveRClass = true
+android.nonFinalResIds = false
 # Task :app:assembleReleaseSignApk fails if this is set to true.
-org.gradle.configuration-cache=false
-org.gradle.caching=true
-version=1.25.0-dev.1
\ No newline at end of file
+org.gradle.configuration-cache = false
```
validcube: Without the code for accessing the file in the build script, we can use Gradle Configuration Cache again.

```suggestion
org.gradle.configuration-cache = true
```

package.json
```diff
@@ -1,10 +1,11 @@
...
+  ],
   "devDependencies": {
-    "@anolilab/multi-semantic-release": "^1.1.10",
-    "gradle-semantic-release-plugin": "^1.10.1",
-    "@saithodev/semantic-release-backmerge": "^4.0.1",
-    "@semantic-release/changelog": "^6.0.3",
-    "@semantic-release/exec": "^6.0.3",
-    "@semantic-release/git": "^10.0.1"
+    "@anolilab/multi-semantic-release": "^1.1.12",
+    "semantic-release": "^22.0.12"
```
validcube: Can we bump semrel to the latest version of 24.x?

Although I think some of the plugins don't support it so the safest upgrade would be 22.0 -> 23.0 since that's just bumping the required node to 20

validcube (approved):
💫🏗️ Infra looks fine to me*

* I can't test the infra CI at the moment, but if your testing works then I have no objections.

.github/workflows/release.yml
```diff
@@ -55,18 +51,38 @@ jobs:
         run: |
           echo "${{ secrets.KEYSTORE }}" | base64 --decode > "app/keystore.jks"
 
-      - name: Semantic Release
-        uses: cycjimmy/semantic-release-action@v4
-        id: semantic
+      - name: Semantic Release (API)
```
oSumAtrIX: use the same name as in other repos. Workflows should be consistent across all repos

.github/workflows/release.yml
oSumAtrIX: This should be clarified

.github/workflows/release.yml
```diff
@@ -55,18 +51,38 @@ jobs:
...
+        run: npx multi-semantic-release --tag-format 'api@${version}' --ignore-packages app
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set latest App tag (before the new release)
```
oSumAtrIX: Why are we setting the app tag? What is "app tag". Is this to skip attestation? If so, multi-semantic release should implement an output for this and i think it wraps semantic version which does have such outputs (perhaps check process error for 1)

api/.releaserc
```diff
@@ -8,34 +8,35 @@
...
     [
-      "@semantic-release/git",
+      "@semantic-release/exec",
       {
-        "assets": [
-          "CHANGELOG.md",
-          "gradle.properties",
-        ],
-        "message": "chore: Release v${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
+        "prepareCmd": "sed -i 's/^version =.*/version = ${nextRelease.version}/' gradle.properties"
```
oSumAtrIX: This wont work if there is no initial version committed beforehand. However no initial version should be committed beforehand because this file is soly touched by workflows.

api/.releaserc
oSumAtrIX, line 1: I have trouble understanding what happens here. Why is so much manual build scripting done here?

.github/workflows/release.yml
Axelen123: Different requirements (the need for two steps in this case) from other repos would justify using a different name, unless there is a good way to only use one step.

.github/workflows/release.yml
Axelen123: Waiting for multi-sem-rel to implement something is undesirable here.

api/.releaserc
Axelen123: Bootstrapping the version is warranted and probably necessary if we want to continue using flutter versioning. Please mark this as resolved or suggest a solution.

oSumAtrIX: The build system should not be designed around this cause. It's implying there is a version already present which isn't the case other than this specific case. Also, business logic like this looks wrong in this place. The solution is to not try to squeeze business logic in this place and instead in the correct place as seen by gradle semantics release which checks if the version is there, otherwise adds it

.github/workflows/release.yml
oSumAtrIX: What I meant with match is to match the convention we use across repos.

```suggestion
      - name: Release API
```

later below

```suggestion
      - name: Release
```

.github/workflows/release.yml
oSumAtrIX: In this case, open and attach an issue as a comment given this project depends on its behavior.

.github/workflows/release.yml
```diff
@@ -55,18 +51,38 @@ jobs:
...
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           KEYSTORE_PASSWORD: ${{ secrets.KEYSTORE_PASSWORD }}
           KEYSTORE_ENTRY_ALIAS: ${{ secrets.KEYSTORE_ENTRY_ALIAS }}
           KEYSTORE_ENTRY_PASSWORD: ${{ secrets.KEYSTORE_ENTRY_PASSWORD }}
 
+      - name: Set latest App tag (after the new release)
+        id: after
+        run: |
+          git fetch --tags
+          TAG=$(git describe --tags --abbrev=0 --match 'v[0-9]*' $(git rev-parse HEAD))
```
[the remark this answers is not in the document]
brosssh: This is wrong.

.github/workflows/release.yml
```diff
@@ -55,18 +51,38 @@ jobs:
...
+        run: npx multi-semantic-release --tag-format 'api@${version}' --ignore-packages app
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set latest App tag (before the new release)
+        id: before
+        run: |
+          git fetch --tags
+          TAG=$(git describe --tags --abbrev=0 --match 'v[0-9]*' $(git rev-parse HEAD))
```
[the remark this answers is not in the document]
brosssh: This is wrong.

api/.releaserc
brosssh: Where should the business logic be placed then?

brosssh: Also I don't agree that we should add the logic to handle the case where the version doesn't exist. This can ONLY ever happen on this PR. I'd rather hardcode a bogus version that will be overwritten as soon as the pr is merged instead of adding additional logic to handle a case that is impossible to happen in the future.

@Axelen123 what do you think?

api/.releaserc
brosssh: What manual scripting? The sed is needed to replace the version in the `gradle.properties` file. This was not needed before as we were using the `gradle-semantic-release` plugin, while now we need custom logic to handle API/app release separately.

package.json
brosssh: I'm not sure why I set this semantic release version. I think the version should be consistent with the rest of the repos, so `^24.2.7`. Do you know what might not be compatible with this?

api/.releaserc
Axelen123: I agree with brosssh here. CI is annoying to test, so it is better to keep it as simple as possible. Version bootstrapping is not an unusual practice (Cargo does it for example).

oSumAtrIX: > Where should the business logic be placed then?

Well I said it here:

> in the correct place as seen by gradle semantics release 

> This can ONLY ever happen on this PR. 

Its still a quirk, and its still business logic that happens randomly in a file other than the expected places.

>  instead of adding additional logic to handle a case that is impossible to happen in the future.

The you breach the problem space. You design the system, not independently of ReVanced. The system is now designed in mind with a specific configuration of the repo in ReVanced and wouldn't work independently of this configuration.

>  CI is annoying to test

You do not need to do an e2e test (which btw is very easy to do with act locally), unit testing is enough.

---

Please see gradle semantic release on where and how to put this logic

api/.releaserc
oSumAtrIX: > What manual scripting

The sed. "updating the version" is a build process script.

> The sed is needed to replace the version in the gradle.properties file.

In other repositories we don't need sed. Why is it used here.

> while now we need custom logic to handle API/app release separately.

Why not just adjust it to handle API/app release separately then? Sounds like the correct flow of logic here, instead of reinventing the wheel on a boat - somewhere it doesn't belong.

brosssh: > In other repositories we don't need sed. Why is it used here.

Because we don't use multi semantic release in other repos.

With multi semantic release, gradle semantic release become incompatible:
Screenshot_2025-08-21-12-14-05-26_1b5a96d325b4aa38841ed4ab46124a4f
https://discord.com/channels/952946952348270622/952987428786941952/1394968781645811712

> Why not just adjust it to handle API/app release separately then?

I don't understand, adjust what and how?

oSumAtrIX: > With multi semantic release, gradle semantic release become incompatible:

In this case, fork and adjust so it works on multi semantic release. I don't understand why you put core build logic (version updating) in releaserc. with sed. Thats not only brittle but also the wrong place. releaserc is a configuration file, not a "build script"

package.json
validcube: @brosssh Actually, I just checked the change log, doesn't seem to include any major breaking changes that's relevant to us. You should be able to upgrade to v24/v25 beta just fine assuming dependency support it.

api/.releaserc
Axelen123: It should be valid JSON now.

Axelen123: It uses upstream gradle-semantic-release-plugin now. I'm marking this as resolved.

.github/workflows/release.yml
oSumAtrIX: You dont need those two steps to get the tag. See https://chatgpt.com/share/68a8daf9-7d54-8011-99f6-8632c029980f.

    TAG=$(npx multi-semantic-release \
      --tag-format 'v${version}' \
      --ignore-packages api \
      --json | jq -r '.[] | select(.name == "app") | .tag')
    echo "APP_TAG=$TAG" >> $GITHUB_OUTPUT

and then later

- name: Attest
  if: steps.release.outputs.APP_TAG != ''

.github/workflows/release.yml
oSumAtrIX: See https://github.com/ReVanced/revanced-manager/pull/2675/files#r2294743348

.github/workflows/release.yml
```diff
@@ -55,18 +51,38 @@ jobs:
         run: |
           echo "${{ secrets.KEYSTORE }}" | base64 --decode > "app/keystore.jks"
 
-      - name: Semantic Release
-        uses: cycjimmy/semantic-release-action@v4
-        id: semantic
+      - name: Release API
+        run: npx multi-semantic-release --tag-format 'api@${version}' --ignore-packages app
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
```
oSumAtrIX, line 63: What is this needed for again?

app/gradle.properties
```diff
@@ -0,0 +1 @@
+
```
oSumAtrIX: The file should be removed, right now its committed as an empty file

api/gradle.properties
```diff
@@ -0,0 +1 @@
+
```
oSumAtrIX: The file should be removed, right now its committed as an empty file

app/gradlew
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh
+
+# Get directory of current script.
+script_dir=$(CDPATH= cd -- "$(dirname -- "$0")" && pwd)
+
+exec "$script_dir"/../gradlew $@
```
oSumAtrIX: this file is "autogenerated"/pulled from gradle. If you run ./gradlew wrapper init to update it according to the gradle version, these changes will be overwritten.

api/gradlew
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+## This script exists because gradle-semantic-release-plugin expects gradlew to be in the directory it was executed in.
```
oSumAtrIX: Do not commit to this file, it should be autogenerated by gradle

app/gradlew
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+## This script exists because gradle-semantic-release-plugin expects gradlew to be in the directory it was executed in.
```
oSumAtrIX: Do not commit to this file, it should be autogenerated by gradle

api/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "app.revanced.revanced-manager-api",
```
oSumAtrIX: Is name supposed to be the entire package name? Where is "name" used in CI?

app/.releaserc
```diff
@@ -22,27 +23,27 @@
       {
         "assets": [
           "CHANGELOG.md",
-          "gradle.properties",
+          "gradle.properties"
         ],
-        "message": "chore: Release v${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
+        "message": "chore: Release App v${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
```
oSumAtrIX:
```suggestion
        "message": "chore: Release v${nextRelease.version} [skip ci]\n\n${nextRelease.notes}"
```

just this is fine

app/gradle.properties
Axelen123: I can test if it works without the file being present, but I am not going to remove it if it doesn't.

app/gradlew
Axelen123: This file is not autogenerated and does not need to be maintained.

api/gradlew
Axelen123: It is not autogenerated.

app/gradlew
Axelen123: Same here.

api/package.json
Axelen123: Someone who knows more about semantic release can answer this.

app/gradle.properties
oSumAtrIX: gradle semantic release generates the file if it doesn't exist, which is correct behaviour because the props file is not something a gradle project depends on

app/gradlew
oSumAtrIX: From what I am understanding you are trying to call gradlew from parent. Simply symlink to it?

.github/workflows/release.yml
brosssh: What's this?

api/package.json
brosssh: The default generate tag for `multi-semantic-release` is `$name@$version`. However, since we are specifying the `tagFormat` this is not the case. It's possible the name is unused, I'm not sure if it can be removed.

.github/workflows/release.yml
brosssh: Was this tested already?

.github/workflows/release.yml
Axelen123: > What is this needed for again?

Github username.

.github/workflows/release.yml
Axelen123: > You dont need those two steps to get the tag. See https://chatgpt.com/share/68a8daf9-7d54-8011-99f6-8632c029980f.
> 
> ```
> TAG=$(npx multi-semantic-release \
>   --tag-format 'v${version}' \
>   --ignore-packages api \
>   --json | jq -r '.[] | select(.name == "app") | .tag')
> echo "APP_TAG=$TAG" >> $GITHUB_OUTPUT
> ```
> 
> and then later
> 
>     * name: Attest
>       if: steps.release.outputs.APP_TAG != ''

This is AI slop. `multi-semantic-release` has no `--json` option.

.github/workflows/release.yml
oSumAtrIX: I know that GitHub actor is the username. What is this needed for.

Axelen123: You need credentials because Github Packages requires it for downloading (even though it shouldn't) and publishing.

api/build.gradle.kts
Axelen123: The publish step in `app` is incompatible with Gradle Config Cache. Should this code be readded?

api/package.json
brosssh: I tested and the name is required by sem rel bot. However it shouldn't be used, so it could be anything I guess.

.github/workflows/release.yml
oSumAtrIX: You can grep the tag from the output of semantic release one way or another

api/build.gradle.kts
oSumAtrIX: This is a bad hack and should be avoided. It depends on git and build should depend on git. Why do we need this code? Why is app incomaptible with gradle config cache? Is caching even useful in ci? Do we need caching in local builds?

api/package.json
oSumAtrIX: Arbitrarily, names should default to "ReVanced Manager API" in this case (unless there's a syntactical convention somewhere), also for the app package json it should say "ReVanced Manager"

package.json
```diff
@@ -1,10 +1,11 @@
 {
+  "private": true,
+  "workspaces": [
+    "api",
+    "app"
+  ],
```
oSumAtrIX, line 6: Whats this for?

gradle.properties
```diff
@@ -6,23 +6,22 @@
...
+org.gradle.jvmargs = -Xmx2048m -Dfile.encoding=UTF-8
 # When configured, Gradle will run in incubating parallel mode.
 # This option should only be used with decoupled projects. More details, visit
 # http://www.gradle.org/docs/current/userguide/multi_project_builds.html#sec:decoupled_projects
 # org.gradle.parallel=true
 # AndroidX package structure to make it clearer which packages are bundled with the
 # Android operating system, and which are packaged with your app's APK
 # https://developer.android.com/topic/libraries/support-library/androidx-rn
-android.useAndroidX=true
+android.useAndroidX = true
```
oSumAtrIX: why was this changed?

gradle.properties
```diff
@@ -6,23 +6,22 @@
 # http://www.gradle.org/docs/current/userguide/build_environment.html
 # Specifies the JVM arguments used for the daemon process.
 # The setting is particularly useful for tweaking memory settings.
```
oSumAtrIX: Should comments be removed?

package.json
Axelen123: Its just NPM workspaces config for monorepos. multi sem rel needs it.

gradle.properties
Axelen123: I don't see a reason to remove them.

gradle.properties
Axelen123: Formatting by gradle probably. The file did have a version property, which had to be removed.

gradle.properties
oSumAtrIX: Two reasons:

- Inconsistent with other repos
- Inconsistent with itself, as some params are and some arent commented.

oSumAtrIX: Another would be that some of the comments are part of the "template". E.g one talks abut parallel, which isn't relevant here

gradle.properties
oSumAtrIX: I am not sure i understand. Gradle does not have a formatted. From what I know the default formatting is no space: https://github.com/Project-ARTist/ArtistGui/blob/master/gradle.properties.template

Axelen123: Then it was done by an editor. It does not appear to have any functional impact, so go fix it yourself if it bothers you.

gradle.properties
Axelen123: That is not within the scope of this PR. The only thing that was needed was to remove the version property.

api/build.gradle.kts
Axelen123: I won't readd it then, nor will I try to figure out why it is incompatible with config cache, since its not within the scope of this PR.

gradle.properties
oSumAtrIX: This must be reverted then. It is not part of the PR, nor should it be done on its own.

gradle.properties
oSumAtrIX: The PR still touches the file close enough to do this simple refactor, i did it now

api/build.gradle.kts
oSumAtrIX: Neither has to be done. The cache is enabled by default, it was incompatible with the removed code. I am asking why you considered readding this code

gradle.properties
brosssh: Resolved by https://github.com/ReVanced/revanced-manager/pull/2675/commits/6168344495831fd56737083c052f7fdd07da38ac.

.github/workflows/release.yml
brosssh: The 2 steps are not there anymore, marking as resolved.

.github/workflows/release.yml
brosssh: @Axelen123 I adjusted your solution on my fork and it's working. Files are correctly attested (https://github.com/brosssh/revanced-manager/actions/runs/17200993329).

api/.releaserc
brosssh: Marking as resolved since we are back using gradle semantic release plugin (which can handle empty `gradle.properties` file).

api/package.json
brosssh: Since what is being set is the npm package name, I believe it should follow the normal npm packages naming convention. Maybe [this](https://github.com/npm/validate-npm-package-name#naming-rules) can be followed.

I'd suggest simply `revanced-manager` and `revanced-manager-api`, or to keep the current naming.

.github/workflows/release.yml
brosssh: 2 steps are not there anymore.

.github/workflows/release.yml
brosssh: They are all up to date with https://github.com/ReVanced/revanced-patches/blob/main/.github/workflows/release.yml.

api/package.json
Axelen123: `revanced-manager` and `revanced-manager-api` or `app` and `api` would follow existing npm conventions more closely.

api/build.gradle.kts
Axelen123: I assumed it was removed in order to enable config cache, but that didn't work anyway, so I asked if it should be re-added.

api/package.json
oSumAtrIX: app and api is fine

Axelen123 (approved):

.github/workflows/release.yml
```diff
@@ -55,18 +57,32 @@ jobs:
...
+        run: npx multi-semantic-release --tag-format 'api@${version}' --ignore-packages app
         env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Release
+        id: release
+        run: |
+          ./gradlew :app:writeVersionEnv
+          echo "OLD_TAG=$(cat app/build/outputs/version.env)" >> $GITHUB_OUTPUT
```
oSumAtrIX: See 

```
> gradlew properties | grep "version:" | cut -d " " -f2
1.25.0-dev.1
```

No custom task, besides you could've just read gradle props directly, so:

```
 cat gradle.properties | grep version | cut -d "=" -f2
```

.github/workflows/release.yml
```diff
@@ -28,10 +29,11 @@ jobs:
           java-version: '17'
 
       - name: Cache Gradle
-        uses: burrunan/gradle-cache-action@v1
+        uses: burrunan/gradle-cache-action@v3
```
brosssh, line 32: `gradle-cache-action@v3` actually breaks msr. I'll double check if it's possible to use `v2`. `v1` is working for sure.

brosssh: Nevermind, now for some reason `v3` seems to be working.

brosssh:
Tested everything on my fork and LGTM.

.github/workflows/release.yml
```diff
@@ -55,18 +57,27 @@ jobs:
...
+        run: npx multi-semantic-release --tag-format 'api@${version}' --ignore-packages app
         env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Release
+        id: release
+        run: |
+          npx multi-semantic-release --tag-format 'v${version}' --ignore-packages api | tee result.log
+          echo "new_release_published=$(grep -q 'Published release' result.log && echo true || echo false)" >> $GITHUB_OUTPUT
```
oSumAtrIX: Why are we writing to a result.log file instead of just directly piping into grep and echoing?

.github/workflows/release.yml
```diff
@@ -55,18 +57,27 @@ jobs:
...
           KEYSTORE_ENTRY_ALIAS: ${{ secrets.KEYSTORE_ENTRY_ALIAS }}
           KEYSTORE_ENTRY_PASSWORD: ${{ secrets.KEYSTORE_ENTRY_PASSWORD }}
 
       - name: Attest
-        if: steps.semantic.outputs.new_release_published == 'true'
+        if: steps.release.outputs.new_release_published == 'true'
         uses: actions/attest-build-provenance@v2
         with:
-          subject-name: 'ReVanced Manager ${{ steps.release.outputs.new_release_git_tag }}'
+          subject-name: 'ReVanced Manager ${{ steps.release.outputs.NEW_TAG }}'
```
oSumAtrIX, line 81: Where does this output come from?

app/.releaserc
```diff
@@ -32,17 +33,17 @@
       {
         "assets": [
           {
-            "path": "app/build/outputs/apk/release/revanced-manager*.apk?(.asc)"
-          },
+            "path": "build/outputs/apk/release/revanced-manager*.apk?(.asc)"
+          }
         ],
-        successComment: false
+        "successCommentCondition": "<% return false; %>"
```
oSumAtrIX: Why was this changed?

app/build.gradle.kts
```diff
@@ -13,6 +14,33 @@ plugins {
...
+    if (!process.waitFor(10, TimeUnit.SECONDS)) {
+        process.destroy()
+        throw IOException("Command timed out: $this")
+    }
+
+    thread.join()
+    return output.toString().trim()
+}
+
+val tagCount = "git tag --list v*".runCommand().lineSequence().count()
```
oSumAtrIX: Instead of doing this, do versionCode = currentVersionCode+1;
Is there a way to determine it?

.github/workflows/release.yml
brosssh: This way, the output of the step is still printed in the GitHub action. If I were to pipe it directly we would lose this information.

oSumAtrIX: Use tee >(grep ...)

app/.releaserc
Axelen123: Because `successComment` is deprecated.

app/build.gradle.kts
Axelen123: This has been suggested to me before, but no one has explained how to implement it. The gradle semantic release plugin is not specific to Android, so it has no concept of a version code (only version) and cannot bump them. I don't know of any good way to determine the version code of the previous version.

app/.releaserc
oSumAtrIX: Hmm, I'd say keep the deprecated one for consistency reasons on our other repos. We'll refactor this in one sweep on all repos that need the change.

app/build.gradle.kts
oSumAtrIX: So right now build system incorporates release CI code. This is not good. The bumping part belongs to the CI not the build system. For this reason the bumping part should be moved out of the build system. This can be done by a success hook in the releaserc file after Sem Release Gradle bumps the version. Basically if you remember what we did for version with releaserc, do that with versionCode. 

1. Create a property in Gradle props
2. Set versionCode to that property 
3. In releaserc. bump version code by one

You need to create the file/property if it does not exist or handle the initial state. Chatgpt probably will be able to do all this. Open to other ideas

.github/workflows/release.yml
brosssh: This is actually a mistake. There is nothing that generated this right now.

In the previous version (https://github.com/ReVanced/revanced-manager/commit/786cfc49d6b338e5f8e54c097ddb17a429ea57ef), it was manually generated. In the other repos, it is generated by the semantic release action. However with msr, as already said, there isn't a built in way to extract the newest app tag (afaik).

brosssh: I think there are 2 possibilites:
- Rolling back to the previous version (https://github.com/ReVanced/revanced-manager/commit/786cfc49d6b338e5f8e54c097ddb17a429ea57ef). There, the app version is already obtained (by grepping the `gradle.properties` file) since it's used to detect if a new app version was released
- Keeping the current version but also grepping in msr logs the generated tag

@oSumAtrIX @Axelen123 wdty?

app/build.gradle.kts
Axelen123: Like I said, version (not code) is handled by the gradle semantic release plugin. Bumping the version code is possible with the `exec` plugin and a script, but you have previously rejected the use of that plugin, so I am not sure what to do.

.github/workflows/release.yml
oSumAtrIX: I think we can extract the version from the log that we already grep and output into $GITHUB_OUTPUT.

Then you'd check if "NEW_TAG" is set to run attest step, and if so, read NEW_TAG for the subject name.

app/build.gradle.kts
oSumAtrIX: The rejection was on the basis that there was already a solution which didnt need the releaserc. file and exec plugin to bump the version. Technically speaking a plugin must be made, just like it was made for gradle, but also more specific for Android, as we don't just need to bump the version for Gradle projects, but also the versionCode. Luckily this is as simple as changing https://github.com/KengoTODA/gradle-semantic-release-plugin/blob/master/src/prepare.ts#L23-L25 so that it also bumps versionCode by upping by one.

Axelen123: Maintaining a plugin or fork is cumbersome and not ideal. Another viable alternative that does not involve tags is by bumping with a gradle task. You can't deny that using tags or a task is a simpler solution for us than maintaining a plugin regardless of whether or not you think those two solutions are correct. Version codes are purely used for comparing versions.
This means the ONLY requirements for correctness here regarding the method for determining version codes are:
- Higher version code => newer version
- Stay away from the integer limits
- Different versions should have different version codes

No solution proposed by either of us violates these rules. You said earlier that bumping belongs to CI, but this is still done by CI, since you bump by creating a new tag. If you want to justify maintaining a plugin, do it by providing an example or scenario where the tag method breaks or is problematic in some way. The problems/breakage also need to be significant enough to justify the fork.

oSumAtrIX: > Maintaining a plugin or fork is cumbersome and not ideal

I disagree. The change is minimal, upstream can be merged.

> bumping with a gradle task

This must be done independently of the app build file. I am not sure how you think, pulling gradle wrapper and a separate gradle script from scratch, that is semantically and structurally separate from the code that is bumping the version, is worth it over the already existing solution.

>  do it by providing an example or scenario where the tag method breaks or is problematic in some way

The issue is that you put CI code in the build system files of manager. Wrong place.
Exec would "correct" the place a little, however as explained earlier:

> Technically speaking a plugin must be made.

If it is not a problem, we can derive code from name:
From least to most significant digits of versioncode:

If the current version is not a prerelease, set the first digit to the verson. Set the next digit to 0.
If the current version is not a prerelease, set the first digit to 0 and the next digit to 9.
Set the next digits to patch, minor and major.


1.2.3-dev.1 becomes ...12301
1.2.3-dev.2 becomes ...12302
1.2.3 becomes ...12390
1.2.4-dev.1 becomes ...12401
10.20.30-dev.40 becomes 102030040.

2147483647 is the max versionCode, however until we hit that, we can bump minsdk. If for any reason we reach max, min(Int.Max, versionCode) can be used until minsdk is bumped.

Given, as of now, we intend to ignore major, we will not reach the max likely.


2.3-dev.1 becomes ...2301
2.3-dev.2 becomes ...2302
2.3 becomes ...2390
2.4-dev.1 becomes ...2401
20.30-dev.40 becomes 2030040.

Axelen123: The only issue that has been raised with deriving the version code from version name is that we may run out of codes, but I guess that is not unsolvable. Other than that, theres no issue with it. One efficient way to derive it would be to divide the bits of the version code among each version component. We could divide the four bytes of the 32-bit version code among each component. The first 8 bits would go to the major and the last 8 bits would go to the prerelease component. This way, each component can be bumped 255 times, and bumping more significant components resets less significant ones. We could also take some bits from major and use them for patch instead, but that might not be necessary. We can always drop support for Android Oreo to get 64-bit version codes in the unlikely event that we need more bits.

oSumAtrIX: Would this be better:

if (prerelease) versionCode += 100 * 9 + prereleaseVersion
versionCode += 1000 * minor
versionCode += 100000 * patch

or short

versionCode += (if (prerelease)  100 * 9 + prereleaseVersion : 0) + 1000 * minor + 100000 * patch;

Otherwise, what you have is okay.

Axelen123: Your method also works, but minor should have a higher multiplier than patch. Other than that, theres no issue. It might not be as efficient, but we can always get 64-bit version codes if we want.

oSumAtrIX: I think instead of shifting, we should just append digits to each other, this way we have a dynamic amount of digits.

Axelen123: Works for me. I could maybe implement this tomorrow.

Ushie: Please don't derive the version code from the version name, we ran into issues with this before with ReVanced Manager flutter but unfortunately it's been a decent amount of time that I don't remember them

Upping by 1 for each build is the standard, stick to that please, there has to be something we're overlooking that such a simple versioncode bump isn't possible, why don't we do it on the github CI itself? read the existing one and save it to env, then revanced manager reads it from env and adds 1 to it

oSumAtrIX: > Please don't derive the version code from the version name

suggest a solution to this review.

> why don't we do it on the github CI itself

Check from here https://github.com/ReVanced/revanced-manager/pull/2675#discussion_r2317093303

> we ran into issues

Which? I cant see any issue

Ushie: Why do it through the semantic release process then? why not do it as it's own step in the release ci?

Ushie: > Which? I cant see any issue

I already mentioned that it had been a while so I no longer remember, but there was absolutely no benefit to base the version code off of the version name, when we have a proper standard dictated by the android docs, there's no motivation to go with that solution

Axelen123: > > why don't we do it on the github CI itself
> 
> Check from here [#2675 (comment)](https://github.com/ReVanced/revanced-manager/pull/2675#discussion_r2317093303)

I don't see any argument here against doing it in gh actions before the build.

Ushie: Yes, what I said earlier can be used here too, start of CI reads and dictates the new version code as an env var, then the normal release process goes in

oSumAtrIX: > why not do it as it's own step in the release ci?

Because bumping the versions belong within in the release step.

> but there was absolutely no benefit to base the version code off of the version name, when we have a proper standard dictated by the android docs, there's no motivation to go with that solution

The reason is technical. Semantic release Gradle can't bump the version code, only the version. Refer to the previous comments which explain all this in detail.

>I don't see any argument here against doing it in gh actions before the build.

It is being done in gh actions. Gh actions runs the release step, which calls Sem release, which in turn calls the Gradle plugin, which bumps the version, before the build and releases it. It's done all in the same step. If you move bumping versionCode to somewhere outside you have scattered locality of code which is bad. So it's either relating version code from version, or forking Sem release Gradle into Sem release android. 

Anything else is problematic. So far I am yet to hear a problem with the current one

Ushie: It would just be another line of this?
https://github.com/ReVanced/revanced-manager/pull/2675/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R58

oSumAtrIX: How does that say anything against what I said. It would be another line, scattered code which is supposed to be together. In one corner of the code base it bumps versionCode in another it bumps version.

Axelen123: Just use the exec plugin then. The current setup doesn't work on Windows anyway (since the semrel plugin uses the unix path) unless you have MSYS2 or Cygwin  (in which case `exec` should also work).

oSumAtrIX: Same issue, as explained earlier it must be in Sem Release Plugin as that's what bumps the version. Place code that bumps anything somewhere else and you have locality issue because two unrelated places of in the code base do one thing.

I don't know what you mean by current setup. If you mean what is currently committed that works as long as git is in path. What I said here doesn't rely on git to begin with

.github/workflows/release.yml
brosssh: `tee >(grep ...)` wouldn't work as expected, https://chatgpt.com/share/68bd736d-b494-8007-aa0e-efd8420eeadd. However we can achieve this with `tee | grep`, as I've done here https://github.com/ReVanced/revanced-manager/pull/2675/commits/1b91261be36870ed0c3b16d1eaccb2dfce7d13c3.

oSumAtrIX: It would you would pipe whatever grep returns and output it tee>(grep ... | output >> github). Your approach is fine.

.github/workflows/release.yml
oSumAtrIX: I think the username can be anything (including the org name)?

.github/workflows/release.yml
```diff
@@ -55,18 +57,26 @@ jobs:
...
+      - name: Release API
+        run: npx multi-semantic-release --tag-format 'api@${version}' --ignore-packages app
         env:
+          GITHUB_ACTOR: ${{ github.actor }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Release
+        id: release
+        run: |
+          echo "NEW_TAG=$(npx multi-semantic-release --tag-format 'v${version}' --ignore-packages api | tee | grep 'Created tag ' | sed -E 's/.*Created tag ([^ ]+).*/\1/')" >> $GITHUB_OUTPUT
```
oSumAtrIX, line 69: Is tag format here necessary? Doesnt it default to this already?

oSumAtrIX, line 69: Instead of two release steps ignoring the respective one, cant you release both, then grep for the app tag?

.github/workflows/release.yml
```diff
@@ -55,18 +57,26 @@ jobs:
...
+        env:
+          GITHUB_ACTOR: ${{ github.actor }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           KEYSTORE_PASSWORD: ${{ secrets.KEYSTORE_PASSWORD }}
           KEYSTORE_ENTRY_ALIAS: ${{ secrets.KEYSTORE_ENTRY_ALIAS }}
           KEYSTORE_ENTRY_PASSWORD: ${{ secrets.KEYSTORE_ENTRY_PASSWORD }}
 
       - name: Attest
```
oSumAtrIX, line 77: Don't we need attest for API module? @validcube iirc answered this once

app/build.gradle.kts
```diff
@@ -118,8 +129,13 @@ android {
         applicationId = "app.revanced.manager"
         minSdk = 26
         targetSdk = 35
-        versionCode = 1
-        versionName = "0.0.1"
+        versionName = if (version == "unspecified") "1.0.0" else version.toString()
+        versionCode = with(versionName!!.toVersion()) {
```
oSumAtrIX: A little correct semantics. We infer code from version. Also naming the variable "versionName" prevents "version.toVersion()"

Maybe reposition code or accept
```suggestion
        val versionString = if (version == "unspecified") "1.0.0" else version.toString()
        versionName = versionString
        versionCode = with(versionString.toVersion()) {
```

app/build.gradle.kts
```diff
@@ -118,8 +129,13 @@ android {
...
+            major * 1_000_000 +
+                    minor * 10_000 +
+                    patch * 100 +
+                    (preRelease?.substringAfterLast('.')?.toInt() ?: 99)
```
oSumAtrIX: Patch and prerelease should have 2, minor should have 3, major can have the rest of the digits.

It is likely we wont use major, if we dont minor will grow infinitly (hence reserving 3 digits). If we decide to use major, we can still adjust the code (with a fixed int to make versioncode be higher than the new adjusted versionCode)

oSumAtrIX (changes requested):
lgtm, just some questions/suggestions

.github/workflows/release.yml
brosssh: No with msr the default is packageName@version.

brosssh: The 2 steps are needed to have separate versioning schemas.

oSumAtrIX: So if we run multi semantic release twice, why is it needed again? Can't we use regular semantic release and run that twice?

brosssh: No because it would e.g. release api when the recent commits are only relevant to the app. Normal semantic release cannot handle this. It was discussed multiple times already, there is no point in going back on this again.

app/build.gradle.kts
brosssh: This code was suggested by you on Discord.

oSumAtrIX: And now I suggest this. I don't know what you mean

Axelen123: By my calculations, all of these are already using the correct power of 10 other than major.

oSumAtrIX: major also has 2 digits

Axelen123: And..? The only thing this code needs to do is shift each component in base 10 based on how many digits the **previous** components have.

.github/workflows/release.yml
Axelen123: > Don't we need attest for API module? @validcube iirc answered this once

Do you have a link or something?

oSumAtrIX: A link for what?

app/build.gradle.kts
oSumAtrIX: Overflows will cause the code to be lower when version is updated:

v1.2.301
would equal to 103001
v1.3.0 
would equal to 103000

Reserve more space as explained above.

.github/workflows/release.yml
Axelen123: I am asking if you have a link (or screenshot) to validcubes answer.

app/build.gradle.kts
Axelen123: I reserved digits according to your first comment. If you said you wanted a different distribution then you didn't explain it very well IMO.

Axelen123: @oSumAtrIX

.github/workflows/release.yml
oSumAtrIX: I don't know where to find it. Perhaps asking him is a good idea

app/build.gradle.kts
oSumAtrIX: I checked the code again, I think it is fine the way it is currently:

major * 10_000_000 +  // major: remaining digits
minor * 10_000 +      // minor: 3 digits
patch * 100 +         // patch: 3 digits
pre                   // preRelease: 2 digits

oSumAtrIX (approved):

.github/workflows/release.yml
```diff
@@ -55,18 +57,26 @@ jobs:
...
-        if: steps.semantic.outputs.new_release_published == 'true'
+        if: steps.release.outputs.NEW_TAG != ''
         uses: actions/attest-build-provenance@v2
         with:
-          subject-name: 'ReVanced Manager ${{ steps.release.outputs.new_release_git_tag }}'
+          subject-name: 'ReVanced Manager ${{ steps.release.outputs.NEW_TAG }}'
           subject-path: app/build/outputs/apk/release/revanced-manager*.apk
```
validcube, line 82: Is build output for app the same?

I saw you changed the location in `.releaserc` from `app/build/outputs/apk/release/revanced-manager*.apk?(.asc)` to `build/outputs/apk/release/revanced-manager*.apk?(.asc)`

Here: https://github.com/ReVanced/revanced-manager/pull/2675/files#diff-140abff1d5b51eb39d74adaa241b58e0fbd92a0600292c00e48ffdcf0ba1b8d9L35-R37


```suggestion
          subject-path: build/outputs/apk/release/revanced-manager*.apk
```
(?.asc) is intentionally omitted because action didn't support extended glob pattern.

.github/workflows/release.yml
validcube: We should attest really about anything that is attached to the release except for unnecessary files like `.asc` (we don't plan on attesting another attestation do we.)

In other word, yes we should attest both `app` and `api`

oSumAtrIX: API is a maven package, not in GitHub releases

validcube: :guh: then no, we do not need to attest the api module

.github/workflows/release.yml
oSumAtrIX: it should be app/build

validcube: resolved

team: cc: https://canary.discord.com/channels/952946952348270622/952987428786941952/1423024185914953809

Axelen123: This has been tested to work and just `build/` is correct in `.releaserc` file because paths are relative to the `app` folder now.

PR #2699, opened 2025-08-16
feat: Allow viewing patches without needing app installed

Aunali321:


oSumAtrIX:
Got a screenshot of how it looks like/where this function is added

Aunali321:
Screenshot_20250817_010855_ReVanced Manager (Development)
Screenshot_20250817_010849_ReVanced Manager (Development)

oSumAtrIX:
the view model needs to be parametrized to not include the checkmark or allow editing options. basically a readonly param

Ushie:
pov main branch

Aunali321:
> the view model needs to be parametrized to not include the checkmark or allow editing options. basically a readonly param

Usually i would agree but this repo is so close to depreciation that it shouldn't be considered. I tried to achieve it with as low code as possible.

lib/ui/widgets/appSelectorView/not_installed_app_item.dart
```diff
@@ -73,6 +73,14 @@ class _NotInstalledAppItem extends State<NotInstalledAppItem> {
                       ),
                     ],
                   ),
+                  Text(
+                    t.appSelectorView.tapToViewPatches, // Use your localization
```
validcube: Might not be necessary needed to have a comment here

```suggestion
                    t.appSelectorView.tapToViewPatches,
```

validcube (approved):
🥞👨‍💻 Close enough, welcome back cossale to the team. :troll:

lib/ui/widgets/appSelectorView/not_installed_app_item.dart
```diff
@@ -73,6 +73,14 @@ class _NotInstalledAppItem extends State<NotInstalledAppItem> {
                       ),
                     ],
                   ),
+                  Text(
```
oSumAtrIX: The UI language doesnt match with the rest of the surrounding UI. Please resolve this with e.g. @validcube

Aunali321:
Screenshot_20250819_141937_ReVanced Manager (Debug)
Screenshot_20250819_141927_ReVanced Manager (Debug)

validcube:
I don't have access to Figma for more than a week now, I don't have my perfect opinion to this.

But I would immediately assume that having a "Tap to view patches" is redundant, why not make a dialog when you select an not installed app said something along the line of ...

image

Aunali321:
i feel like that would be too long. and cancel/ok doesn't make sense because there is nothing to cancel there

oSumAtrIX:
Since tap to view patches is now added, the list of patches should be moved somewhere near to it. Right now its in the title next to the app name, it should be moved to a line for example:

View patches (1)

validcube:
cc: @PalmDevs UI/UX

PalmDevs:
> But I would immediately assume that having a "Tap to view patches" is redundant, why not make a dialog when you select an not installed app said something along the line of ...

I think having dialog blocking an action that isn't destructive isn't ideal. I'd simply show a highlighted card at the top of the page with similar contents.

validcube:
👋 Your PR have been closed due to transition to from ReVanced Manager Flutter to ReVanced Manager Compose (in `dev` branch) have been complete thus your PR is not applicable anymore.

PR #2710, opened 2025-08-27
feat(Compose): Component to handle permissions request (A13+)

brosssh:
Starting from Android 13, notification permission is now not granted by default. With this motivation, I created a component to handle permission requests (`PermissionRequestHandler`). This will show a dialog with an explantion of why the permission is requested, and will allow the user to give or reject the permission. Currently this is just used for the notifications permission, but it's possible other permissions will be needed in the future.

image0

I have also added a notification when the patching process is completed (both in case of success and failure).
Screenshot_2025-08-27-15-43-29-91_b783bf344239542886fee7b48fa4b892


app/src/main/res/values/strings.xml
```diff
@@ -314,8 +310,12 @@
...
     <string name="patcher_step_write_patched">Write patched APK file</string>
     <string name="patcher_step_sign_apk">Sign patched APK file</string>
-    <string name="patcher_notification_title">Patching in progress…</string>
-    <string name="patcher_notification_text">Tap to return to the patcher</string>
+    <string name="patcher_notification_running_title">Patching in progress…</string>
+    <string name="patcher_notification_running_text">Tap to return to the patcher</string>
+    <string name="patcher_notification_success_title">Patching completed</string>
+    <string name="patcher_notification_success_text">Patching completed! Tap to return to the patcher</string>
+    <string name="patcher_notification_failure_title">Failure while patching</string>
+    <string name="patcher_notification_failure_text">Patching completed with errors. Tap to return to the patcher</string>
```
brosssh, line 318: My english sucks, please propose anything that works better in these strings.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -235,7 +273,12 @@ class PatcherWorker(
             updateProgress(state = State.COMPLETED) // Signing
 
             Log.i(tag, "Patching succeeded".logFmt())
+            patcherResult = PatcherResult.SUCCESS
+
             Result.success()
+        } catch (_: kotlinx.coroutines.CancellationException) {
+            patcherResult = PatcherResult.STOPPED
+            Result.failure()
```
Axelen123: Coroutine cancellations should be rethrown.

app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
```diff
@@ -48,19 +52,48 @@ import app.revanced.manager.ui.component.AppScaffold
...
+            onResult = { _ -> askNotificationPermission = false }
+        )
+    }
+
+    PatcherMainScreen(onBackClick, viewModel)
+}
+
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+private fun PatcherMainScreen(
```
Axelen123: Why was the function split?

app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
```diff
@@ -48,19 +52,48 @@ import app.revanced.manager.ui.component.AppScaffold
...
 
-@OptIn(ExperimentalMaterial3Api::class)
+@SuppressLint("InlinedApi")
 @Composable
 fun PatcherScreen(
     onBackClick: () -> Unit,
     viewModel: PatcherViewModel
+) {
+    val context = LocalContext.current
+    var askNotificationPermission by remember { mutableStateOf(true) }
```
Axelen123: Move this to a viewmodel object. I described it in a different comment.

app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
```diff
@@ -48,19 +52,48 @@ import app.revanced.manager.ui.component.AppScaffold
...
+            input = Manifest.permission.POST_NOTIFICATIONS,
+            title = stringResource(R.string.ask_permission_notification),
+            description = stringResource(R.string.ask_permission_notification_description),
+            icon = Icons.Outlined.Notifications,
+            onDismissRequest = { askNotificationPermission = false },
+            onResult = { _ -> askNotificationPermission = false }
+        )
+    }
+
+    PatcherMainScreen(onBackClick, viewModel)
```
Axelen123: Why do we need two functions? Can't you put the new permission stuff inside the existing function?

app/src/main/java/app/revanced/manager/ui/component/RequestPermissionDialog.kt
```diff
@@ -0,0 +1,87 @@
...
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.stringResource
+import app.revanced.manager.R
+import app.revanced.manager.util.permissions.PermissionHelper
+
+@Composable
+fun PermissionRequestHandler(
+    contract: ActivityResultContract<String, Boolean>,
+    input: String,
```
Axelen123: `permission` would be more clear.

app/src/main/java/app/revanced/manager/ui/component/RequestPermissionDialog.kt
```diff
@@ -0,0 +1,87 @@
...
+                    launcher.launch(input)
+                }
+            )
+        }
+        PermissionHelper.PermissionState.DeniedPermanently -> {
+            LaunchedEffect(Unit) {
+                onResult(false)
+            }
+        }
+    }
```
Axelen123: The complexity of the permission tracking needs/logic has increased. The [Unidirectional data flow](https://developer.android.com/develop/ui/compose/architecture#udf) pattern should be applied and the tracking logic should be moved to an object stored in the viewmodel.

All this object needs to do is expose/initialize the current permission state as a read-only variable and have a function that the UI can use to notify the object that permissions may have changed. `PermissionRequestHandler` can then be simplified to just showing a dialog and launching the permission contract when needed, and other code can read/react to the permission state with compose APIs. Let me know if you have any questions or want clarification.

app/src/main/java/app/revanced/manager/util/permissions/Helper.kt
```diff
@@ -0,0 +1,44 @@
...
+import android.content.Context
+import android.content.pm.PackageManager
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
+import androidx.core.content.edit
+
+class PermissionHelper(private val context: Context) {
+    private val prefs by lazy {
+        context.getSharedPreferences("permissions_pref", Context.MODE_PRIVATE)
+    }
```
Axelen123: We have `PreferencesManager`. There is no need to use shared preferences here.

app/src/main/java/app/revanced/manager/util/permissions/ContextExtension.kt
```diff
@@ -0,0 +1,24 @@
...
+        PermissionHelper(this).isPermissionGranted(Manifest.permission.POST_NOTIFICATIONS)
+    else
+        true
+}
+
+fun Context.shouldAskNotificationPermission(): Boolean {
+    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU)
+        !PermissionHelper(this).isPermissionGranted(Manifest.permission.POST_NOTIFICATIONS) &&
+                PermissionHelper(this).getPermissionState(
+                    this as Activity, Manifest.permission.POST_NOTIFICATIONS
```
Axelen123: The extension function should be on `Activity` if it is only supposed to be used on activities.

app/src/main/java/app/revanced/manager/util/permissions/Helper.kt
```diff
@@ -0,0 +1,44 @@
+package app.revanced.manager.util.permissions
+
+import android.app.Activity
+import android.content.Context
+import android.content.pm.PackageManager
+import androidx.core.app.ActivityCompat
+import androidx.core.content.ContextCompat
+import androidx.core.content.edit
+
+class PermissionHelper(private val context: Context) {
```
Axelen123: Use Koin dependency injection instead of instantiating this everywhere. Check the codebase (`di` folder is important) and Koin documentation.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -80,46 +79,51 @@ class PatcherWorker(
...
         }
-
-        try {
-            // This does not always show up for some reason.
-            setForeground(getForegroundInfo())
-        } catch (e: Exception) {
-            Log.d(tag, "Failed to set foreground info:", e)
+        if (applicationContext.hasNotificationPermission()) {
+            try {
+                // This does not always show up for some reason.
```
Axelen123: That reason is probably because we never requested notification permissions. This comment should be removable now.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -80,46 +79,51 @@ class PatcherWorker(
...
+                0,
+                Intent(applicationContext, MainActivity::class.java).apply {
+                    flags = Intent.FLAG_ACTIVITY_SINGLE_TOP or Intent.FLAG_ACTIVITY_CLEAR_TOP
+                },
+                PendingIntent.FLAG_IMMUTABLE
+            ),
+            applicationContext.getString(R.string.patcher_notification_running_title),
+            applicationContext.getString(R.string.patcher_notification_running_text),
+            autoCancel = false
+        ).first
```
Axelen123: You could probably do something like this to get rid of `.first`:
```kt
  val (patcherRunningNotif, _) = workerRepo.createNotification(....)
```

app/src/main/res/values/strings.xml
Axelen123: Flutter manager might have similar strings. Otherwise, I can try and help you come up with strings later.

Axelen123 (changes requested):
I don't know if permissions should be requested on app launch or in the patcher screen. This is up to the UI team to decide.

brosssh:
> I don't know if permissions should be requested on app launch or in the patcher screen. This is up to the UI team to decide.

My reasoning for putting the dialog in the patcher screen is that there is already a popup shown at app first launch (to configure updates). Once that is closed, there are two mores "alert", one for the battery optimization and one for the downloaders configuration.

I didn't want to add another dialog to not overwhelm the user; also the notification permission is striclty related to the patching process (for now at least). Of course, it can be changed according to what UI team thinks.

app/src/main/java/app/revanced/manager/ui/component/RequestPermissionDialog.kt
```diff
@@ -0,0 +1,40 @@
...
+
+@Composable
+fun PermissionRequestDialog(
+    vm: PermissionViewModel,
+    contract: ActivityResultContract<String, Boolean>,
+    title: String,
+    description: String,
+    icon: ImageVector
+) {
+    val activity = LocalActivity.current!!
```
brosssh, line 24: Not sure if the `!!` is fine in this context.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -80,46 +79,50 @@ class PatcherWorker(
...
-
-        try {
-            // This does not always show up for some reason.
-            setForeground(getForegroundInfo())
-        } catch (e: Exception) {
-            Log.d(tag, "Failed to set foreground info:", e)
+        if (applicationContext.hasNotificationPermission()) {
+            try {
+                setForeground(getForegroundInfo())
+            } catch (e: Exception) {
```
Axelen123, line 127: We should rethrow cancellations here.

app/src/main/java/app/revanced/manager/util/permissions/ContextExtension.kt
```diff
@@ -0,0 +1,12 @@
...
+import android.Manifest
+import android.content.Context
+import android.os.Build
+
+fun Context.hasNotificationPermission(): Boolean {
+    return if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU)
+        PermissionHelper().isPermissionGranted(Manifest.permission.POST_NOTIFICATIONS)
+    else
+        true
+}
```
Axelen123: This probably makes more sense as a method on `PermissionHelper`.

app/src/main/java/app/revanced/manager/ui/viewmodel/PermissionViewModel.kt
```diff
@@ -0,0 +1,34 @@
...
+import androidx.lifecycle.viewmodel.compose.SavedStateHandleSaveableApi
+import app.revanced.manager.util.permissions.AndroidPermission
+import app.revanced.manager.util.permissions.PermissionHelper
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.launch
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
+
+@OptIn(SavedStateHandleSaveableApi::class)
```
Axelen123: I don't think this is used.

app/src/main/java/app/revanced/manager/ui/viewmodel/PermissionViewModel.kt
```diff
@@ -0,0 +1,34 @@
...
+class PermissionViewModel(
+    @AndroidPermission val permission: String
+) : ViewModel(), KoinComponent {
```
Axelen123: You can probably rename this and get rid of `ViewModel()`. Callers from Composables or Viewmodels can easily get their own coroutine scopes.

app/src/main/java/app/revanced/manager/ui/screen/AppSelectorScreen.kt
```diff
@@ -59,6 +64,17 @@ fun AppSelectorScreen(
...
+    val androidContext = LocalContext.current
+
+    if (vm.android11BugActive)
+        PermissionRequestDialog(
+            vm = koinViewModel { parametersOf(androidContext.packageName) },
+            contract = RequestInstallAppsContract,
+            title = stringResource(R.string.ask_permission_installation),
+            description = stringResource(R.string.ask_permission_installation_description),
+            icon = Icons.Outlined.BugReport
+        )
```
Axelen123: Why was this moved here? For context, the bug on android 11 means that apps will be killed if the user grants them installation perms, so it seems better to do it earlier.

app/src/main/java/app/revanced/manager/ui/component/RequestPermissionDialog.kt
```diff
@@ -0,0 +1,40 @@
...
+    val shouldShowDialog by vm.shouldShowDialog.collectAsStateWithLifecycle(false)
+
+    LaunchedEffect(Unit) {
+        vm.refreshPermissionState(activity)
+    }
+
+    val launcher = rememberLauncherForActivityResult(contract) {
+        _ -> vm.refreshPermissionState(activity)
```
Axelen123: Move `_ ->` after the `{` or get rid of it.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -235,19 +271,28 @@ class PatcherWorker(
...
+            patcherResult = PatcherResult.STOPPED
+            throw e
         } catch (e: ProcessRuntime.RemoteFailureException) {
-            Log.e(
-                tag,
-                "An exception occurred in the remote process while patching. ${e.originalStackTrace}".logFmt()
-            )
+            Log.e(tag,
+                ("An exception occurred in the remote process while patching." +
+                        " ${e.originalStackTrace}").logFmt())
```
Axelen123, line 286: Is there a reason to split this?

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -139,7 +142,40 @@ class PatcherWorker(
...
+            ),
+            applicationContext.getString(
+                if (success) R.string.patcher_notification_success_text
+                else R.string.patcher_notification_failure_text
+            )
+        ).also { (notification, notificationManager) ->
+            if (applicationContext.hasNotificationPermission())
+                notificationManager.notify(
+                    System.currentTimeMillis()
+                        .toInt(), // An unique ID, each patch process should show a notification
```
Axelen123:
```suggestion
                        .toInt(), // A unique ID, each patch process should show a notification
```

app/src/main/java/app/revanced/manager/ui/component/RequestPermissionDialog.kt
Axelen123: It is valid to use `!!`, since we can expect there to be an activity present.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
brosssh: Just to respect the right margin.

kitadai31:
- The new "Patching completed" notification should not include `Tap to return to the patcher` sentence.
Because it will be displayed even when the patcher screen is open. This text was originally created for background running notifications. It's not necessary in the patching completion notification.
- The new "Patching completed" notification is a bit annoying. What's particularly annoying is having to manually clear the notifications after every patching.
It would be convenient if notifications were automatically deleted upon leaving the patcher screen.

app/src/main/java/app/revanced/manager/ui/component/RequestPermissionDialog.kt
```diff
@@ -0,0 +1,43 @@
...
+import androidx.compose.runtime.derivedStateOf
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.remember
+import androidx.compose.ui.graphics.vector.ImageVector
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
+import app.revanced.manager.ui.viewmodel.PermissionStateHolder
+
+@Composable
+fun PermissionRequestDialog(
+    vm: PermissionStateHolder,
```
Axelen123: call this state or stateHolder

Axelen123 (changes requested):

MarcaDian:
Hi. I see in your screenshot that there is a problem with the white ReVanced logo being displayed on a white background. I plan to fix this in https://github.com/ReVanced/revanced-manager/pull/2762, can you please rebase this PR to the dev branch? Thanks

By the way, does this PR fix the lack of a notification icon in the status bar during patching?

Issue #2150, opened 2024-08-18
feat: Ability to disable a bundle


indrastorms:
### Motivation

For some reason if i don't want to see the patches of bundle provided by api in patch selection screen, i can't delete the bundle because readding will not be easy by design. 

But we can have a disable option for every bundle.
For temporarily disable a bundle for testing purposes and for the ability to get rid of api bundle too.

### Acknowledgements

- [X] I have checked all open and closed feature requests and this is not a duplicate.
- [X] I have chosen an appropriate title.
- [X] The feature request is only related to ReVanced Manager.

brosssh:
Closed for the same reason of the linked pr.

PR #2731, opened 2025-09-08
feat: Disable bundle

brosssh:
Added backend support for enabling and disabling bundles. When a bundle is disabled, the related patches are hidden from both the app selection screen and the patch selection screen.

UI is missing for now, awaiting feedback from designers.

Closes https://github.com/ReVanced/revanced-manager/issues/2150.

### DIFF
--- a/app/src/main/java/app/revanced/manager/patcher/patch/PatchBundleInfo.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/patch/PatchBundleInfo.kt
@@ -22,7 +22,7 @@ sealed class PatchBundleInfo {
     abstract val uid: Int
 
     /**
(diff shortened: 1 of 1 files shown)


brosssh:
Closed because @oSumAtrIX doesn't want this. I will maintain it in my fork, when/if the compose manager get released.

oSumAtrIX:
Please attach the reasoning otherwise this looks like I just disgard it for no reason.

brosssh:
The reasoning can be found in the duplicate issues, as you said. Otherwise you can add it yourself, I don't want to talk for you.

oSumAtrIX:
Linking to it is fine

Issue #1191, opened 2023-07-20
bug: issues with current UI and UX


PalmDevs:
These are some UI/UX issues in the current implementation of ReVanced Manager.

## 📄 General things

- [ ] 🛑 **VERY IMPORTANT**: Use Storage Access Framework instead of custom file picker. It has been proven that users don't want to give All files access permissions to Manager.
- [ ] 🟠 **Splash screen is a broken in light mode**
  Current splash screen's **V** shape can hardly be seen in light mode. This should be fixed by putting the current **V** logo in the splash screen inside a circle which should always be dark.
- [ ] 🔵 Scrim for dialogs/bottom sheets/etc. (gray overlay) does not extend to the unsafe area (where time, notifications, and phone status are displayed)
- [x] ~🔴 **Downloader shouldn't download in Installer**
  So that the user can view the compatible patches before even downloading the file, meaning they can switch to any other version without wasting any bandwidth~

## ~~📱 Select an app screen~~

- [x] ~~🟠 The **Select from storage** option should have a description like in the design~~
- [x] ~~🔵 **The search feature doesn't have any page transition when clicked**
  See Material Design website on which transition to use.~~
- [x] ~~🔵 The trailing supporting text for the app list entries shouldn't have the word `Patches` capitalized~~
## ~~👆🏻 Select version screen~~
- [x] ~🟠 Version selector doesn't use radio buttons and FAB~

## 🪄 Select patches screen

- [ ] 🔴 **Filter options will need to be relocated**
  According to the design, it will need to be relocated into a bottom sheet which can be triggered by a top app bar icon. 
- [ ] 🔵 **List items for patches that have options look odd**
  The gear icon looks too far off, try decreasing the right padding/margin.

## ⬇️ Installer screen

- [ ] 🟠 **Action buttons are not always at the bottom of the screen**
  This happens when the card heights do not go beyond the screen's height.
- [ ] 🟡 The text and icon sizes do not follow the design
![Screenshot 2023-07-20 215239](https://github.com/ReVanced/revanced-manager-compose/assets/67689562/5b980ec2-2781-4fe0-a98a-cb0a9c63dbd3)

## ⚙️ All settings screens

- [ ] 🔴 Inconsistencies between pages
  Some use a normal list item, some use an extended/custom one.

- [ ] 🟡 Battery optimization notification in settings page is wrongly colored
- [ ] 🔵 The list item and the label container itself must always have 24dp padding in this part of the app

### ⚙️ Settings

[...]

Issue #1124, opened 2023-08-10
bug: monochrome icon is not properly sized (again)


validcube:
## 💄 Monochrome icon is not properly sized (again)

The monochrome icon for the ReVanced Manager is not properly the sized, the size for the monochrome icon is very small when compared to other apps.

Below this is an screenshot taken by Lawnchair Launcher (Debug, CI Build #2054). This result can also be achieve using Samsung launcher (but personal opinion :trollface:)

![Screenshot_20230810_194221_Lawnchair (Debug)](https://github.com/ReVanced/revanced-manager/assets/93124920/6369b49d-a8e9-4cc8-bb9c-fda3d45d325e)

-------------------------------------------------------

Related issues: #714, #753

Ushie:
Standing by

validcube:
> @Ushie: Standing by

Done:tm:!

ILoveOpenSourceApplications:
For this to be solved properly, we would also have to resize our original icon size as well, as it is small compared to other icons. Currently the original icon size matches the monochrome. So if we update the monochrome icon, we'll also have to resize the original icon.

oSumAtrIX:
ReVanced Manager will soon be released as [a full rewrite](https://github.com/ReVanced/revanced-manager/tree/compose-dev). Over time, many feature requests have been made for the current version of ReVanced Manager, but to accelerate the development of the new version of ReVanced Manager, we have to focus on it, which I am closing this issue according to the [contribution guidelines](https://github.com/ReVanced/revanced-manager/blob/main/CONTRIBUTING.md#-submitting-a-feature-request):

> Requests can be accepted or rejected at the discretion of maintainers of ReVanced Patches

This feature request has been closed for one or more of the following reasons:

- Stale for too long/ Low priority/ Maintainers un-interested/ Too complex to realize
- Feature planned/ already implemented for the new version of ReVanced Manager

Regardless, anyone who comes across this feature request can PR it to the current version of ReVanced Manager until the new version is ready. Thanks for your feature request!

validcube:
Re-open, this issue still persist even in rvmc context.

MarcaDian:
@validcube will be fixed by https://github.com/ReVanced/revanced-manager/pull/2762

validcube:
This is technically already fixed but that just make our icon have two different sizing target for original and monochrome. How fun!

PR #2635, opened 2025-07-05
fix: Use system background colour for V-shaped outline

validcube:

PR #2762, opened 2025-10-01
fix: Adjust ReVanced icons

MarcaDian:
Fix incorrect icon logo,  https://github.com/ReVanced/revanced-manager/issues/1124 and `Splash screen is a broken in light mode` from https://github.com/ReVanced/revanced-manager/issues/1191#issue-1873984131
___
### Round/Adaptive icon
| Before (xxhdpi)  | After (xxhdpi) |
| ------------- | ------------- |
| ic_launcher | ic_launcher |

### Monochrome icon

___
### Splash screen
| Dark  | Light |
| ------------- | ------------- |
|  |  |


MarcaDian:
@oSumAtrIX Look at these 2 icons and finally decide which one to use? rounded or sharp?

https://github.com/ReVanced/revanced-branding/blob/main/assets/revanced-logo/revanced-logo.svg

https://github.com/ReVanced/revanced-branding/blob/main/assets/revanced-logo/revanced-logo-shape.svg

image

Ushie:
Which one is in our revanced-branding repository?

LisoUseInAIKyrios:
@Ushie both files are in the branding repo.

Somehow two slightly different designs exist there, and nobody has noticed until now.

We should make a decision, then delete/fix the incorrect icons in that repo.

LisoUseInAIKyrios:
It seems [this](https://github.com/ReVanced/revanced-branding/blob/main/assets/revanced-logo/revanced-logo-shape.svg) file is the only file with the conflicting design (sharp edges and not rounded).

Since all the other branding image resources use rounded edges, I think was can say this the rounded design used by the 7 other files is the correct design.

So the solution is to use the rounded design here, and open a new issue in the branding repo so that incorrect file can be fixed/deleted?

MarcaDian:
The second logo (the sharp one) has been used as the icon in the manager for 2 years, although with incorrect proportions of the foreground area of the adaptive icon.

Because of this, we updated the icons in the custom branding patch for YouTube, since we thought this was the actual icon.

It’s very strange that there are 3 different icon designs in the ReVanced projects, and it seems that no one ever noticed this - but as soon as we fixed it, questions started to arise.

LisoUseInAIKyrios:
It seems everytime the icons are touched there is some kind of controversy.

Well, do we stick with the "wrong" sharp icon here that's been used with Manager for the past 2 years? Or do we change to the rounded (and make another PR in Patches to change to that)?

Patches repo could add one or two additional preset icons for the custom branding patch. One preset with the color ring, and maybe another with an [expanded ring](https://github.com/ReVanced/revanced-patches/pull/6029#issuecomment-3357919193) that was available with the prior icon design that worked with odd launcher icon shapes.  But that is unrelated to the question of which icon to use.

LisoUseInAIKyrios:
If we can't come to a decision then we should merge what is currently here, because it is the icon style Manager has already been using the past 2 years.

Ushie:
Use the rounded ones, an issue should be opened for the sharp one to be fixed

oSumAtrIX:
I don't understand whats going on here. What issue does the branding repo have? They are all svgs, I dont understand what you mean with sharp/not sharp. 

The PR body shows the diamond scaled more than current. Please use the scaling thats also used in the logo from the branding repo.

oSumAtrIX:
> @oSumAtrIX Look at these 2 icons and finally decide which one to use? rounded or sharp?
> 
> [ReVanced/revanced-branding@`main`/assets/revanced-logo/revanced-logo.svg](https://github.com/ReVanced/revanced-branding/blob/main/assets/revanced-logo/revanced-logo.svg?rgh-link-date=2025-10-01T22%3A42%3A53Z)
> 
> [ReVanced/revanced-branding@`main`/assets/revanced-logo/revanced-logo-shape.svg](https://github.com/ReVanced/revanced-branding/blob/main/assets/revanced-logo/revanced-logo-shape.svg?rgh-link-date=2025-10-01T22%3A42%3A53Z)
> 
> image


I do not understand what you mean. The links you sent are two different things. The shape is ONLY the shape, the round one is the full logo with the ring.

You should use the shape in scenarios where the background is configured separately, in other words:

1. Create an adaptive icon background
2. Extract the shape from the FULL logo from the branding, make sure the scaling is kept
3. Use the extracted shape in that scale as the adaptive icon foreground

I am not sure if android does additional scaling processing so you'll have to compare with the android spec. If they do not apply any additional scaling, the final logo should look exactly like the logo from the branding.

LisoUseInAIKyrios:
@oSumAtrIX look carefully at those two images. They are not the same. One has rounded corners, the other is "sharp" corners. Look at the bottom point of the inner small V. The two images are different.

https://github.com/ReVanced/revanced-manager/pull/2762#issuecomment-3358407527

The non rounded icon image in the branding repo must be a mistake, because all other images are rounded. It should be fixed there, but also fixed here.

The logo here has no round circle background, since there isn't a cut shape for adaptive icons. The old manager icon did not use the correct scale and went against android guidelines. This PR fixes all of this for Manager.

oSumAtrIX:
I see the problem and I know why it happens.

Affinity Design handles roundness based on the canvas size. I think when I created the shape, I did those steps:

1. Remove the background layer
2. Scale the shape to fit the canvas

Step two scales the shape, but it does not scale the rounding value. 
A simple fix is to scale down the canvas to the size of the shape. This way the shape fills the entire canvas but attributes like the roundness remain correct. I will fix this.

oSumAtrIX:
And yes, the full logo is the "correct" one. All variations are derrivates of it

oSumAtrIX:
Ok seems it was caused by https://github.com/ReVanced/revanced-branding/pull/10/files

oSumAtrIX:
Fixed now https://github.com/ReVanced/revanced-branding/pull/31/commits/41c7155a58401eefc32acc22d23791f51148f202

MarcaDian:
I think `ic_launcher_round.png` can be deleted, since it is not used anywhere, it just takes up space

LisoUseInAIKyrios:
They can still be used in some edge cases, ~~so I think they should be kept~~ Edit: Legacy round icon can be deleted.

The files are too small to make any difference.

MarcaDian:
It has an unusual size of 400 by 400, while xxxhdpi has a maximum of 192

LisoUseInAIKyrios:
400x400 looks like it's much larger than needed.

But the [manifest](https://github.com/ReVanced/revanced-manager/blob/dev/app/src/main/AndroidManifest.xml) is missing the `android:roundIcon` declaration.

If it's gone this long and nobody has noticed it's missing in any of the edge cases, then that image should be deleted because the file isn't doing anything without the manifest declaration.

MarcaDian:
Also fix `Splash screen is a broken in light mode` from https://github.com/ReVanced/revanced-manager/issues/1191#issue-1873984131

MarcaDian:
@brosssh 

| Dark  | Light |
| ------------- | ------------- |
|   |   |

Ushie:
For the splash screen, only the shape logos should be used

The background of the icon isn't necessary here as it's displayed directly on the app's background

Ushie:
I would say to use image, although I admit that it would look worse for the dark mode users

I'm wondering whether an official light/dark shape icon with the diamond colored in can be made 

@oSumAtrIX ?

Essentially this but with a dark V https://github.com/ReVanced/revanced-branding/blob/main/assets/revanced-logo/revanced-logo-shape.svg

ReVanced Manager would be able to do that as the branding guidelines allow for it, but ReVanced may be interested in officially including it in the branding repository?

MarcaDian:
I think this is easy to do using an attribute `android:fillColor="?android:attr/textColorPrimary"` in svg/xml, this is how we color the icons in the YouTube ReVanced settings.

oSumAtrIX:
> I'm wondering whether an official light/dark shape icon with the diamond colored in can be made

Yes, sounds reasonable. I will probably get rid of the dark light variants in branding repo and just include a single color example, since you can just invert the shades depending on light or dark mode. Doing it now.

MarcaDian:
50% of logo

| Dark  | Light |
| ------------- | ------------- |
|  |  |

Ushie:
Don't you think it's too big?

MarcaDian:
The default size is 30% (https://github.com/ReVanced/revanced-manager/pull/2762#issuecomment-3363388648), in the screenshot above I set it to 50% for the test, referring to https://github.com/ReVanced/revanced-manager/pull/2635#pullrequestreview-3176155355

MarcaDian:
Default 30%

| Dark  | Light |
| ------------- | ------------- |
|  |  |

MarcaDian:
I also tried 40%

| Dark  | Light |
| ------------- | ------------- |
|  |  |

Leaving 30%?

Ushie:
30% looks perfect

oSumAtrIX:
Isn't there an android standard for the size

MarcaDian:
There’s no strict standard size defined for splash screens.
The system uses the same scaling as the launcher adaptive icon - the foreground layer is 72dp within a 108dp area (66% scale).
Everything else is adjusted automatically depending on device density.

> Splash screen dimensions
The splash screen icon uses the same specifications as adaptive icons, as follows:

> Branded image: this must be 200×80 dp.
App icon with an icon background: this must be 240×240 dp and fit within a circle 160 dp in diameter.
App icon without an icon background: this must be 288×288 dp and fit within a circle 192 dp in diameter.
For example, if the full size of an image is 300×300 dp, the icon needs to fit within a circle with a diameter of 200 dp. Everything outside the circle turns invisible (masked).

MarcaDian:
@validcube hello, according to [this comment](https://github.com/ReVanced/revanced-patches/pull/6059#issuecomment-3370919896), the old "correct" scaling for the monochrome icon needs to be restored, which means [this issue](https://github.com/ReVanced/revanced-manager/issues/1124) is open again.
Should I make these changes or leave the current scaling?
And I see that there is no consensus on this issue in the team. If we return it, then this issue should be closed so as not to cause confusion.

validcube:
> @validcube hello, according to [this comment](https://github.com/ReVanced/revanced-patches/pull/6059#issuecomment-3370919896), the old "correct" scaling for the monochrome icon needs to be restored, which means [this issue](https://github.com/ReVanced/revanced-manager/issues/1124) is open again. Should I make these changes or leave the current scaling? And I see that there is no consensus on this issue in the team. If we return it, then this issue should be closed so as not to cause confusion.

Sorry for the slow reply, I did not receive any notification from GitHub at all.

Edited, sorry I missed understand your point.

Edited again, yeah, you should leave it...

validcube (approved):
🥞💄 LGTM!

Verified with Android Studio, personally I would make the icon bigger by resizing it to 115% or 130% to account for optical alignment *slightly outside the circle or touching the smallest square grid because our logo's a V* but an decision has already been made so that's fine by me (or not, there's like 5 episodes of this discussion).

Image

Haven't verified with an actual device because I can't disable Google Play Protect, it won't let me do it for some reason.

ILoveOpenSourceApplications:
I'm confused here from all the screenshots, but is the monochrome icon slightly bigger than the original icon?

oSumAtrIX:
Yes

LisoUseInAIKyrios:
Monochrome icons don't have a background. The sizing here matches Android design specs and is correct .

See https://github.com/ReVanced/revanced-manager/issues/1124

validcube:
Assuming no pushback, I'm merging this by tomorrow.

oSumAtrIX:
To confirm, is the shape scaled according to the official logo & is the monochrome icon scaled according to androids size standard? Are the scalings guessed by eye or calculated?

LisoUseInAIKyrios:
Scaling is correct for both brand guidelines for color icons and Android spec for monochrome. Nothing is eyeballed.

oSumAtrIX:
Last time I tried scaling, I had to eyeball it. What's the formula you used to scale?

app/src/main/res/drawable/ic_launcher_foreground.xml
```diff
@@ -1,30 +1,40 @@
+<!-- Copyright 2024 ReVanced. Not licensed under GPL. See https://github.com/ReVanced/revanced-branding -->
```
validcube, line 1: Personally I would state the warning right away in readme or do both, also I don't think adding years make sense as that can be outdated every year or probably raises weird loophole legality questions in the future. (I'm not a lawyer)

LisoUseInAIKyrios: 2024 is the year of the logo redesign (the origin of that particular branded resources). But it could be the year the project started. I don't think it makes any difference which year is used (or if no year at all is used).

See the patches repo for the open PR that adds a simple text file explaining the branding isn't GPL. After it's merged Manager could add the same file.

I don't think the "brand identity is not GPL" is needed in the readme. All the major open source projects I've come across that have copyrighted non open source files have the disclaimer somewhere on the files, or they go one step further and separate the branding into a separate project (such as Chromium and the Chrome browser).

ILoveOpenSourceApplications:
Who and why was this PR closed?

Ushie:
It was closed by the author, explanation here:
https://github.com/ReVanced/revanced-patches/issues/2740#issuecomment-3678118256

Issue #2780, opened 2025-10-03
bug: Inconsistencies with pre-release option naming


ILoveOpenSourceApplications:
### Bug description

For ReVanced Manager pre-release:
![Image](https://github.com/user-attachments/assets/c3060690-0492-478a-97e3-2584bf2bc0e1)

For ReVanced Patches pre-release:
![Image](https://github.com/user-attachments/assets/c043c208-a544-4701-bb86-1b9b7be48540)

Both should follow a similar naming and description style. 

### Version of ReVanced Manager and version & name of app you are patching

v.126.0-dev.6

### Installation method

None

### ReVanced Manager logs

```shell
Nil
```

### Patch logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

Ushie:
I have this solved locally, will be in my next push

Ushie:
https://github.com/ReVanced/revanced-manager/commit/e1b768c4679ecae8bff8007bdab56ff6544b12b6

ILoveOpenSourceApplications:
One small inconsistency remains,  one says `pre-release version` while the other says `pre-release versions`.

PR #2785, opened 2025-10-03
docs: Fix typos and settings order

KobeW50:


docs/2_6_settings.md
```diff
@@ -4,29 +4,29 @@ Learn how to configure ReVanced Manager.
...
     Disabling this will allow you to patch any version of apps
     >   ⚠️ Warning  
     >   Patches not compatible with the selected version of the app will not be used.
     >   Unless you know what you are doing, it is recommended to keep this enabled.
-  - **Allow universal patches**: Patches that do not specify compatibility with an app explicitly are forcibly disabled.
+  - **Allow changing patch selection and options**: The default selection of patches is enforced.
+    Enabling this will allow you to change the patch selection
+    >   ⚠️ Warning  
+    >   Changing the selection may cause unexpected issues.
+    >   Unless you know what you are doing, it is recommended to keep this disabled.
```
KobeW50, line 25: In 1.26.0-dev.7, this option is enabled by default. Is that a bug, or does this warning need to be changed?

validcube: That is enabled by default yes, this is not "expected", I already notified the team in discord.

ILoveOpenSourceApplications:
Please look into https://github.com/ReVanced/revanced-manager/issues/2780#issuecomment-3366470769 as well.

Ushie:
Settings aren't considered final, this PR should be postponed

app/src/main/res/values/strings.xml
```diff
@@ -438,7 +438,7 @@
     <string name="add_patches">Add patches</string>
     <string name="auto_update_description">Automatically update when a new version is available</string>
     <string name="patches_prereleases">Use pre-releases</string>
-    <string name="patches_prereleases_description">Use pre-release version of %s</string>
+    <string name="patches_prereleases_description">Use pre-release versions of %s</string>
```
ILoveOpenSourceApplications, line 442:
```suggestion
    <string name="patches_prereleases_description">Use pre-release version of %s</string>
```
Isn't it better to use version at both places?

KobeW50: In that case the toggle titles will also need to be singular. Currently they are plural ("Use pre-release**s**")

Also, I think "versions" better reflects that the patches and Manager are constantly updated. There isn't a single pre-release version. Rather, it is constantly changing. Not sure if I'm explaining myself well. I understand your perspective too :sweat_smile:

KobeW50:
> Settings aren't considered final, this PR should be postponed

Isn't it best to have it reflect the current state of settings? If the settings are changed, a new PR can be opened in the future. 

I wish not to have this PR hang open, because then I will be responsible to reflect all future changes and mistakes. I don't like putting workloads on myself, so I was hoping this would quickly be merged or closed.

Same for my other PR: https://github.com/ReVanced/revanced-manager/pull/2791

validcube:
> Settings aren't considered final, this PR should be postponed

I don't think this PR should be postponed, this PR is relating to the current branch `dev` not the settings V2 refactor branch, the owner of the refactor branch should be the one updating the documentation instead.

PR #2794, opened 2025-10-03
chore: Use raw string encoding to prevent AI translation errors

LisoUseInAIKyrios:
Crowdin AI pre-translation has issues when using `\n` characters in strings, and it regularly replaces them with new line breaks (which are stripped away in non-raw strings and thus never appear in the app).

Using raw strings for new lines fixes it, because even if the AI gets confused and uses the wrong new line encoding, with raw strings it still works because both regular line breaks and `\n` are valid and work as intended.

Raw strings are also easier to edit on Crowdin because multiline text is shown and edited as multiline (and not a single line with `\n` sprinkled throughout).

### DIFF
--- a/app/CHANGELOG.md
+++ b/app/CHANGELOG.md
@@ -1,3 +1,10 @@
+# app [1.26.0-dev.11](https://github.com/ReVanced/revanced-manager/compare/v1.26.0-dev.10...v1.26.0-dev.11) (2025-10-25)
+
+
+### Features
+
+* Add pure black theme ([#2824](https://github.com/ReVanced/revanced-manager/issues/2824)) ([3d75ffe](https://github.com/ReVanced/revanced-manager/commit/3d75ffe6a7a39efdebe13dbd07c937c1de409ead))
+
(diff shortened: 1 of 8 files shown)


LisoUseInAIKyrios:
Speaking of Crowdin, I manually hide all the flutter strings on Crowdin (can be unhidden if needed), because some users are still making occasional edits to the strings but they are wasting their time since Flutter is on the way out.

LisoUseInAIKyrios:
Someone else needs to merge because I don't have access on this repo.

Until Crowdin actions are setup for Manager Compose, should I manually upload the strings.xml file to Crowdin?

Some users have asked to start translating the new Manager strings, but they can't yet because nothing is on Crowdin right now.

validcube:
> Someone else needs to merge because I don't have access on this repo.
> 
> Until Crowdin actions are setup for Manager Compose, should I manually upload the strings.xml file to Crowdin?
> 
> Some users have asked to start translating the new Manager strings, but they can't yet because nothing is on Crowdin right now.

The decision to upload compose strings to crowdin is something that I very strongly approve of however I would have to get other team member opinion for that instead.

PR #2824, opened 2025-10-16
feat: Add pure black theme

### LINKED ISSUE #2607: feat: Add `Enable Black Theme` option in settings.
### Feature description

This patch will allow users to use `Black Theme` in Compose manger.

There'll be a button, same like Material theme, to enable the black theme. So that, users can choose whether they want to use Black theme or normal theme as dark theme.

If user have enabled that option, the dark theme will automatically use Black theme.

### Motivation

This is very useful for Amoled screens. It'll save more battery in compare to normal Dark theme.

### Acknowledgements

aliernfrog:
Adds a toggle to use pure black backgrounds for dark theme. The toggle is hidden when theme preference is set to light.
Closes #2607

This simply overrides the `background` and `surface` colors of the color scheme and it is pretty much enough.

## Some screenshots
<details>
<summary>Toggle disabled</summary>
image
</details>

<details>
<summary>Toggle enabled</summary>
image
</details>

<details>
<summary>Toggle being hidden when theme is set to light</summary>
image
</details>

<details>
<summary>Homepage with the pure black theme enabled</summary>
image
</details>

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -86,8 +86,8 @@
     <string name="contributors_description">View the contributors of ReVanced</string>
     <string name="dynamic_color">Dynamic color</string>
     <string name="dynamic_color_description">Adapt colors to the wallpaper</string>
-    <string name="pitch_black_theme">Pitch black theme</string>
-    <string name="pitch_black_theme_description">Use pure black backgrounds for dark theme</string>
+    <string name="pure_black_theme">Pure black theme</string>
+    <string name="pure_black_theme_description">Use pure black backgrounds for dark theme</string>
     <string name="theme">Theme</string>
     <string name="theme_description">Choose between light or dark theme</string>
     <string name="safeguards">Safeguards</string>
--- a/app/src/main/java/app/revanced/manager/MainActivity.kt
+++ b/app/src/main/java/app/revanced/manager/MainActivity.kt
@@ -81,7 +81,7 @@ class MainActivity : ComponentActivity() {
             )
             val theme by vm.prefs.theme.getAsState()
             val dynamicColor by vm.prefs.dynamicColor.getAsState()
-            val pitchBlackTheme by vm.prefs.pitchBlackTheme.getAsState()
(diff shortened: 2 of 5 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -86,6 +86,8 @@
     <string name="contributors_description">View the contributors of ReVanced</string>
     <string name="dynamic_color">Dynamic color</string>
     <string name="dynamic_color_description">Adapt colors to the wallpaper</string>
+    <string name="amoled_theme">AMOLED theme</string>
```
oSumAtrIX:
```suggestion
    <string name="amoled_theme">Pitch black theme</string>
```

or

```suggestion
    <string name="amoled_theme">Amoled theme</string>
```

validcube: "AMOLED" is an acronym so the original styling is correct, however it's a display technology (that's now become synonymous with pitch black or pure dark theme), me personally would opt for "pitch black" like osu suggested, or "extra dark" instead.

MarcaDian: In the revanced patches we called it `Pure black`, maybe it would be correct if it had the same name everywhere.
https://github.com/ReVanced/revanced-patches/blob/96454c843b3cec4fa55f85eca54fa4714b71c5e2/patches/src/main/kotlin/app/revanced/patches/shared/layout/theme/BaseThemePatch.kt#L62

aliernfrog: Renamed it to "Pitch black theme" with all the code references.

ILoveOpenSourceApplications: I agree with @MarcaDian here, the naming should be consistent across all ReVanced repos.

validcube: Well in that case (of an already established decision), sure, we can name it "Pure black" for consistency sake.

aliernfrog: Sorry, i missed the "pure" in MarcaDian's comment and thought he was talking about "pitch black". I will do another refactor shortly.

aliernfrog: Done.

image

PR #2889, opened 2025-12-19
feat: Show patches as individual steps in patcher screen

### LINKED ISSUE #2568: feat: Show the complete list of applied patches during the patching process
### Feature description

This can be same as the official RV manager have. In this, instead of using a simple name that this patch been applied, having a detailed patch log/list within the manager will be good. In compose manager, list will go top to bottom, rather than bottom to top (like in RVM), within a specified area. And, after all the patches applied, the area will be automatically collapsed and shows `All patches applied` or `23 patches applied`.

Image

### Motivation

It'll help user to see the patch list and the patch description (like in manager) and can live check the applied patches, instead of generating a log file for that or go back to patch options to see them.

Like in Hide ads, we've so many sub-options like Microsoft ads, yandex ads, etc. User will see which ads patch is applied and which not, same like RV manager.

### Acknowledgements

CnC-Robert:
Closes #2568


app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -497,10 +487,13 @@ class PatcherViewModel(
             LogLevel.ERROR -> Log.e(TAG, msg)
         }
 
-        fun generateSteps(context: Context, selectedApp: SelectedApp): List<Step> {
+        fun generateSteps(context: Context, selectedApp: SelectedApp, patchSelection: PatchSelection): List<Step> {
             val needsDownload =
                 selectedApp is SelectedApp.Download || selectedApp is SelectedApp.Search
 
+            val patchSteps = patchSelection.values.flatten().sortedBy { it }
```
Axelen123: I'm not sure why we are sorting, but wouldn't `sorted` work?

oSumAtrIX:
Can a screenshot of the change be uploaded? Spares me compiling on my side

CnC-Robert:
> Can a screenshot of the change be uploaded? Spares me compiling on my side

Screenshot_20251221_235735_ReVanced Manager (Debug)

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
CnC-Robert: Because the patches are provided in random order but always executed in alphabetical order.
Yes I changed it to using sorted().

oSumAtrIX: Keep in mind, order is not contracted. The order can change at any time. You should not make an assumption about it until contracted with the API. Instead of showing all patches right away, "build" the list with the latest patch executing, just like it is done in flutter manager. In flutter manager, we simply append to the output from the flow returned by executing the patcher.

CnC-Robert: Would it be better to show the full sorted list from the start but mark them as completed in whatever order the patcher completes them? Worst case the patches don't finish in the order they are shown but the user still has a full list.

oSumAtrIX: I think this would be weird UX, so dont show them all upfront

CnC-Robert: So it should only show a list of patches that already completed?

oSumAtrIX: Like explained here

> Instead of showing all patches right away, "build" the list with the latest patch executing, just like it is done in flutter manager. In flutter manager, we simply append to the output from the flow returned by executing the patcher.

CnC-Robert: https://github.com/user-attachments/assets/1ad52789-5e7a-4379-be59-785cf89dd1b0

So something like this?

Axelen123: That seems fine to me.

oSumAtrIX: Maybe the last item in the list can be a loading icon with a placeholder text or similar to show that theres a patch loading

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -215,24 +191,7 @@ class PatcherViewModel(
...
-
-                        if (state == State.COMPLETED && currentStepIndex != steps.lastIndex) {
-                            currentStepIndex++
-
-                            steps[currentStepIndex] =
-                                steps[currentStepIndex].copy(state = State.RUNNING)
-                        }
-                    }
-                }
+                onEvent = ::handleProgressEvent,
```
Axelen123: This should probably run on the main thread.

app/src/main/java/app/revanced/manager/patcher/ProgressEventParcel.kt
```diff
@@ -0,0 +1,105 @@
+package app.revanced.manager.patcher
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+
+@Parcelize
+data class ProgressEventParcel(
```
Axelen123: Why is this type necessary? The types in `PatcherProgress.kt` are already parcelable.

app/src/main/java/app/revanced/manager/patcher/PatcherProgress.kt
```diff
@@ -0,0 +1,81 @@
...
+
+@Parcelize
+data class RemoteError(
+    val type: String? = null,
+    val message: String? = null,
+    val stackTrace: String? = null,
+    val code: Int? = null,
+) : Parcelable
+
+fun Exception.toRemoteError(): RemoteError = RemoteError(
```
Axelen123:
```suggestion
fun Exception.toRemoteError() = RemoteError(
```

Type inference should work here.

app/src/main/java/app/revanced/manager/patcher/PatcherProgress.kt
```diff
@@ -0,0 +1,81 @@
...
+    data object WriteAPK : StepId()
+    @Parcelize
+    data object SignAPK : StepId()
+}
+
+@Parcelize
+data class RemoteError(
+    val type: String? = null,
+    val message: String? = null,
+    val stackTrace: String? = null,
```
Axelen123: Do all of these truly need to be nullable?

app/src/main/java/app/revanced/manager/patcher/PatcherProgress.kt
```diff
@@ -0,0 +1,81 @@
...
+    val code: Int? = null,
```
Axelen123: What is the point of the `code` field?

app/src/main/java/app/revanced/manager/patcher/ProgressEventParcel.kt
CnC-Robert: Because AIDL doesn’t support Kotlin sealed classes

Axelen123: Fair, but wouldn't this work then?:

```kt
@Parcelize
data class ProgressEventParcel(val event: ProgressEvent) : Parcelable
```
If it doesn't, use a `Bundle` and `Bundle.getParcelable`.

app/src/main/java/app/revanced/manager/patcher/PatcherProgress.kt
```diff
@@ -0,0 +1,80 @@
...
+sealed class ProgressEvent : Parcelable {
+    abstract val stepId: StepId?
+
+    @Parcelize
+    data class Started(override val stepId: StepId) : ProgressEvent()
+
+    @Parcelize
+    data class Progress(
+        override val stepId: StepId,
+        val done: Long? = null,
```
Axelen123: `current` would be a better name in my opinion. `done` sounds like the name for a boolean.

app/src/main/java/app/revanced/manager/patcher/ProgressEventParcel.kt
CnC-Robert: Yes the wrapper works, didn't know that was possible

app/src/main/java/app/revanced/manager/patcher/PatcherProgress.kt
```diff
@@ -0,0 +1,86 @@
...
+
+import android.os.Parcelable
+import kotlinx.parcelize.Parcelize
+
+
+@Parcelize
+sealed class ProgressEvent : Parcelable {
+    abstract val stepId: StepId?
+
+    @Parcelize
```
Axelen123: The annotation shouldn't be required for classes that inherit from a `sealed` class marked as `Parcelize` according to the documentation: https://developer.android.com/kotlin/parcelize#parcelize_with_sealed_classes_and_sealed_interfaces

app/src/main/java/app/revanced/manager/patcher/PatcherProgress.kt
```diff
@@ -0,0 +1,86 @@
...
+    data class Failed(
+        override val stepId: StepId?,
+        val error: RemoteError,
+    ) : ProgressEvent()
+}
+
+@Parcelize
+data class ProgressEventParcel(val event: ProgressEvent) : Parcelable
```
Axelen123, line 36: It would be good to add a comment explaining this class only exists because AIDL cannot directly use sealed classes.

Axelen123 (approved):
Looks good to me other than these two comments.

Issue #2793, opened 2025-10-03
bug:


KobeW50:
### Bug description

**Steps to reproduce**

1. Clear the app data of the Manager
2. Open the Manager and select YouTube, either from the installed app list or from storage

The Manager should now show that several dozen patches are selected. However, if you go to inspect the patch selection, only the `GmsCore support` will be selected. If you exit the patch selection screen, the Manager will now show that 1 patch is selected. 

https://github.com/user-attachments/assets/3f47792e-83bf-47c3-ad98-3e305eef76eb

Even if you save the patch selection as the single patch, and then go back to the app selector screen, and then return to patch YouTube, the Manager still sometimes says that 57 patches are selected.

__

This bug is really two issues:

1. When patching YouTube for the first time, the Manager will say "57 patches selected", when in reality only 1 patch is selected. This is a cosmetic issue.

2. All 57 default patches for YouTube should be selected, instead of just the `GmsCore support` patch, which is the only patch selected by default for some reason. This is a functional issue.

I believe this issue is specific to YouTube since it seems an unlikely coincidence that it is the `GmsCore support` patch is the one selected. The issue does not occur with YT Music or any other app that I tested. I didn't test other apps that use the `GmsCore support` patch. Perhaps it is partly because YouTube is also the first app in the App selection screen.

### Version of ReVanced Manager and version & name of app you are patching

1.26.0-dev.7

### Installation method

None

### ReVanced Manager logs

```shell
N
```

### Patch logs

```shell

```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

Ushie:
This also affects migrated list of patch selection, it might show 57 patches selected but in reality its only 3 patches migrated from flutter manager, meaning no user interaction is needed to be effected by this

KobeW50:
> This also affects migrated list of patch selection, it might show 57 patches selected but in reality its only 3 patches migrated from flutter manager, meaning no user interaction is needed to be effected by this

Wait, I think this bug is only in the case you described. I'm too busy (and lazy) to test though.

Axelen123:
@CnC-Robert did your PR resolve this?

CnC-Robert:
> @CnC-Robert did your PR resolve this?

It fixes the first of the mentioned issues

Axelen123:
[...]

PR #2896, opened 2025-12-26
fix: Update selected patch count when SelectionState changes

CnC-Robert:
Fixes cosmetic issue in #2793.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/SelectedAppInfoScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/SelectedAppInfoScreen.kt
@@ -79,7 +79,7 @@ fun SelectedAppInfoScreen(
     val allowIncompatiblePatches by vm.prefs.disablePatchVersionCompatCheck.getAsState()
     val patches by remember {
         derivedStateOf {
-            vm.selectionState.patches(bundles, allowIncompatiblePatches)
+            vm.getPatches(bundles, allowIncompatiblePatches)
         }
     }
     val selectedPatchCount = patches.values.sumOf { it.size }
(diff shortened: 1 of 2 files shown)


app/src/main/java/app/revanced/manager/ui/screen/SelectedAppInfoScreen.kt
```diff
@@ -76,12 +77,12 @@ fun SelectedAppInfoScreen(
...
 
     val allowIncompatiblePatches by vm.prefs.disablePatchVersionCompatCheck.getAsState()
-    val patches = remember(bundles, allowIncompatiblePatches) {
-        vm.getPatches(bundles, allowIncompatiblePatches)
-    }
-    val selectedPatchCount = remember(patches) {
-        patches.values.sumOf { it.size }
+    val patches by remember {
+        derivedStateOf {
+            vm.selectionState.patches(bundles, allowIncompatiblePatches)
```
Axelen123: Use `vm.getPatches` instead. That way, you don't have to expose the `selectionState` at all.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/screen/SelectedAppInfoScreen.kt
CnC-Robert: `vm.getPatches` returns a value instead of a state so it won't update in the UI.

Axelen123: `derivedStateOf` should be able to spot the dependency on `vm.selectionState` at runtime, since `vm.patches` uses it.

CnC-Robert: Yep you're right. Fixed it.

Issue #2729, opened 2025-09-07
bug: Install button does not reappear if Android native install dialog is closed


oSumAtrIX:
### Bug description

When the manager composes, the install dialog appears, the install button is hidden, and if I exit the app and reopen it, the install dialog is gone along with the install button, leaving me stuck.

### Version of ReVanced Manager and version & name of app you are patching

-

### Installation method

None

### ReVanced Manager logs

```shell
--
```

### Patch logs

```shell
-
```

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

Axelen123:
This should be fixed by #2900. Please reopen this issue along with a recording if the issue persists.

PR #2900, opened 2025-12-29
fix: install dialog getting stuck

Axelen123:
Android's package installer API behaves inconsistently across Android versions/devices, partly because of implementation bugs, likely causing issues such as #2729 . [The author of Ackpine even wrote a blog post about these quirks](https://github.com/ReVanced/revanced-manager/issues/2729). This PR should fix them.


oSumAtrIX:
What about not hiding the install button instead of introducing an entire custom lib. Note, install code from here will eventually have to be migrated to revanced library

Axelen123:
> What about not hiding the install button instead of introducing an entire custom lib.

One benefit of hiding the button is not having to deal with potentially concurrent or stuck installation attempts.

> Note, install code from here will eventually have to be migrated to revanced library

The work in this PR is still beneficial, since the installer in revanced library has a `suspend` function API just like Ackpine.
Note that revanced lib would now be responsible for dealing with all the quirks mentioned in the linked blog post, and should ideally support recovering installation sessions across process death.

oSumAtrIX:
Iirc there is a callback when installation is started, so you can disable it at this time, so that when the dialog is dismissed before installation is started, the button can reappear. fixes #2729.

> The work in this PR is still beneficial, since the installer in revanced library has a suspend function API just like Ackpine.
Note that revanced lib would now be responsible for dealing with all the quirks mentioned in the linked blog post, and should ideally support recovering installation sessions across process death.

Library API currently provides an install and uninstall method you can call, as well as callbacks for the install result. The rest can happen elsewhere.

PR #2913, opened 2026-01-03
feat: Add language settings

Ushie:
This implements Crowdin workflows and a language selector, Android 13+ will use the system language app manager for apps instead of the dialog

Screenshot_20260105_003951_ReVanced Manager (Debug)

### DIFF
--- a/.github/workflows/pull_strings.yml
+++ b/.github/workflows/pull_strings.yml
@@ -1,43 +1,43 @@
 name: Pull strings
 
 on:
-    schedule:
-        - cron: "0 */12 * * *"
-    workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * 0"
+  workflow_dispatch:
 
 jobs:
-    pull:
-        name: Pull strings
-        permissions:
-            contents: write
-            pull-requests: write
-        runs-on: ubuntu-latest
-        steps:
-            - name: Checkout
-              uses: actions/checkout@v5
-              with:
-                  ref: dev
-                  clean: true
+  pull:
+    name: Pull strings
+    permissions:
+      contents: write
+      pull-requests: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
+        with:
+          ref: dev
+          clean: true
 
-            - name: Pull strings
-              uses: crowdin/github-action@v2
-              with:
-                  config: crowdin.yml
-                  upload_sources: false
-                  download_translations: true
-                  skip_ref_checkout: true
-                  localization_branch_name: feat/translations
-                  create_pull_request: false
-              env:
-                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-                  CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
-                  CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
+      - name: Pull strings
+        uses: crowdin/github-action@v2
+        with:
+          config: crowdin.yml
+          upload_sources: false
+          download_translations: true
+          skip_ref_checkout: true
+          localization_branch_name: feat/translations
+          create_pull_request: false
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
+          CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
 
-            - name: Open pull request
-              if: github.event_name == 'workflow_dispatch'
-              uses: repo-sync/pull-request@v2
-              with:
-                  source_branch: feat/translations
-                  destination_branch: dev
-                  pr_title: "chore: Sync translations"
-                  pr_body: "Sync translations from [crowdin.com/project/revanced](https://crowdin.com/project/revanced)"
+      - name: Open pull request
+        if: github.event_name == 'workflow_dispatch'
+        uses: repo-sync/pull-request@v2
+        with:
+          source_branch: feat/translations
+          destination_branch: dev
+          pr_title: "chore: Sync translations"
+          pr_body: "Sync translations from [crowdin.com/project/revanced](https://crowdin.com/project/revanced)"
--- a/.github/workflows/push_strings.yml
+++ b/.github/workflows/push_strings.yml
@@ -1,26 +1,26 @@
 name: Push strings
 
 on:
-    workflow_dispatch:
-    push:
-        branches:
-            - dev
-        paths:
-            - app/src/main/res/values/strings.xml
+  workflow_dispatch:
+  push:
+    branches:
+      - dev
+    paths:
+      - app/src/main/res/values/strings.xml
 
 jobs:
-    push:
-        name: Push strings
-        runs-on: ubuntu-latest
-        steps:
-            - name: Checkout
-              uses: actions/checkout@v5
+  push:
+    name: Push strings
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v5
 
-            - name: Push strings
-              uses: crowdin/github-action@v2
-              with:
-                  config: crowdin.yml
-                  upload_sources: true
-              env:
-                  CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
-                  CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
+      - name: Push strings
+        uses: crowdin/github-action@v2
+        with:
+          config: crowdin.yml
+          upload_sources: true
+        env:
+          CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
+          CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -145,7 +145,6 @@ android {
         debug {
             applicationIdSuffix = ".debug"
             resValue("string", "app_name", "ReVanced Manager (Debug)")
-            isPseudoLocalesEnabled = true
 
             buildConfigField("long", "BUILD_ID", "${Random.nextLong()}L")
         }
@@ -276,7 +275,7 @@ tasks {
                 |object GeneratedLocales {
                 |    val SUPPORTED_LOCALES = listOf(
                 |        Locale.ENGLISH,
-                |        $locales
+                |        $locales,
                 |    )
                 |}
                 """.trimMargin()
--- a/app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
@@ -1,9 +1,6 @@
 package app.revanced.manager.ui.screen.settings
 
-import android.content.Intent
-import android.net.Uri
 import android.os.Build
-import android.provider.Settings
 import androidx.compose.animation.AnimatedVisibility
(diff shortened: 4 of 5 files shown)


.github/workflows/pull_strings.yml
validcube, line 1: [Consistency] Use indentation of 2 lines for GitHub yml/yaml files

.github/workflows/push_strings.yml
validcube, line 1: [Consistency] Use indentation of 2 lines for GitHub yml/yaml files

.github/workflows/pull_strings.yml
```diff
@@ -0,0 +1,43 @@
+name: Pull strings
+
+on:
+    schedule:
+        - cron: "0 */12 * * *"
```
validcube: [CI] Running workflow every 12 hours seems aggressive too, even for translation.
Perhaps try changing cron to every week or month (depending on activity), or before version release.

(Weekly)
```suggestion
        - cron: "0 0 * * 0"
```

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
```diff
@@ -38,23 +46,49 @@ import app.revanced.manager.ui.theme.Theme
...
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            context.startActivity(
+                Intent(Settings.ACTION_APP_LOCALE_SETTINGS).apply {
+                    data = "package:${context.packageName}".toUri()
+                }
+            )
+        } else {
+            showLanguagePicker = true
+        }
+    }
```
validcube: [UX/Thought] 🤔 We could show our own built-in language picker for Tiramisu because `LocaleManager` support syncing with the per-app language settings, I'm sure this is what Android encourage to have consistent UX in the app. (doesn't have to be from Android but I believe the consensus agreement would not sending user inconsistently to elsewhere depending on their API level)

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
```diff
@@ -148,4 +200,64 @@ private fun ThemePicker(
...
+            TextButton(
+                onClick = {
+                    onConfirm(selectedLocale)
+                    onDismiss()
+                }
+            ) {
+                Text(stringResource(R.string.apply))
+            }
+        }
+    )
```
validcube, line 244: [UX] Considering the amount of language we support for ReVanced Manager, it would be nice for the language picker to be (1) full screen; (2) have system default language at the top and a quick reset button at the title bar; (3) and is searchable with some kind of multilingual search (English + System default + language code)

app/build.gradle.kts
```diff
@@ -243,13 +239,56 @@ android {
...
+                |
+                |import java.util.Locale
+                |
+                |object GeneratedLocales {
+                |    val SUPPORTED_LOCALES = listOf(
+                |        Locale.ENGLISH,
+                |        $locales
+                |    )
+                |}
+                """.trimMargin()
```
validcube, line 281: [Code/Styling] Can we append the end of line comma to `$locales` variable? If not we might have to keep in mind about generated file (which we should) once we decided to implement formatter/linter.

app/build.gradle.kts
validcube, line 233: [Code/Thought] I assume you can't access the language data from Tiramisu res so you need a locale generator task right?

app/src/main/java/app/revanced/manager/util/SupportedLocales.kt
```diff
@@ -0,0 +1,34 @@
...
+    fun getSupportedLocales(context: Context): List<Locale> {
+        val locales = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            runCatching {
+                android.app.LocaleConfig(context).supportedLocales?.toList()
+            }.getOrNull() ?: GeneratedLocales.SUPPORTED_LOCALES
+        } else {
+            GeneratedLocales.SUPPORTED_LOCALES
+        }
+
+        return locales.filterNot { it.country.run { length == 2 && startsWith("X") } }
```
validcube: [Code] Filter pseudo locale `en-XA`, `en-XB` instead because this code would filter out Kosovo language too whose region start with X (`sq-XK`, `sr-Cyrl-XK`, `sr-Latn-XK` as supported by Android). 


[Code/Workaround] Since pseudo locale isn't enabled in release build but is enabled in debug, and debug build is broken we can safely add a if condition that check if the build is debug mode, if debug we check for pseudolocale, if not we assume no consequences by not filtering it.

app/src/main/java/app/revanced/manager/util/SupportedLocales.kt
```diff
@@ -0,0 +1,34 @@
...
+import android.os.LocaleList
+import androidx.appcompat.app.AppCompatDelegate
+import androidx.core.os.LocaleListCompat
+import java.util.Locale
+
+object SupportedLocales {
+    fun getSupportedLocales(context: Context): List<Locale> {
+        val locales = if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+            runCatching {
+                android.app.LocaleConfig(context).supportedLocales?.toList()
```
validcube, line 14: [Code/Thought] 🤔 Does this work? We don't have `locales_config.xml` right?

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
```diff
@@ -74,6 +108,24 @@ fun GeneralSettingsScreen(
...
+            SettingsListItem(
+                modifier = Modifier.clickable { openLanguageSettings() },
+                headlineContent = stringResource(R.string.language),
+                supportingContent = stringResource(R.string.language_description),
+                trailingContent = {
+                    FilledTonalButton(onClick = openLanguageSettings) {
+                        Text(
+                            currentLanguageDisplay
+                                ?: stringResource(R.string.language_system_default)
+                        )
```
validcube, line 106: [UX/Thought] 🤔 `language_system_default` is "System default", won't this cause the text to overflow on some devices with low dpi? We can call it "Default" instead or call it whether language the system default set to.

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
```diff
@@ -74,6 +108,24 @@ fun GeneralSettingsScreen(
...
             GroupHeader(stringResource(R.string.appearance))
 
+            val currentLocale = viewModel.getCurrentLocale()
+            val currentLanguageDisplay = remember(currentLocale) {
+                currentLocale?.let { viewModel.getLocaleDisplayName(it) }
+            }
+            SettingsListItem(
+                modifier = Modifier.clickable { openLanguageSettings() },
+                headlineContent = stringResource(R.string.language),
+                supportingContent = stringResource(R.string.language_description),
```
validcube, line 100: [UX/Thought] 🤔 I think just the headline "Language" is descriptive enough that it would not warrant a description.

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
validcube, line 244: [General] Having a UI preview for the dialog would be nice.

.github/workflows/pull_strings.yml
```diff
@@ -0,0 +1,43 @@
...
+                  config: crowdin.yml
+                  upload_sources: false
+                  download_translations: true
+                  skip_ref_checkout: true
+                  localization_branch_name: feat/translations
+                  create_pull_request: false
+              env:
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
+                  CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
```
validcube: [Repository] We don't have secret setup for Crowdin (now), notify the member that manage secret for ReVanced Manager.

.github/workflows/push_strings.yml
```diff
@@ -0,0 +1,26 @@
...
+              uses: actions/checkout@v5
+
+            - name: Push strings
+              uses: crowdin/github-action@v2
+              with:
+                  config: crowdin.yml
+                  upload_sources: true
+              env:
+                  CROWDIN_PROJECT_ID: ${{ secrets.CROWDIN_PROJECT_ID }}
+                  CROWDIN_PERSONAL_TOKEN: ${{ secrets.CROWDIN_PERSONAL_TOKEN }}
```
validcube: [Repository] We don't have secret setup for Crowdin (now), notify the member that manage secret for ReVanced Manager.

.github/workflows/pull_strings.yml
Ushie: These are org wide secrets

.github/workflows/push_strings.yml
Ushie: These are org wide secrets

.github/workflows/pull_strings.yml
oSumAtrIX: Week is good, this can also be changed in patches repo

validcube: > Week is good, this can also be changed in patches repo

I call dibs on changing cron for translation CI

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
Ushie: What about consistency with other settings? All of them have descriptions, i think this can be part of a bigger settings rehaul

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
Ushie: I think system default is standard, also otherwise which default? app default (english)? or system default? it's not any longer than language names like English (United States so I think its fine)

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
Ushie: Added to OP

Ushie: So System Default pinned to the top? like a sticky header? or just in the list?

app/build.gradle.kts
Ushie: Yea, it's not usable below Android 13 hence the generator, the XML can be read at runtime but I figured gradle task is cleaner

app/src/main/java/app/revanced/manager/util/SupportedLocales.kt
Ushie: I think I'll just remove that filter, pseudo locales can help us test the scalability of the UI in other languages

app/src/main/java/app/revanced/manager/util/SupportedLocales.kt
Ushie: We do 
image

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
validcube: Yep.

app/src/main/java/app/revanced/manager/ui/screen/settings/GeneralSettingsScreen.kt
validcube: Normally the list can be both descript and non-descript but if the rest of the settings is only descript intentionally, then I guess this is fine.

app/src/main/java/app/revanced/manager/util/SupportedLocales.kt
validcube: Ah right, I forgor (I coded that)

PR #2946, opened 2026-02-18
fix: flutter migration

janisslsm:


oSumAtrIX (changes requested):
Some changes need to be made:

- Business logic probably belongs somewhere else than in MainActivity. 
- During migration the keystore password is ideally rotated by the user (it shouldnt be hardcoded ideally)
- The old keystore file can be deleted including the old preferences
- Kotlin code can be made more idiomat (e.g. using .apply)

janisslsm:
> * Business logic probably belongs somewhere else than in MainActivity.

Would moving the whole logic inside MainViewModel::applyLegacySettings be appropriate?
> * During migration the keystore password is ideally rotated by the user (it shouldnt be hardcoded ideally)

I don't quite understand, should the user be prompted to input a keystore password?
> * The old keystore file can be deleted including the old preferences

Yeah, for the time being I kept it as is so the data isn't lost while testing.

Ushie:
> During migration the keystore password is ideally rotated by the user (it shouldnt be hardcoded ideally)

Is this really a problem considering the fact that it is only accessible by ReVanced Manager?

Axelen123:
> During migration the keystore password is ideally rotated by the user (it shouldnt be hardcoded ideally)

There is no point in doing that as part of a migration process. Also, the keystore is protected by filesystem permissions, so nothing else can access it.

oSumAtrIX:
> Would moving the whole logic inside MainViewModel::applyLegacySettings be appropriate?

This makes more sense to me but it would have to be confirmed by someone else

> I don't quite understand, should the user be prompted to input a keystore password?

This may fall outside the intended scope. Currently, the keystore passphrase is stored directly within ReVanced Manager; however, a passphrase should always remain user-dependent and must never be stored in plaintext. On Android, sensitive data is expected to be handled through the platform’s secure credential storage. Therefore, the user flow should mandatorily require explicit user authentication when signing an APK, either immediately before signing or earlier, such as at application launch or when patching begins.

From a technical perspective, the keystore itself would remain protected by its passphrase, while that passphrase is encrypted using a key derived from the platform keystore. Access to this key is granted only after the user unlocks it via system authentication (for example, fingerprint, PIN, or password). This effectively binds the keystore to a device-protected key that can be accessed solely through user authorization and not by the app itself.

oSumAtrIX:
> Is this really a problem considering the fact that it is only accessible by ReVanced Manager?

Yes, saving a passphrase effectively nullifies its purpose. Why have a keystore passphrase thats automatically unlocked when you can also just not have a passphrase.

oSumAtrIX:
> There is no point in doing that as part of a migration process. Also, the keystore is protected by filesystem permissions, so nothing else can access it.

Its only protected by policy. We could push an update to dump the key from the app proving its merely a policy. However secrets/passphrases and similar are by principle supposed to involve the user at least once at minimum. A technical example is encrypting the passphrase with the systems key derived from the system keystore. So practically, for example right before the app is signed, a system unlock dialog is shown which unlocks the keystore, cryptographically. The reason this is different from policy is because the fingerprint/pin/passphrase of the system keystore is used in cryptographic context. The key used to encrypt the passphrase cryptographically depends on the users input, not just by policy. 

But this is out of scope for this PR. Can be introduced later

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -81,24 +86,37 @@ class MainViewModel(
...
+                put(key.replace("flutter.", ""), value)
+            }
+
+            File(app.getExternalFilesDir(null), "/revanced-manager.keystore").apply {
+                if (exists()) {
+                    val keystoreBase64 = Base64.encodeToString(readBytes(), Base64.DEFAULT)
+                    put("keystore", keystoreBase64)
+                    delete()
+                }
+            }
```
Axelen123: The keystore could be imported directly using [`KeystoreManager`](https://github.com/ReVanced/revanced-manager/blob/dev/app/src/main/java/app/revanced/manager/domain/manager/KeystoreManager.kt) instead of base64 encoding and decoding it in `applyLegacySettings`.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -81,24 +86,37 @@ class MainViewModel(
...
+
+            File(app.filesDir.parentFile.absolutePath, "/app_flutter/selected-patches.json").apply {
+                if (exists())
+                {
+                    val patches = String(readBytes(), Charsets.UTF_8)
+                    put("patches", JSONObject(patches))
+                }
+            }
```
Axelen123: `selected-patches.json` can be decoded to `SerializedSelection` and passed to `applyLegacySettings` instead.

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -74,18 +72,15 @@ class MainActivity : AppCompatActivity() {
...
             val pureBlackTheme by vm.prefs.pureBlackTheme.getAsState()
 
             EventEffect(vm.legacyImportActivityFlow) {
-                try {
-                    launcher.launch(it)
-                } catch (_: ActivityNotFoundException) {
+                val flutterPrefs = getSharedPreferences("FlutterSharedPreferences", MODE_PRIVATE)
+                if (flutterPrefs.all.isNotEmpty())
+                {
+                    vm.applyLegacySettings(flutterPrefs)
```
Axelen123: It should be possible to get rid of the `EventEffect`and its associated `Channel` by moving this code into the `ViewModel`. shared prefs can be obtained using `app.getSharedPreferences(..)`.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -72,43 +74,55 @@ class MainViewModel(
...
-            app.toast(app.getString(R.string.legacy_import_failed))
-            Log.e(
-                tag,
-                "Got unknown result code while importing legacy settings: ${result.resultCode}"
-            )
-            return
+    fun applyLegacySettings(flutterPrefs: SharedPreferences) {
+        Log.d(tag, "Migrating flutter preferences")
+        val data = JSONObject().apply {
+            put("keystorePassword", "s3cur3p@ssw0rd")
```
Axelen123, line 87: The hardcoded password can be moved into `applyLegacySettings` and the field can be removed.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
```diff
@@ -144,20 +158,40 @@ class MainViewModel(
...
                 settings.keystorePassword,
                 keystoreBytes.inputStream()
             )
         }
-        settings.patches?.let { selection ->
+        patches?.let { selection ->
             patchSelectionRepository.import(0, selection)
         }
+        settings.patchedApps?.let { apps ->
+            val LIST_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu!"
```
Axelen123:
```suggestion
            const LIST_PREFIX = "VGhpcyBpcyB0aGUgcHJlZml4IGZvciBhIGxpc3Qu!"
```

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
janisslsm: The password is overridden if keystorePassword exists in the flutter preferences. I could check if keystorePassword doesn't exist and then use the hardcoded one otherwise use the one from the keystore

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
janisslsm: can't be made constant unless extracted outside the class, wasn't sure if it's something we want to do or not.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
Axelen123: Looks like I forgot about that.

app/src/main/java/app/revanced/manager/ui/viewmodel/MainViewModel.kt
Axelen123: It makes more sense as a constant to me and it is named like one. You can make it private at the top of the file.

janisslsm: done

Axelen123 (approved):
Thanks

PR #2949, opened 2026-02-18
feat: Better packaging rules

rushiiMachine:
Removes unnecessary junk and keeps files necessary for debugging in debug builds.
Final apk size is 17mb

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -217,23 +217,6 @@ android {
         includeInBundle = false
     }
 
-    packaging {
-        resources.excludes.addAll(
-            listOf(
-                "/prebuilt/**",
-                "META-INF/DEPENDENCIES",
-                "META-INF/**.version",
-                "DebugProbesKt.bin",
-                "kotlin-tooling-metadata.json",
-                "org/bouncycastle/pqc/**.properties",
-                "org/bouncycastle/x509/**.properties",
-            )
-        )
-        jniLibs {
-            useLegacyPackaging = true
-        }
-    }
-
     ksp {
         arg("room.schemaLocation", "$projectDir/schemas")
     }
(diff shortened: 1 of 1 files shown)


planshim (approved):

app/build.gradle.kts
```diff
@@ -254,6 +237,51 @@ android {
...
+
+            // Reflection symbol list (https://stackoverflow.com/a/41073782/13964629)
+            excludes += "/**/*.kotlin_builtins"
+        }
+        jniLibs {
+            // 32-bit x86 is dead
+            excludes += "/lib/x86/*.so"
+
+            // Equivalent of AndroidManifest's extractNativeLibs=true to ensure libs are compressed
+            useLegacyPackaging = true
```
oSumAtrIX, line 268: This field suggests legacy? Is speed not more desired here?

app/build.gradle.kts
```diff
@@ -254,6 +237,51 @@ android {
             version = "3.22.1"
         }
     }
+
+    packaging {
+        resources {
+            // Useless files
+            excludes += "/XPP3_*_VERSION"
+            excludes += "/font-awesome-license.txt"
```
oSumAtrIX, line 245: Is it allowed to remove licenses just like that

app/build.gradle.kts
rushiiMachine: Legacy = compress libraries in apk which is what was intended here. Only the necessary binaries for the target architecture gets extracted upon installation. Uncompressed+aligned packaging of native libraries is more suitable for split apks since the unused architectures won't end up being stored on the device.

Otherwise, all architectures are uncompressed, inflating the apk size, and will remain installed on the device uncompressed as well.

This property doesn't really have anything to do with speed

app/build.gradle.kts
rushiiMachine: You already ship license rules separately so I don't see a problem with this. It's not like a user will ever see this ever in the first place lmao

app/build.gradle.kts
oSumAtrIX: So this is just more suitable for regular apks and since manager is not a split apk its why we use this property? I was put off by the name "legacy". The reason why I still believe its due to speed is due to https://developer.android.com/guide/practices/page-sizes. If the binaries are uncompressed and aligned they can be directly mapped by the operating system using mmap which spares a copy to memory and therefore launch speed is faster. Does this property make much difference in size?

rushiiMachine: > So this is just more suitable for regular apks and since manager is not a split apk its why we use this property?]

Yes.

> I was put off by the name "legacy"

I dunno why they named it so poorly, the equivalent deprecated property in the manifest was "extractNativeLibs"

> If the binaries are uncompressed and aligned they can be directly mapped by the operating system using mmap which spares a copy to memory and therefore launch speed is faster. Does this property make much difference in size?

If the binaries are compressed/unaligned in the APK, they are extracted to the filesystem during installation time, and there is no loading performance penalty thereafter in either case. The main point of this property is to avoid extraction and save disk space on the device. I was gonna turn this on, but then I realized that you only end up saving disk space if the APK *only* ships binaries for the target architecture (as is the case for Google Play, where split apks is the standard). Otherwise, the rest of the binaries end up sitting in the APK uncompressed, and it uses even more space than just extracting one set of binaries for one architecture.

PR #2953, opened 2026-02-19
feat: Migrate option file selector to SAF

X1nto:
This gets rid of external file access requirements by moving the patcher options path selector to Storage Access Framework. Because of this change, it has become necessary to 1) separate folder and file selector actions and 2) copy selected folder/file to the app's internal storage. Internal storage refers to Filesystem.tempDir. 

androidx.documentfile was added to handle SAF and everything related to external storage permissions have been removed.

https://github.com/user-attachments/assets/bed1ca7e-44ca-4e27-a441-e0cdaae06e7e


oSumAtrIX:
If we exit the app does it remove these files?

Ushie:
Yes, it's stored in the same place as the temporary files generated for patcher

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
```diff
@@ -223,27 +230,69 @@ fun <T : Any> OptionItem(
...
-            PathSelectorDialog(
-                root = fs.externalFilesDir()
-            ) {
-                showFileDialog = false
-                it?.let { path ->
-                    fieldValue = path.toString()
+        val filePathLauncher = rememberLauncherForActivityResult(contract = ActivityResultContracts.OpenDocument()) {
+            it?.let { uri ->
+                fs.openFileDocument(uri)?.let { documentFile ->
+                    fieldValue = copyFile(documentFile)
```
Axelen123: `copyFile` shouldn't run on the main thread.

Axelen123 (changes requested):
The temporary files will be gone after an app restart and won't be present if manager attempts to use saved patch options. This is partly why I chose not to use SAF when implementing patch options.

app/src/main/java/app/revanced/manager/ui/component/patches/OptionFields.kt
X1nto: Ideally this'd be done in a ViewModel, but in this case I think rememberCoroutineScope() should work

gradle/libs.versions.toml
```diff
@@ -49,6 +50,7 @@ activity-compose = { group = "androidx.activity", name = "activity-compose", ver
 work-runtime-ktx = { group = "androidx.work", name = "work-runtime-ktx", version.ref = "work-runtime" }
 preferences-datastore = { group = "androidx.datastore", name = "datastore-preferences", version.ref = "preferences-datastore" }
 appcompat = { group = "androidx.appcompat", name = "appcompat", version.ref = "appcompat" }
+documentfile = { group = "androidx.documentfile", name = "documentfile", version.ref="documentfile" }
```
Axelen123:
```suggestion
documentfile = { group = "androidx.documentfile", name = "documentfile", version.ref  = "documentfile" }
```

Axelen123:
@X1nto app restarts still need to be handled, since the temporary directory will be cleared on restarts. To fix this, you could:
- Create a new data class for holding the SAF URI (and whether or not it is a folder). The data class probably needs to be `java.io.Serializable` and/or `Parcelable`.
- Change `StringOptionEditor` to `: OptionEditor<Any>`.
- Set it to a string whenever a string is specified and the custom type otherwise.
- In [Options.kt](https://github.com/ReVanced/revanced-manager/blob/dev/app/src/main/java/app/revanced/manager/data/room/options/Option.kt), define and implement a json object representation for the custom type whenever a patch option with the string type is encountered. The SAF URI will be valid even after a restart and can be persisted in the database.
- Move the SAF copying logic and transform the custom types into strings right before passing the options to `runtime.execute` in [PatcherWorker.kt](https://github.com/ReVanced/revanced-manager/blob/e0ce6b4ef370c3dc98644f295a34d0683155ddb8/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt#L232)

X1nto:
> @X1nto app restarts still need to be handled, since the temporary directory will be cleared on restarts. To fix this, you could:
> 
> * Create a new data class for holding the SAF URI (and whether or not it is a folder). The data class probably needs to be `java.io.Serializable` and/or `Parcelable`.
> * Change `StringOptionEditor` to `: OptionEditor<Any>`.
> * Set it to a string whenever a string is specified and the custom type otherwise.
> * In [Options.kt](https://github.com/ReVanced/revanced-manager/blob/dev/app/src/main/java/app/revanced/manager/data/room/options/Option.kt), define and implement a json object representation for the custom type whenever a patch option with the string type is encountered. The SAF URI will be valid even after a restart and can be persisted in the database.
> * Move the SAF copying logic and transform the custom types into strings right before passing the options to `runtime.execute` in [PatcherWorker.kt](https://github.com/ReVanced/revanced-manager/blob/e0ce6b4ef370c3dc98644f295a34d0683155ddb8/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt#L232)

TBF this would be much easier if patches themselves could use a different Option type for URIs. The only reason why copying files is even needed is because patches expect a string path instead of URIs. I'll take a look at your suggestion, but sounds like a lot of work.

oSumAtrIX:
> TBF this would be much easier if patches themselves could use a different Option type for URIs. The only reason why copying files is even needed is because patches expect a string path instead of URIs

The reason for this is the direction of dependency. The Manager depends on the Patcher, but not the other way around. Therefore, the Patcher cannot consider the requirements of the Manager and instead, it is the Manager that must take the Patcher into account. 

Options are specified with typed intentions for basic types such as string, int, long, boolean, and arrays of these types. These options are compatible with any platform and can be used as base types for arbitrary purposes. You can express any kind of option using them. There are no specific types to avoid unnecessary or unwanted specialization, as the API is designed to be generalized. This is why Options dont have "URI" type. Instead there is the string type, that can be used and serialize as URIs

PR #2956, opened 2026-02-21
feat: Add patch selection filters

### LINKED ISSUE #2955: feat: Add filters to patch selection screen
### Feature description

 (clicking should ask if for all apps or just the current app)

- Restore default selection (clicking asks for all patches or <patches source name>)
- Deselect all (clicking asks for all patches or <patches source name>)
- Invert selection (clicking asks for all patches or <patches source name>)
- Deselect all except <patches source name> patches

Additionally the tabs of patches can have a checkmark for quick selection/deselection. Unchecking unchecks all, while rechecking asks if for current app, for all and if the default selection should be selected instead of all patches (all is never good)

### Motivation

Self explanatory i hope lol

### Acknowledgements

Aunali321:
Closes #2955 

https://github.com/user-attachments/assets/b4534b6e-8259-40af-bd49-e0089c0c8b88

The existing restore defaults FAB (above the save button) may now be redundant since "Restore default selection" is available in the bottom sheet actions.  Should we remove it, or keep it for quick access? Note that the FAB also resets patch options, while the new action only resets selection.

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/PatchesSelectorScreen.kt
@@ -226,7 +226,7 @@ fun PatchesSelectorScreen(
                     onClick = {
                         guardedAction {
                             executeScopedAction { uid ->
-                                viewModel.restoreDefaults(bundles, uid)
+                                viewModel.restoreDefaults(uid)
                             }
                         }
                     }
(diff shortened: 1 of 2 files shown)


oSumAtrIX:
The popup currently asks for this vs all patches by using buttons to click. Perhaps a radio button is more fit?

We have an internal design branch, which would otherwise take care of things like this, so doesn't matter too much

oSumAtrIX:
The FAB should be removed. I think having multiple ways to do things in such close places adds unnecessary redundancy. Perhaps an orange • (dot) could be shown next to the filter button and next to the filter cards' reset to default button, when the defaults were changed, nagging the user to use the default selection, so the visibility question is resolved this way.

The patch options I am not sure how to solve. The filters are made for patches list. Perhaps clicking reset could show radio buttons for all/current patches/except this patches and a checkmark "also apply to options" 

@planshim @Axelen123 @Ushie

PR #2957, opened 2026-02-22
feat(PatcherScreen): Export more info in logs

Ushie:
- [ ] This is yet to be tested due to patches not loading during this stage in development
- [ ] Display patch selection/option changes

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
+++ b/app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
@@ -801,6 +801,3 @@ class PatcherViewModel(
         }
     }
 }
-
-
-


app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -327,18 +334,76 @@ class PatcherViewModel(
...
+        val patchesInfo = patchBundleRepository.sources.first()
+            .joinToString("\n") { source ->
+                val count = patchCounts[source.uid] ?: 0
+                "  - ${source.name}: ${source.version ?: "N/A"} ($count patches)"
+            }
+        val hasRoot = rootInstaller.hasRootAccess()
+        val suggestedVersion = patchBundleRepository.suggestedVersions.first()[packageName]
+
+        val deviceContent = """
+            - Device Info
```
oSumAtrIX: Reorder the details according their usefulness. Model for example is something that would be at the bottom while touched settings in manager and the selected patches and co would be at the top

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -327,18 +334,76 @@ class PatcherViewModel(
...
+
+        val sendIntent = Intent(Intent.ACTION_SEND).apply {
+            type = "text/plain"
             putExtra(
-                Intent.EXTRA_TEXT,
-                logs.asSequence().map { (level, msg) -> "[${level.name}]: $msg" }.joinToString("\n")
+                Intent.EXTRA_TEXT, "$deviceContent\n- Logs\n$logsContent"
             )
-            type = "text/plain"
+            addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
Axelen123: Is there a reason why this was added? Not having the flag is fine if the context comes from an `Activity`. The reason why `exportLogs` takes a context parameter instead of using `app` is precisely for this reason. If it works without the flag, it can be removed and Copilot's reviews here can be ignored.

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -327,18 +334,76 @@ class PatcherViewModel(
...
+        val deviceContent = """
+            - Device Info
+            Model: ${Build.MODEL}
+            Android version: ${Build.VERSION.RELEASE} (API ${Build.VERSION.SDK_INT})
+            Supported architectures: ${Build.SUPPORTED_ABIS.joinToString()}
+            Root permissions: ${if (hasRoot) "Yes" else "No"}
+            
+            RAM: ${Formatter.formatFileSize(context, memInfo.availMem)} / ${Formatter.formatFileSize(context, memInfo.totalMem)} available
+            Storage: ${Formatter.formatFileSize(context, statFs.availableBytes)} / ${Formatter.formatFileSize(context, statFs.totalBytes)} available
+            Memory limit: ${activityManager.memoryClass}MB (large: ${activityManager.largeMemoryClass}MB)
```
Axelen123: The process runtime settings should be shown as well if enabled, since they imply a larger memory limit. The prefs for them can be found [here](https://github.com/ReVanced/revanced-manager/blob/f19080a9862be4600860b67cd981b9a3a56b21e7/app/src/main/java/app/revanced/manager/domain/manager/PreferencesManager.kt#L17-L18)

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/PatcherViewModel.kt
```diff
@@ -642,4 +800,7 @@ class PatcherViewModel(
             )
         }
     }
-}
\ No newline at end of file
+}
+
+
+
```
oSumAtrIX:
```suggestion
}
```

PR #2961, opened 2026-02-25
feat: improve AppSelectorScreen caching/perf

rushiiMachine:
- Cache app list in the viewmodel to prevent constant reloading when navigating back to the screen
- Moves app list filtering into the viewmodel to be run async


rushiiMachine:
~~Unfortunately it seems that fetching patches is broken on dev branch for some reason, but I could get it to work on the private fork of manager~~

Can be tested by reverting https://github.com/ReVanced/revanced-manager/commit/8667051283f934a32ac7b7cc76178397dc45a0e6 temporarily

PR #2964, opened 2026-02-27
fix: Show previously downloaded APKs in source selector dialog

### LINKED ISSUE #2535: bug(Compose): No option to choose from file if the app was installed via plugin
### Bug description

I have installed yt music via apkmirror plugin. If I want to patch it again in a second moment, it will by default select the already downloaded file (correct and very cool).

Image
Image

Problem is that if I click on "Change source", I cannot see the downloaded file in the list.

Image

This does not happens if I select the apk from the storage, there I can see the file listed as a source.

Image

### Version of ReVanced Manager and version & name of app you are patching

Compose 

### Installation method

None

### ReVanced Manager logs

```shell
.
```

### Patch logs

```shell

```

### Acknowledgements

Aunali321:
What it does: Shows previously downloaded APKs in the "Change source" dialog. Closes #2535

Test steps:

1. Install the APK
2. Select an app to patch (e.g. YouTube), let it download the APK via a downloader
3. Patch and install it (so the download is saved in the DB)
4. Go to patch the same app again → tap "Select APK source"
5. In the dialog, between "Installed" and the downloader entries, you should see a "Downloaded" entry with the version number
6. Tap it. It should select that downloaded APK as the source
7. If there's a required version and the downloaded version doesn't match, the entry should appear greyed out

Screenshot_20260227_062741_ReVanced Manager (Debug)


oSumAtrIX:
"Downloaded" could be confusing "Already downloaded" or something similar would be better. Also maybe it should show up at the top of the list. Auto should also select an already downloaded apk instead of just auto choose one of the downloaders and thus the subtitle should be changed to accomodate that.

Aunali321:
@oSumAtrIX 

Screenshot_20260227_203812_ReVanced Manager (Debug)

Aunali321:
Auto already prefers a previously downloaded APK via `MainViewModel.findDownloadedApp` (PR #2967)

oSumAtrIX:
thx

PR #2966, opened 2026-02-27
feat: Show patch bundle version used to patch an app

### LINKED ISSUE #995: feat: Show what version of ReVanced Patches was used to patch an app within 'app info' for a patched app.
### Type

Functionality

### Issue

* Selecting 'applied patches' in 'App info' for a patched app, shows the names of the patches installed, but there's no way for a user to double-check the source of the patches/version of the patch source
* Also seeing as individual patch version numbers are apparently going to be removed, being able to check the version of ReVanced Patches (or alternative) used is the next best thing.

### Feature

* When patching an app, save the source of the patches + version number of the patch source somewhere
* Display this info in 'app info' via the manager
* OPTIONAL - inform a user if a newer version of the patch source exists

### Motivation

* Permits users to double-check what source they used to patch their app, and what version of that source they patched it with
* Allowing users to see this allows users whom need to report a bug with a particular patch to verify that they aren't using an outdated version of the patch
* Also could permit a reminder for a user to re-patch their app if they install a new Patches version

### Additional context

Pretty much the same overall request as #994 - but using the closest available thing to individual patch version numbers seeing as those are apparently going to get removed.

### Acknowledgements

- [X] I have searched the existing issues and this is a new and no duplicate or related to another open issue.
- [X] I filled out all of the requested information in this issue properly.
- [X] The issue is related solely to the ReVanced Manager

Aunali321:
What it does: Shows which patch bundle version was used when you patched an app. Closes #995                             
                                                            
Test steps:
1. Install the APK, open it
2. Patch any app (e.g. YouTube) — this saves the bundle info
3. Go to the Dashboard → tap the patched app → App info
4. Look below "Applied patches" — you should see a new "Patch bundle" row showing something like ReVanced Patches v5.x.x

Screenshot_20260227_064710_ReVanced Manager (Debug)


oSumAtrIX:
Multiple bundles/sources of patches can be used to patch an app, not just one, also it would be better to then simply group the list of patches by the source then in the "Viwe applied patches" tab.

Aunali321:
It shows all bundles that were used not just official one. I personally believe that it's not a good idea to move them to a sub-screen but I leave that design to you

Screenshot_20260227_080634_ReVanced Manager (Debug)

oSumAtrIX:
For the reasons:

1. Patches originate from bundles. The way its displayed here doesnt convey that
2. The semantic from where patches come from is lost
3. No duplicate line "Patch bundle"

Aunali321:
What about this?

Screenshot_20260227_195353_ReVanced Manager (Debug)

oSumAtrIX:
Pretty much, yup, works for now, design someone can take over from our team

PR #2968, opened 2026-02-27
feat: Improve changelogs

Ushie:
- [x] Switch to API instead of mock
- [x] Display all changelogs between current and latest version in update


Ushie:
> Display all changelogs between current and latest version in update

Decides its better to just show the full list regardless

PR #2992, opened 2026-03-10
Copilot/add code for new patches

samsquanch01:


Ushie:
ReVanced Manager flutter is no longer in use and will soon be replaced with Jetpack Compose

PR #3045, opened 2026-03-15
feat: Change source filepicker type for downloaders

secp192k1:
When picking a source both for `Patches` and `Downloaders`, it expects a binary file.
For the downloaders, it builds a `.apk` file, so this adds a check that explicitly changes the file type to expect an apk file


oSumAtrIX:
File handling via file extension is messy. The Java Gradle plugin can't link maven dependencies unless they end on .jar because of this reason, and revanced patches are .rvp. Downloaders would also be renamed to .rvd and now Manager doesn't recognize that

Ushie:
We're checking MIME, not file extension, if ReVanced Downloaders switch to RVD it'll still be recognized because internally it's an APK

PR #3111, opened 2026-03-18
fix: Mount install

omardotdev:
PatcherWorker seems to delete the input apk file right after patching finishes but this causes mounting to not work since it has to install the input apk first but it was already removed so it errors, i made it so it would check if manager is rooted so it would only remove the apk file if manager isn't granted root (the mount option wouldn't be there anyway)

### DIFF
--- a/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
@@ -266,7 +266,8 @@ class PatcherWorker(
             Result.failure()
         } finally {
             patchedApk.delete()
-            // Only delete the input APK when the user isn't rooted, since it would be needed for mounting
+            // Only delete the input APK right after patching finishes when the user isn't rooted, since it would be needed for mounting
+            // (it would be deleted right after installing with root)
             if (args.input is SelectedApp.Local && args.input.temporary && Shell.isAppGrantedRoot() == false) {
                 args.input.file.delete()
             }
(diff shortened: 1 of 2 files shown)


secp192k1:
I think it would be better to delete the file/s when the user leaves the patching screen instead of only deleting for rooted users, so you're not left with temporary files and with the possibility of this also happening to non-rooted users.

omaawr:
> I think it would be better to delete the file/s when the user leaves the patching screen instead of only deleting for rooted users, so you're not left with temporary files and with the possibility of this also happening to non-rooted users.

it already deletes the temp directory on startup, which is what the `onFreshProcessStart()` function on ManagerApplication.kt does

oSumAtrIX:
But then if you never reopen manager, then the file would stale there and consume space

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -265,7 +266,8 @@ class PatcherWorker(
             Result.failure()
         } finally {
             patchedApk.delete()
-            if (args.input is SelectedApp.Local && args.input.temporary) {
+            // Only delete the input APK when the user isn't rooted, since it would be needed for mounting
```
oSumAtrIX: Probably need to delete it after patching finishes then though

omaawr:
> But then if you never reopen manager, then the file would stale there and consume space

oh right im stupit

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
omaawr: i guess i'll make it so the file is deleted right after the app is installed?
if patching finishes and the file is deleted right after mounting wouldn't work since it would try to install the input file first

PR #3165, opened 2026-03-22
fix: Improve dialogs

PalmDevs:
This PR aims to improve some dialogs.

**Installation error dialog**: Icon removed since we can't center the description text. And when the user cancels the install, no dialog is shown.

image

**Add patches/downloader dialog**: Dialog title changed and description removed to reflect the actual purpose. Previously, the dialog title just said "Select".

image
image

**Change API URL**: Removed redundant "ReVanced Manager connects to ..." text.

image

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/screen/settings/DeveloperSettingsScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/settings/DeveloperSettingsScreen.kt
@@ -194,11 +194,6 @@ private fun APIUrlDialog(currentUrl: String, defaultUrl: String, onSubmit: (Stri
                 verticalArrangement = Arrangement.spacedBy(16.dp)
             ) {
                 Text(
-                    text = stringResource(R.string.api_url_dialog_description),
-                    style = MaterialTheme.typography.bodyMedium,
-                    color = MaterialTheme.colorScheme.onSurfaceVariant
-                )
-                Text(
                     text = stringResource(R.string.api_url_dialog_warning),
                     style = MaterialTheme.typography.bodyMedium,
                     color = MaterialTheme.colorScheme.error


PalmDevs:
> I believe there's a better way the dialog can be better without consolidating into one string

The dialog title is redundant.

> Currently, the dialog description doesn't show the reason as to _why_ it happens and _how_ to fix it.

The dialog description already tells the user why and how.

```xml
    <string name="installation_failed_description">There was a problem installing the app</string>
    <string name="installation_aborted_description">Installation was canceled</string>
    <string name="installation_blocked_description">Installation was blocked. Try adjusting the security settings of your device, and try again.</string>
    <string name="installation_conflict_description">An existing app is preventing the install. Uninstall the app, and try again.</string>
    <string name="installation_incompatible_description">This app isn’t compatible with your device. Use an APK that is compatible, and try again.</string>
    <string name="installation_invalid_description">There was a problem installing the app. Uninstall the app, and try again.</string>
    <string name="installation_storage_issue_description">There’s not enough storage space to install this app. Free up some space, and try again.</string>
    <string name="installation_timeout_description">The installation took too long</string>
```

Ushie:
I was supportive of this change initially but after seeing them side to side, I seen to like what we already have better

I get the reason at a glance, I don't need to read the description if I already understand what these mean, and as for the icon in the dialog, the addition of the dialog centers the title which more unnatural eye movement, there doesn't seem to be a direct benefit to either changes

PalmDevs:
> I get the reason at a glance, I don't need to read the description if I already understand what these mean.

Right, I think I'll revert those changes then.

> and as for the icon in the dialog, the addition of the dialog centers the title which more unnatural eye movement

I was going to center the description, but Compose seems very opinionated around having the description left aligned, I couldn't get it to work the way I wanted. So I suppose removing the icon is the best interest here.

PalmDevs:
What about Add patches/downloader?

Ushie:
Sure, those seem fine

PalmDevs:
I suppose redundant strings can be removed later.

PR #3185, opened 2026-03-23
feat: Show release dates and patch count in sources viewers

### LINKED ISSUE #3168: feat: Add the release date for patches
### Feature description

Please add the release date next to the available patches in the list. This feature was available in previous versions of the patcher but was removed in the latest release.

Image

Currently, I have to manually check the patch version before installation, which isn't as convenient as it used to be.

### Motivation

I previously posted this feature suggestion on Reddit, and it received over 200 upvotes and many comments. It seems that many people would also benefit from this functionality.

Image

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate

PalmDevs:
- Show version and release date at a glance (and in the respective info screens)

[Video](https://github.com/user-attachments/assets/811b1819-dee5-4ed4-acb5-a0aa7e566878)

Note that existing installs won't refetch every source again just to show the release date (they need to wait for the next releases). It's only a few lines of code to fix that, but I don't think it's worth the maintenance burden.

closes #3168

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -451,7 +451,6 @@ It’s only compatible with these versions: %2$s</string>
     <string name="license">License</string>
     <string name="source">Source</string>
     <string name="repository">Repository</string>
-    <string name="bundle_information">Bundle info</string>
     <string name="bundle_information_by_author">By %1$s</string>
     <string name="version">Version</string>
     <string name="version_compatibility_warning">Selected version may be incompatible with selected patches</string>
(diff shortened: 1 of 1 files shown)


Ushie:
Expanding arrows are usually on the right side

PalmDevs:
> Expanding arrows are usually on the right side

Yes, but info icon on the left may imply no action. Since tapping the whole list now expands the bundle. Having the arrow icon on the left is best here.

oSumAtrIX:
image

Filled icon is taking too much attention and would also appear on every patches source, simply replace with arrow icon, people will figure out by interaction automatically, no visual que needed

Metadata like last updated:

image

Should be here:

image

Same with this, simply keep it next to the version

image


Here:

image

PalmDevs:
Preview video updated.

PR #3213, opened 2026-03-25
feat: Improve patch options

### LINKED ISSUE #3029: bug: Can't enter "null" in custom branding
### Bug description

In the Custom Branding patch we cannot set the patch option to null anymore so that we can keep the original app icon. 

### Patch logs

```shell

```

### Debug logs

Version: 2.0.0 (200000099)
Build type: release
Model: Pixel 9a
Android version: 16 (36)
Supported Archs: arm64-v8a
Memory limit: 256MB (Normal) - 512MB (Large)

### Acknowledgements

PalmDevs:
closes #3029 

- New `chrome://flags` inspired UI for patch options
- Required fields are now annotated with red asterisks after name instead of red text under description
- Allows setting null on non-required options (by unchecking the checkbox)
- Allows viewing and using defaults and "presets" (specified by patches in Option.values)
  - Using defaults will remove the override and let the patcher handle default values.
- Empty string values can be passed into string list options
- List option validation works, and prevents the user from saving if not passing
- Required list options now have a warning if the user attempts to save with no items, but the user can still choose to continue
- Read-only mode on empty list items now has a placeholder text.

## Preview

https://github.com/user-attachments/assets/a219a573-fa44-4de3-933f-1eae553703b7

image

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/component/NumberInputDialog.kt
+++ b/app/src/main/java/app/revanced/manager/ui/component/NumberInputDialog.kt
@@ -1,6 +1,7 @@
 package app.revanced.manager.ui.component
 
 import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material3.AlertDialog
 import androidx.compose.material3.ButtonDefaults
 import androidx.compose.material3.ExperimentalMaterial3ExpressiveApi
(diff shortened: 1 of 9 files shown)


Ushie:
What's left for this to be complete?

PalmDevs:
Video updated

PalmDevs:
Additional content added after video:

image

Ushie (approved):
Somehow perfect?

Issue #3188, opened 2026-03-23
bug: Root mount is lost after reboot or removing ReVanced Manager


Lavenderatol:
### Bug description

When patching YouTube or YouTube Music with the latest version using the Root method, the app mounts and works without issues initially. However, the mount is not persistent.

If the device is rebooted, or if the ReVanced Manager is removed/uninstalled, the mount drops. The app either reverts to the unpatched stock version or stops working shortly after. After a reboot, opening the app launches the standard, unpatched YouTube/YT Music.

To make it work again, it is necessary to open ReVanced Manager and manually re-apply the mount every single time. It behaves as if the mount simply deactivates itself automatically.

Steps to reproduce:

Patch YouTube or YouTube Music using the Root method.
Mount the app (works normally).
Reboot the device OR uninstall ReVanced Manager.
Open the patched app (it reverts to the stock version or crashes).
Open Manager and manually activate the mount again to fix it.

Environment:

Device: Tecno Spark 20 Pro+
Target Apps: YouTube / YouTube Music
Install Method: Root

### Patch logs

```shell

```

### Debug logs

.

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

sideskroll:
Does it mount correctly for you? Meaning, when that happens (ever since v2 release) I mount the apps after a reboot, but I need to force stop them and then launch them again for the patched app to work otherwise it does nothing and the unpatched app stays unpatched (for lack of a better word).
The only thing I've managed to find is that in the logs of each magisk module (installed by revanced after patching by the way, which is new) it says "Not mounting as versions don't match".
Also, I used to (manually) update the photos app through play store since revanced doesn't require an specific version and patch it through revanced. It doesn't work like that anymore though, revanced wants me to download from apkmirror (even though the app is already updated) since it doesn't detect the versionto be honest, I don't know why they changed something that was already working :D

HubertsProjects:
> Does it mount correctly for you? Meaning, when that happens (ever since v2 release) I mount the apps after a reboot, but I need to force stop them and then launch them again for the patched app to work otherwise it does nothing and the unpatched app stays unpatched (for lack of a better word).
> The only thing I've managed to find is that in the logs of each magisk module (installed by revanced after patching by the way, which is new) it says "Not mounting as versions don't match".
[...]

PR #3223, opened 2026-03-26
fix: Incorrect version check during mounting

VarunS2002:
```
version="__VERSION__"
stock_version="$(dumpsys package "$package_name" | grep versionName | cut -d "=" -f2)"

if [ "$version" != "$stock_version" ]; then
  echo "Not mounting as versions don't match"
  exit 1
fi
```

This comparison is flawed as `stock_version` is actually ends up being multiple values not one single number so the equals comparison fails. This happens in the case of pre-installed apps such as Youtube who's package exists in /system and /data (latest update). So the version numbers for both are returned. This prevents the mounting.

A simple fix is to check if the patched version matches at least one of the two.
```
if ! echo "$stock_version" | grep -qwn "$version"
```

I directly modified `service.sh` in the installed magisk module and it fixed it for me.
This should fix multiple mounting related bug reports such as #3188 unless there are other flaws too.

### DIFF
--- a/app/src/main/assets/root/service.sh
+++ b/app/src/main/assets/root/service.sh
@@ -25,7 +25,7 @@ if mount | grep -q "$stock_path" ; then
   exit 1
 fi
 
-if [ "$version" != "$stock_version" ]; then
+if ! echo "$stock_version" | grep -qwn "$version"; then
   echo "Not mounting as versions don't match"
   exit 1
 fi


PalmDevs:
@validcube Can you test this?

VarunS2002:
Also, this is the log file where the `echo` is sent.  The stock version is in the format: 
`data_folder_version\nsystem_folder_version`

log:
```
base_path: /data/adb/modules/com.google.android.youtube-revanced/com.google.android.youtube.apk
stock_path: /data/app/~~xYXZjdi2p2ncbjt8WadnNg==/com.google.android.youtube-a36I29vERHyoEb_7_0Uxmg==/base.apk
base_version: 20.40.45
stock_version: 20.40.45
17.32.38
```

sideskroll:
> ```
> version="__VERSION__"
> stock_version="$(dumpsys package "$package_name" | grep versionName | cut -d "=" -f2)"
> 
> if [ "$version" != "$stock_version" ]; then
>   echo "Not mounting as versions don't match"
>   exit 1
> fi
> ```
> 
> This comparison is flawed as `stock_version` is actually ends up being multiple values not one single number so the equals comparison fails. This happens in the case of pre-installed apps such as Youtube who's package exists in /system and /data (latest update). So the version numbers for both are returned. This prevents the mounting.
> 
> A simple fix is to check if the patched version matches at least one of the two.
> 
> ```
> if ! echo "$stock_version" | grep -qwn "$version"
> ```
> 
> I directly modified `service.sh` in the installed magisk module and it fixed it for me. This should fix multiple mounting related bug reports such as #3188 unless there are other flaws too.

Would that fix the issue with the photos app also? (The manager doesn't detect the installed version unless it's updated from an external apk)
I usually update photos from play store and just patch that.

oSumAtrIX:
This might also be relevant in https://github.com/ReVanced/revanced-library/blob/main/library/src/commonMain/kotlin/app/revanced/library/installation/installer/Constants.kt#L40. Can you PR it there too? Manager will eventually switch to libraries impl

VarunS2002:
> > ```
> > version="__VERSION__"
> > stock_version="$(dumpsys package "$package_name" | grep versionName | cut -d "=" -f2)"
> > 
> > if [ "$version" != "$stock_version" ]; then
> >   echo "Not mounting as versions don't match"
> >   exit 1
> > fi
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This comparison is flawed as `stock_version` is actually ends up being multiple values not one single number so the equals comparison fails. This happens in the case of pre-installed apps such as Youtube who's package exists in /system and /data (latest update). So the version numbers for both are returned. This prevents the mounting.
> > A simple fix is to check if the patched version matches at least one of the two.
> > ```
> > if ! echo "$stock_version" | grep -qwn "$version"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I directly modified `service.sh` in the installed magisk module and it fixed it for me. This should fix multiple mounting related bug reports such as #3188 unless there are other flaws too.
> 
> Would that fix the issue with the photos app also? (The manager doesn't detect the installed version unless it's updated from an external apk) I usually update photos from play store and just patch that.

Probably not, this pr will only fix the mounting on the Magisk side (so what happens automatically on reboot). You might be facing a similar issue within the manager code I guess?

VarunS2002:
> This might also be relevant in https://github.com/ReVanced/revanced-library/blob/main/library/src/commonMain/kotlin/app/revanced/library/installation/installer/Constants.kt#L40. Can you PR it there too? Manager will eventually switch to libraries impl

Good call. I took a look at that script, but it doesn't seem to have a version check. If I understand it correctly, that will mount the patched apk regardless of installed version/s.

sideskroll:
> > > ```
> > > version="__VERSION__"
> > > stock_version="$(dumpsys package "$package_name" | grep versionName | cut -d "=" -f2)"
> > > 
> > > if [ "$version" != "$stock_version" ]; then
> > >   echo "Not mounting as versions don't match"
> > >   exit 1
> > > fi
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > This comparison is flawed as `stock_version` is actually ends up being multiple values not one single number so the equals comparison fails. This happens in the case of pre-installed apps such as Youtube who's package exists in /system and /data (latest update). So the version numbers for both are returned. This prevents the mounting.
> > > A simple fix is to check if the patched version matches at least one of the two.
> > > ```
> > > if ! echo "$stock_version" | grep -qwn "$version"
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I directly modified `service.sh` in the installed magisk module and it fixed it for me. This should fix multiple mounting related bug reports such as #3188 unless there are other flaws too.
> > 
> > 
> > Would that fix the issue with the photos app also? (The manager doesn't detect the installed version unless it's updated from an external apk) I usually update photos from play store and just patch that.
> 
> Probably not, this pr will only fix the mounting on the Magisk side (so what happens automatically on reboot). You might be facing a similar issue within the manager code I guess?

Yeah, didn't think so.... The issue seems to be that the manager can't detect the version for stock apps. Unless you first update it with a standalone apk. Doesn't matter if it's the same exact version....

maxulloa01:
I tried this fix by editing service.sh and rebooting. Still have the same issue

VarunS2002:
> I tried this fix by editing service.sh and rebooting. Still have the same issue

Can you upload the log file which would be in the same folder as service.sh. So `/data/adb/modules/com.google.android.youtube-revanced/log`

Upload both logs, the one before changing and one after making the change please.

Ushie:
This is going to be superseded by a switch to ReVanced Library for root install (this is already done locally, awaiting a fix in revanced library), if version checks are necessary for the functionality of the patched app, could you open a PR to ReVanced library? Thanks in advance!

If you're in the Discord server, you can ping me for the contributor role

c0des1ayr:
I found an easier fix. `stock_path` already only gets the first apk path, so all that needs to be done is to replace the `stock_version` variable with this: `stock_version="$(dumpsys package "$package_name" | grep versionName | cut -d "=" -f2 | head -n1)"`

Demigoddd:
I noticed your solution late, did some digging myself, and came to the same [conclusion](https://github.com/ReVanced/revanced-manager/issues/3188#issuecomment-4352018884).
I'd like to know if there's any progress on this issue ?

PR #3227, opened 2026-03-26
fix: Preserve applied patches on patched apps when source patches is deleted

### LINKED ISSUE #3167: bug: Applied patches disappear when bundle is removed
### Bug description

1. Add bundle XYZ
2. Patch any app with patches available on XYZ
3. Install / Mount app
4. Remove XYZ bundle
5. App info shows 'Applied patches' as zero

Example

Image

### Patch logs

```shell

```

### Debug logs

None

### Acknowledgements

Aunali321:


oSumAtrIX:
This will be interesting when we replace this screen with the patches viewer screen. I think how deletion would work in this case is that it will be defered until no referece to it exists, until then a deletion would simply hide it from ui/disable updates

Ushie:
Since its mostly database related I think sharing the screens shouldnt be a problem

oSumAtrIX:
The issue is when we delete a source of patches, the viewer wouldnt be able to load it. Hence why here defering would be needed or a mock of the patches or similar

Ushie:
I think the list of patches used is being stored in-database, not necessarily dependent on the patch bundle, because:

- The version of the patches used
- There's no guarantee the patches will remain on updates of the patches, in that case it would be no different to deleting the patches

Currently the implemention seems to already be storing it in the database, cossale says the section was just getting deleted with the patches source, probably they were connected to eachother(? i dont understand databases)

Ushie:
I think the viewer should be able to work with the minimal info

PR #3235, opened 2026-03-28
feat: Use fullscreen dialog for selectors

Ushie:
https://github.com/user-attachments/assets/207ccab7-247b-49da-9bb6-5d689eee5f1b

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -578,6 +578,7 @@ Tap them for more details.</string>
 
     <string name="sideeffect_restart">Restart the app to see changes</string>
     <string name="sideeffect_no_network">You’re offline. Check your internet connection.</string>
+    <string name="source_selector_category_downloaded">Previously downloaded</string>
 
     <plurals name="patch_count">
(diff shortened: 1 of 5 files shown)


PalmDevs:
Can the overlines be replaced with a single section header/label instead?

app/src/main/res/values/strings.xml
```diff
@@ -578,6 +578,7 @@ Tap them for more details.</string>
 
     <string name="sideeffect_restart">Restart the app to see changes</string>
     <string name="sideeffect_no_network">You’re offline. Check your internet connection.</string>
+    <string name="source_selector_category_downloaded">Previously downloaded</string>
```
PalmDevs, line 581: I think "previously" is misleading here. I assume once cache is cleared, the apps would disappear. Just say "Downloaded" maybe?

PalmDevs:
I don't like this at all:

image

Ushie:
Maybe we can make Version immutable when its Any, so basically you just cant reach that screen?

Afterall there's no purpose to

PR #3246, opened 2026-03-29
feat: Redesign SelectedAppInfoScreen

Ushie:
Based on #3235, continuation of #3241 made by @planshim 

<details>
<summary>Video preview, non-final design</summary>

https://github.com/user-attachments/assets/f953cb70-e389-4a37-8d9a-c20fd51744df

</details>

### DIFF
--- a/app/src/main/java/app/revanced/manager/MainActivity.kt
+++ b/app/src/main/java/app/revanced/manager/MainActivity.kt
@@ -264,6 +264,7 @@ private fun ReVancedManager(vm: MainViewModel) {
 
                 SelectedAppInfoScreen(
                     onBackClick = navController::popBackStackSafe,
+                    onSettingsClick = navController::navigateSafe,
                     onPatchClick = {
                         it.lifecycleScope.launch {
                             navController.navigateComplex(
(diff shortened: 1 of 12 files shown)


oSumAtrIX:
Can you add screenshots

Ushie:
Added a screen recording

PalmDevs:
Is this still active?

Ushie:
No, there's complications involved

oSumAtrIX:
The redesign is still relevant, which complications?

TUSHAR91316 (commented):
The redesign looks great! The blurred icon header and `BannerScaffold` integration are a huge visual improvement.

While reviewing the implementation, I noticed a few logic issues that need a fix:

1. **`compact?.let` bug in `Sections.kt`**:
   In `InfoListItem`, `compact` defaults to `false`. Because `false` is not `null`, `compact?.let { ... return@Surface }` always executes and returns early. This prevents the expanded `Column` and its action buttons (like "Open downloaders") from ever rendering.
   * **Fix:** Use `if (compact == true) {` instead of `compact?.let {`.

2. **Inverted warning conditions in `SelectedAppInfoScreen.kt`**:
   - `if (!hasModifiedPatchSelection)` shows the "Patch selection has changed" warning when the selection has *not* been modified.
   - `if (!showVersionCompatibilityWarning)` shows the compatibility warning when the version *is* compatible.
   * **Fix:** Remove the `!` negation from both conditions.

3. **Safeguard parameter in `SelectedAppInfoScreen.kt`**:
   In `onPatchSelectorClick`, `allowIncompatiblePatches` is passed instead of `effectiveAllowIncompatible` to `vm.getCustomPatches()`.

Issue #3034, opened 2026-03-15
bug: Error installing patched app inside Revanced Manager, but no issue installing saved APK


whodda:
### Bug description

When I patch YouTube in ReVanced Manager, I get to the end of the patching process, save the APK first, and then try to update in ReVanced Manager and get installation failed as per screenshot below. If I install the saved APK using CX File Explorer, no issues and it updates. Also, in ReVanced Manager YouTube still showing as an App that can be patched even though patched and installed (should be listed above with YT Music ReVanced as per screenshot).

**EDIT:** Issue present with ReVanced Manager version 2.0.1 as well, with the additional bug that the update to version 2.0.1 showed as failed initially, followed by the update notification again, then showing failed update error again but was installed. Attached additional debug file for 2.0.1.

<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/87bcdbe4-82e5-4edf-8a9e-3a9676d2625a" />
<img width="1080" height="2400" alt="Image" src="https://github.com/user-attachments/assets/3e12d77a-2699-41a8-8b49-709ea683092c" />

### Patch logs

```shell

```

### Debug logs

[revanced-manager_logcat_2026-03-15T11_29_35.451377.txt](https://github.com/user-attachments/files/26003698/revanced-manager_logcat_2026-03-15T11_29_35.451377.txt)

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

whodda:
Debug file for version 2.0.1

[revanced-manager_logcat_2026-03-15T14_47_22.103366.txt](https://github.com/user-attachments/files/26005134/revanced-manager_logcat_2026-03-15T14_47_22.103366.txt)

Axelen123:
This appears in the logs:
```
03-15 14:46:35.121 26618 26637 W System.err: com.android.apksig.internal.asn1.Asn1DecodingException: Failed to instantiate com.android.apksig.internal.x509.SubjectPublicKeyInfo
03-15 14:46:35.121 26618 26637 W System.err: 	at com.android.apksig.internal.asn1.Asn1BerParser.parseSequence(Asn1BerParser.java:225)
[...]

PR #3249, opened 2026-03-29
fix(Updater): Handle session death

### LINKED ISSUE #3091: bug: Unable to update ReVanced Manager
### Bug description

Image

Device: Pixel 9 pro xl 
System update: March 5th 2026

### Patch logs

```shell

```

### Debug logs

Attempted to start an adb session, but when going back to the manager the update installed? So I'm unable to get adb logs for now. 

### Acknowledgements

Ushie:
Fixes #3091

### DIFF
--- a/app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
+++ b/app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
@@ -218,7 +218,7 @@ class UpdateViewModel(
     override fun onCleared() {
         super.onCleared()
         if (installerSessionId == null) {
-            location.delete()
+            cancelUpdate()
         }
     }
 
--- a/app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
+++ b/app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
@@ -201,6 +201,20 @@ class UpdateViewModel(
         }
     }
 
+    var backPressedOnce by mutableStateOf(false)
+        private set
+
+    fun cancelUpdate() {
+        location.delete()
+    }
+
+    fun onBackPressed() {
+        if (!backPressedOnce) {
+            backPressedOnce = true
+            app.toast(R.string.press_back_again_to_cancel_update)
+        }
+    }
+
     override fun onCleared() {
         super.onCleared()
         if (installerSessionId == null) {
--- a/app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
+++ b/app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
@@ -26,6 +26,7 @@ import app.revanced.manager.network.api.ReVancedAPI
 import app.revanced.manager.network.dto.ReVancedAsset
 import app.revanced.manager.network.dto.ReVancedAssetHistory
 import app.revanced.manager.network.service.HttpService
+import app.revanced.manager.util.saveableVar
 import app.revanced.manager.util.toast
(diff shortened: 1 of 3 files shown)


app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
```diff
@@ -79,14 +84,46 @@ class UpdateViewModel(
 
     private val location = fs.tempDir.resolve("updater.apk")
 
+    private var installerSessionId: ParcelUuid? = savedStateHandle["installerSessionId"]
+        set(value) {
+            field = value
+            savedStateHandle["installerSessionId"] = value
+        }
```
Axelen123: No need for this. use `by savedStateHandle.saveableVar` like in `PatcherViewModel`.

app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
```diff
@@ -79,14 +84,46 @@ class UpdateViewModel(
 
     private val location = fs.tempDir.resolve("updater.apk")
```
Axelen123: `uiTempDir` is needed to implement this properly. You may also have to change the logic in `onCleared` as well. The file should be deleted in response to a confirmation that the user wants to leave the page and cancel the update instead.

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/ui/viewmodel/UpdateViewModel.kt
Ushie: > The file should be deleted in response to a confirmation that the user wants to leave the page and cancel the update instead

I'm not sure how to impl this, a cancelUpdate() that gets called by a dialog thats shown to the user before leaving?

Axelen123: You would just use a `BackHandler` which deletes the files. Patcher screen does something similar.

solrudev:
Hey, Ackpine maintainer here. I saw the issue which this PR tries to fix, and I'm not sure if this really could fix that. Was it tested and confirmed to work? It more seems like an edge case which should be handled in the library, and I'm happy to assist there.

Ushie:
Hey!

This PR fixed the issue consistently for those who have been struggling with it, the solution is also backed by the fact that the same users were able to follow through with the installation in another part of the app which wasn't missing the savedState, but failing this that was missing it

There haven't been any reports for this screen after this fix (which makes me assume that it is fixed)

There is still another open issue regarding ackpine sessions, though: https://github.com/ReVanced/revanced-manager/issues/3034 which is the ackpine session dying despite the session being saved and reloaded

solrudev:
I'm planning to implement detailed logging in the library, so that issues like this can be debugged more easily.

solrudev:
Hi there. Logging have been added to Ackpine [0.22.2](https://github.com/solrudev/Ackpine/releases/tag/0.22.2) and [0.22.3](https://github.com/solrudev/Ackpine/releases/tag/0.22.3). You can give it a try.

Axelen123:
Thank you. I will make a prerelease and try to debug the issue.

solrudev:
Hi! There is a "Session X is dead." failure happening also in [Morphe Manager](https://github.com/MorpheApp/morphe-manager/issues/449) after migration to Ackpine.

I've released a [0.22.7](https://github.com/solrudev/Ackpine/releases/tag/0.22.7) version which tries to fix the issue. After some tests by their users, it seems that the fix is working. The issue was caused by interplay of Play Protect and some library heuristics. You can try it.

PR #3279, opened 2026-04-06
fix: BusError when running 32b binary on armv8 with compatibility

Taknok:
Use the new binary of ReVanced/aapt2#32

The binaries provenance is attested by GitHub at https://github.com/ReVanced/aapt2/attestations. The artifacts are available at https://github.com/ReVanced/aapt2/releases/tag/v1.1.0.

Related issue: https://github.com/ReVanced/aapt2/issues/30


Taknok:
I didn't find the binary in the cli, the patcher or the library. Tell me the others repos that need the binary and I will create the requested PR to propagate the fix.

validcube (approved):
🥞 Approved (+2), iirc revanced manager is the only repo that uses libaapt2

PR #3285, opened 2026-04-07
feat: Magisk module install

secp192k1:
<details>
<summary>1. Successful patching</summary>

| New `Magisk module` option | Module installed, suggestion to restart |
|-|-|
| image | image |
</details>

<details>
<summary>2. After restart</summary>

| New module for patched the patched in your root manager app | 
|-|
| image |

</details>


oSumAtrIX:
Is a system restart really needed? Cant the magisk module be "invoked" somehow? Its just a mount command, so manager could also technically just invoke the mounting and magisk could handle reboots from there on.

secp192k1:
> Is a system restart really needed? Cant the magisk module be "invoked" somehow? Its just a mount command, so manager could also technically just invoke the mounting and magisk could handle reboots from there on.

Although Linux supports live file mounting, Magisk modules require a system reboot to safely inject code into the foundational Zygote process during startup, apply strict SELinux patches, and execute essential boot scripts without causing crashes - just like how Xposed/LSPosed work

https://topjohnwu.github.io/Magisk
https://topjohnwu.github.io/Magisk/details.html
https://topjohnwu.github.io/Magisk/guides.html

app/src/main/java/app/revanced/manager/ui/component/InstallerStatusDialog.kt
```diff
@@ -116,6 +116,13 @@ enum class DialogKind(
         contentStringResId = R.string.installation_timeout_description,
         confirmButton = installerStatusDialogButton(R.string.try_again) { it.install() },
         dismissButton = installerStatusDialogButton(R.string.cancel),
+    ),
+    SUCCESS_MAGISK(
+        flag = 1000,
+        title = R.string.magisk_install_success_title,
+        contentStringResId = R.string.magisk_install_success_description,
+        confirmButton = installerStatusDialogButton(R.string.restart_now) { it.reboot() },
+        dismissButton = installerStatusDialogButton(R.string.close),
```
Ushie, line 125: Later would be better

secp192k1: Made a lot of modifications to the library implementation, it should allow you to use the patched app without rebooting however it would be on very special occasions

Regardless, worst case scenario is the stock app not opening and patched app not showing until you reboot

oSumAtrIX:
Moved to https://github.com/ReVanced/revanced-manager/pull/3461

PR #3329, opened 2026-05-02
fix: isolate invalid patch bundle import failures

mostafaNazari702:
The cause of the bug was that a local patch file was dropped into the patch source folder and then picked up by the normal pipeline that loads all patch bundles together.
The problem started when an invalid file made it to the android patch loader. `Patch.androidKt.loadPatches` tried to read it as a dex file using `MultiDexIO.readDexFile` but it wasn’t actually a dex file, so it crashed before it could clearly identify where the bad input came from. After that, `PatchBundleRepository.loadMetadata` treated the whole batch as failed, marked all patch sources as broken, and returned no metadata at all. since that metadata is also used to decide which apps are compatible and should be shown, one bad file ended up making the whole default bundle look broken. which caused the entire "Apps" section to disappear until the invalid file was removed.

### DIFF
--- a/app/src/main/java/app/revanced/manager/patcher/patch/PatchBundle.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/patch/PatchBundle.kt
@@ -3,11 +3,13 @@ package app.revanced.manager.patcher.patch
 import kotlinx.parcelize.IgnoredOnParcel
 import android.os.Parcelable
 import app.revanced.patcher.patch.Patch
-import app.revanced.patcher.patch.loadPatches
+import app.revanced.patcher.patch.loadPatches as loadPatcherPatches
 import kotlinx.parcelize.Parcelize
 import java.io.File
 import java.io.IOException
 import java.util.jar.JarFile
+import java.util.zip.ZipException
+import java.util.zip.ZipFile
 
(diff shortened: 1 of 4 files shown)


app/src/main/java/app/revanced/manager/patcher/patch/PatchBundle.kt
```diff
@@ -55,18 +57,78 @@ data class PatchBundle(val patchesJar: String) : Parcelable {
...
         private fun patches(bundles: Iterable<PatchBundle>) = buildMap {
-            val bundleMap = bundles.associateBy { it.patchesJar }
+            bundles.forEach { bundle ->
+                this[bundle] = runCatching { bundle.loadPatchSet() }
+            }
+        }
+
+        private fun PatchBundle.loadPatchSet(): Set<Patch> {
+            val file = File(patchesJar)
+            file.requireDexContainer()
```
oSumAtrIX, line 68: Do we need explicit checks like this or should we just raise the original exceptions? @Axelen123

Axelen123:
If I remember correctly, this stuff was going to be implemented in patcher instead.

oSumAtrIX:
Yep and it was already, found in dev branch

Issue #3201, opened 2026-03-24
bug: Install button does not appear on Android 8 (API 26)


Swaien:
### Bug description

The patching process completes successfully, but the ReVanced Manager does not show an install button after building the APK.

This issue occurs on Android 8 (API 26) and affects multiple apps such as YouTube and Twitch. (Youtube Music worked for some reason though.)

The patched APK is successfully created and saved to:
/data/user/0/app.revanced.manager.flutter/app_ephemeral/patcher/result.apk

However, the Manager does not proceed with installation or provide any way to install the APK.

Using the manager:

- Patching completes without errors
- APK is aligned and signed
- No install prompt appears

![Image](https://github.com/user-attachments/assets/a5c76a81-a42a-4b01-a3b2-a92f4c107de4)

Using the CLI instead (same patches and APK) works perfectly:
- APK is created
- APK installs successfully when opened manually

This suggests the issue is related to the Manager v2 installation method on Android 8 rather than patching itself.

### Patch logs

```shell
App: tv.twitch.android.app (Selected: 16.9.1, Suggested: 16.9.1)
Source: ReVanced Patches
Added: Disable Sentry telemetry, Prevent screenshot detection, Remove screen capture restriction, Remove screenshot restriction, Spoof Play Age Signals
Manager version: 2.5.1
Manager auto-update: true
Allow changing patch selection: true
Show universal patches: false
Memory limit: 384MB (large: 512MB)
Root permissions: No
RAM: 1.31 GB / 3.97 GB available
Storage: 33.26 GB / 55.29 GB available
Android version: 8.0.0 (API 26)
Supported architectures: arm64-v8a, armeabi-v7a, armeabi
Model: RNE-L01

Deleting existing temporary files directory
Decoding manifest
Loading resource table from file: /data/user/0/app.revanced.manager.flutter/cache/framework/1.apk
Warning: Could not decode attr value, using undecoded value instead: ns=android, name=theme, value=0x7f150332
Warning: Could not decode attr value, using undecoded value instead: ns=android, name=label, value=0x7f1400d8
Warning: Could not decode attr value, using undecoded value instead: ns=android, name=icon, value=0x7f110000
Warning: Could not decode attr value, using undecoded value instead: ns=android, name=networkSecurityConfig, value=0x7f170001
Warning: Could not decode attr value, using undecoded value instead: ns=android, name=resource, value=0x7f1400ec
[...]

PR #3333, opened 2026-05-04
fix: install button not appearing on android 8

germanamv:
#3201 
* `NotificationChannel` is created before calling `setForeground`. If it isn't created, the notification won't show, and the "Foreground" status will be rejected.
* `WorkManager` automatically holds a `WakeLock` while `doWork()` is running. Manually acquiring one is usually unnecessary and can sometimes conflict with EMUI aggressive power management.

### DIFF
--- a/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
+++ b/app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
@@ -117,27 +117,31 @@ class PatcherWorker(
         }
 
         try {
-            // This does not always show up for some reason.
+            createNotificationChannel() // Safe to call multiple times
             setForeground(getForegroundInfo())
         } catch (e: Exception) {
-            Log.d(tag, "Failed to set foreground info:", e)
+            Log.e(tag, "Failed to set foreground info:", e)
+            // On Android 8, if this fails, the job is likely doomed
+            return Result.failure()
         }
-
-        val wakeLock: PowerManager.WakeLock =
-            (applicationContext.getSystemService(Context.POWER_SERVICE) as PowerManager)
-                .newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "$tag::Patcher")
-                .apply {
-                    acquire(10 * 60 * 1000L)
-                    Log.d(tag, "Acquired wakelock.")
-                }
 
         val args = workerRepository.claimInput(this)
 
         return try {
-            runPatcher(args)
-        } finally {
-            wakeLock.release()
+            runPatcher(args) // WorkManager holds its own WakeLock internally
+        } catch (e: Exception) {
+            Log.e(tag, "Patcher encountered an error", e)
+            Result.failure()
         }
+    }
+
+    private fun createNotificationChannel() {
+        val channel = NotificationChannel(
+            "patcher_channel", "Patcher Service",
+            NotificationManager.IMPORTANCE_LOW
+        )
+        val manager = applicationContext.getSystemService(NotificationManager::class.java)
+        manager.createNotificationChannel(channel)
     }
 
     private suspend fun runPatcher(args: Args): Result {


app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -117,29 +117,33 @@ class PatcherWorker(
...
-                    Log.d(tag, "Acquired wakelock.")
-                }
-
         val args = workerRepository.claimInput(this)
 
         return try {
-            runPatcher(args)
-        } finally {
-            wakeLock.release()
+            runPatcher(args) // WorkManager holds its own WakeLock internally
```
oSumAtrIX, line 131:
```suggestion
            createNotificationChannel() // Safe to call multiple times.
setForeground(getForegroundInfo())
} catch (e: Exception) {
            Log.e(tag, "Failed to set foreground info:", e)
            // On Android 8, if this fails, the job is likely doomed.
            return Result.failure()
}

val args = workerRepository.claimInput(this)

return try {
            runPatcher(args) // WorkManager holds its own WakeLock internally.
```

germanamv: What is the change here? The code looks exactly the same.

Axelen123: Yeah the suggestion does look kinda weird.

Axelen123 (approved):
Seems fine to me but I am not sure about the other suggestion.
Also, the PR title does not match the implemented code.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -117,29 +117,33 @@ class PatcherWorker(
         }
 
         try {
-            // This does not always show up for some reason.
+            createNotificationChannel() // Safe to call multiple times
```
kitadai31, line 120: Creating the notification channel here should be unnecessary, as it has already been done within the `createNotification()` method above that is called from `getForegroundInfo()`.

Also, if you have really confirmed that creating a notification channel twice here fixes the problem, use the same method within createNotification().
The ID of the notification channel created by the new method is different from the one that is actually being used.
This change will make a duplicate notification channel.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
kitadai31, line 131: iirc, WorkManager doesn't holds WakeLock
Try turning off the screen while patching without WakeLock
It should stop patching while the screen is off

kitadai31 (commented):
Overall, I have serious doubts as to whether this pull request will actually fix the issue.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
Axelen123: I assumed the statement in the comment was true. Is it documented anywhere?

kitadai31: See conversation in #2147
It is not documented, but some reports can be found on StackOverflow

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -119,37 +117,35 @@ class PatcherWorker(
...
```
TUSHAR91316, line 120: The channel ID `"patcher_channel"` used here is hardcoded and differs from the channel ID used in `createNotification()`. If a channel is needed before `setForeground`, we should reuse the existing channel ID / method to avoid creating a duplicate notification channel.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -119,37 +117,35 @@ class PatcherWorker(
...
         val args = workerRepository.claimInput(this)
 
         return try {
-            runPatcher(args)
-        } finally {
-            wakeLock.release()
+            runPatcher(args) // WorkManager holds its own WakeLock internally
+        } catch (e: Exception) {
+            Log.e(tag, "Patcher encountered an error", e)
+            Result.failure()
```
TUSHAR91316, line 134: Returning `Result.failure()` here will abort the entire patching operation if `setForeground()` throws. We should log the exception and allow patching to proceed instead of failing the job.

app/src/main/java/app/revanced/manager/patcher/worker/PatcherWorker.kt
```diff
@@ -119,37 +117,35 @@ class PatcherWorker(
...
-        @SuppressLint("WakelockTimeout")
-        val wakeLock: PowerManager.WakeLock =
-            (applicationContext.getSystemService(Context.POWER_SERVICE) as PowerManager)
-                .newWakeLock(PowerManager.PARTIAL_WAKE_LOCK, "$tag::Patcher")
-                .apply {
-                    // acquire without a timeout as we're managing the wakelock ourselves
-                    // in the finally-block, and therefore, it will be released regardless (of failure/success eventually).
-                    acquire()
-                    Log.d(tag, "Acquired wakelock.")
-                }
```
TUSHAR91316, line 139: Removing the `WakeLock` might cause the patching process to stop prematurely when the screen turns off (see PR #2147). WorkManager does not guarantee a CPU wakelock for `CoroutineWorker`.

TUSHAR91316 (commented):
A few observations regarding this PR:

1. **Notification Channel ID**: The new `createNotificationChannel()` creates a channel with the ID `"patcher_channel"`, which doesn't match the channel ID used in `createNotification()`. This will end up creating a duplicate unused channel.
2. **Aborting on Notification Failure**: Returning `Result.failure()` inside the `catch` block for `setForeground()` will cause the entire patch job to fail if the foreground notification fails. It might be better to let the patch process continue even if the notification fails to display.
3. **WakeLock**: `CoroutineWorker` in WorkManager doesn't hold a CPU wakelock automatically across the entire job duration. Removing the manual `WakeLock` could cause patching to freeze when the device screen turns off, especially on OEMs with aggressive battery management.

PR #3343, opened 2026-05-10
feat: Implement API fallback system

mostafaNazari702:
Implemented an API client layer that retries failed requests up to three times and automatically fails over to a backup endpoint ( hardcoded in the code ) when the primary is unavailable. A file system cache provides limited functionality in fully offline/degraded mode. session fail-over state persists across app restarts with an optional non-blocking UI prompt to restore primary endpoint usage once it recovers.


Ushie (changes requested):
Why is ReVanced API a repository now when it previously wasnt? And why is that not reflected in the package? besides we already have a ReVancedAPI file so why did you make a new one

What is the need of all these new files?

The cache implementation seems too hacky, don't we use something like Ktor already? Why is cache manual?

If you're going to be using agentic tools for your contributions, the bare minimum is expected, leaving large amounts of duplicate dead code and such hacky implementations is unacceptable

mostafaNazari702:
> Why is ReVanced API a repository now when it previously wasnt? And why is that not reflected in the package? besides we already have a ReVancedAPI file so why did you make a new one
> 
> What is the need of all these new files?
> 
> The cache implementation seems too hacky, don't we use something like Koin already? Why is cache manual?
> 
> If you're going to be using agentic tools for your contributions, the bare minimum is expected, leaving large amounts of duplicate dead code and such hacky implementations is unacceptable

I only used it for file structuring because when i had it integrated into the exsiting functions the files became in thousands of lines ( which i remember getting cooked for it in revanced-website migration ). Besides, this is not it just so you know. I have like..1 week for refinements that i will look at sometime !

for the naming, i put it in network/api because its more of an HTTP cordinator than a normal domain repo. It doesnt really work like the other ones that sit on top of "Room" or "DataStore" and keeping the retry stuff close to HttpService felt cleaner to me. But if you’d rather keep everything under domain/repository/ for consistency ( That specific aspect i was also scrutinized for by no one exept you a long time ago ), I can move it and rename the class, should **Axel or Ohan** want!
My recommendation to you is that you start some Kotlin course on codeacademy, there you will learn some stuff and also see what i done here, the pattern, i mean 😊.

If you read the pdf file in the chat that i had then you will see why the caching was manual, it just requires some bravery to read it :)

rushiiMachine:
There is no reason to write manual http caching when Ktor has a caching plugin that is well tested and functions fine

mostafaNazari702:
> There is no reason to write manual http caching when Ktor has a caching plugin that is well tested and functions fine

Great. Say that to both Axel and OsumAtrix.

Ushie:
osumatrix and axel are not the sole authorative members of revanced, infact regarding revanced manager osumatrix has no authority regarding the codebase, so fix your attitude

regarding the "pdf", not a single part of it implies custom manual caching

image

If I'm mistaken, then feel free to respond with it

the same pdf that *you drafted*, also says the following:

image
image

which applies to both issues I raised regarding the cache and the file structure

> My recommendation to you is that you start some Kotlin course on codeacademy, there you will learn some stuff and also see what i done here, the pattern, i mean 😊.

Justifications for such things are expected, comments like that going forward isn't going to be tolerated

mostafaNazari702:
> osumatrix and axel are not the sole authorative members of revanced, infact regarding revanced manager osumatrix has no authority regarding the codebase, so fix your attitude
> 
> regarding the "pdf", not a single part of it implies custom manual caching
> image
> 
> If I'm mistaken, then feel free to respond with it
> 
> the same pdf that _you drafted_, also says the following:
> image image
> 
> which applies to both issues I raised regarding the cache and the file structure
> 
> > My recommendation to you is that you start some Kotlin course on codeacademy, there you will learn some stuff and also see what i done here, the pattern, i mean 😊.
> 
> Justifications for such things are expected, comments like that going forward isn't going to be tolerated

Thanks for the info. It was conversational, not on paper. The chat got long but it should still be there.

mostafaNazari702:
- session-start endpoint state, `getPatchesHistory rerouting` + cache key, TTL contradicting the spec, probe error-handling structure, god-class
- ReVancedAPI.kt: rm
- // Unit test ( scope creep that i will have to handle later )

Axelen123:
I will admit that I missed we were using OkHttp cache (which is apparently broken) instead of the Ktor plugin (which should be tested). That was a mistake on my part, but if that cache is insufficient then caching in a wrapper repository is still the correct implementation. There is no requirement to write unit tests here. Nothing in manager is currently unit tested.

mostafaNazari702:
> I will admit that I missed we were using OkHttp cache (which is apparently broken) instead of the Ktor plugin (which should be tested). That was a mistake on my part, but if that cache is insufficient then caching in a wrapper repository is still the correct implementation. There is no requirement to write unit tests here. Nothing in manager is currently unit tested.

Huh? No, the tests are only for me and my implementation, when i am done and want to submit i will remove it.

Axelen123:
Oh. I thought you were complaining about having to add them.

mostafaNazari702:
> Oh. I thought you were complaining about having to add them.

That comment that has three points is just a reminder for myself.Not a message to anyone.

mostafaNazari702:
I truly hope that i did not misinterpret "infront of ReVancedAPI".
{96DAB04E-DAA3-4DA5-8122-52EF6DD93721}

app/build.gradle.kts
```diff
@@ -119,6 +119,10 @@ dependencies {
     // Ackpine
     implementation(libs.ackpine.core)
     implementation(libs.ackpine.ktx)
+
+    // Unit test
```
mostafaNazari702: Was:
"// Unit test ( scope creep that i will have to handle later )"

but now: "// Unit test".

That does NOT mean it will stay / is part of this push, it will be removed.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -9,56 +8,110 @@ import app.revanced.manager.network.dto.ReVancedAssetHistory
...
+    suspend fun getPatchesUpdate(): APIResponse<ReVancedAsset> =
+        request("patches${prefs.usePatchesPrereleases.prereleaseString()}")
 
-    suspend fun getAnnouncements() = request<List<ReVancedAnnouncement>>("announcements")
+    suspend fun getPatchesHistory(
+        apiUrl: String,
+        prerelease: Boolean,
+    ): APIResponse<List<ReVancedAssetHistory>> {
+        val route = "patches/history${prerelease.prereleaseString()}"
+        val normalized = apiUrl.trimEnd('/')
```
oSumAtrIX: Why is this being done here out of nowhere?

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -9,56 +8,110 @@ import app.revanced.manager.network.dto.ReVancedAssetHistory
...
+    suspend fun getInfo(): APIResponse<ReVancedInfo> {
+        val (response, servedBy) = requestTracked<ReVancedInfo>("about")
+        if (response is APIResponse.Success && servedBy == EndpointState.ActiveEndpoint.PRIMARY) {
+            endpointState.updateFallbackFromAbout(response.data.api?.fallback)
+        }
+        return response
+    }
 
-    suspend fun getPatchesUpdate() = request<ReVancedAsset>("patches${prefs.usePatchesPrereleases.prereleaseString()}")
+    suspend fun probePrimary(): Boolean = withContext(Dispatchers.IO) {
```
oSumAtrIX: But is the fallback system supporting nested fallbacks? 

Primary -> Fallback A -> Fallback B

If currently on B, it should probe A. if on A, it should probe Primary.
If on B, but A down and Primary online, it should be able to restore to Primary, even though A is down.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
mostafaNazari702: getPatchesHistory is used for patch-bundle sources which can be either the official RV source or a custom remote added by the user. the fallback logic should only apply to the official source, if someone adds a third-party source it should always talk directly to that server and never get redirected to RV's backup. that's why there's an apiUrl == primaryUrl() check there.

i understand that using a URL comparison to figure out "is this the official source?" isn't ideal. i would rather make that explicit at the call site (maybe with a flag or a separate method for the official source) instead of relying on string matching here. i will add a comment for now and can refactor it to be more explicit.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
mostafaNazari702: the system is intentionally just primary + one backup which is the scope we agreed on. right now /about only exposes a single fallback URL so supporting something like Primary => A => B would need API changes as well before we could even represent that chain.

on the main concern though , restoring back to primary even if an intermediate fallback is down, that's already how it works. probePrimary() talks directly to the primary endpoint and recovery only depends on whether the primary is reachable. The backup's health isn't part of that decision. so if we're currently on the backup and the primary comes back online, we will show the restore prompt regardless of what's happening with the backup.

he re-check happens on app launch (per the spec), not continuously during a session so we don't interrupt someone while they're using the app.

if we ever want true N-level fallback chains the implementation already routes through an ordered endpoints() list, so extending it wouldn't be difficult. since that's outside the current scope and would also require /about changes, I'd rather treat it as a separate discussion than fold it into this PR.

I hope this clears it for you.

mostafaNazari702:
removed dead retry plugin and added a clarifying comment *

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
Axelen123: Single backup is all we need in my opinion. The way to increase reliability would be sever side measures.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -9,56 +8,112 @@ import app.revanced.manager.network.dto.ReVancedAssetHistory
...
+    ): APIResponse<List<ReVancedAssetHistory>> {
+        val route = "patches/history${prerelease.prereleaseString()}"
+        val normalized = apiUrl.trimEnd('/')
+        // Only the official source (the configured primary endpoint) participates in the primary => fallback logic.
+        // A user-added custom remote source must query its own server directly and must never be redirected to ReVanced's backup endpoint.
+        return if (normalized == endpointState.primaryUrl()) {
+            request(route)
+        } else {
+            directRequest(normalized, route)
+        }
```
Axelen123: This is a nitpick, but I believe this should be extracted to a private function.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
oSumAtrIX: The fallback can be down too, the functionality has already been discussed on our discord

oSumAtrIX: > the system is intentionally just primary + one backup which is the scope we agreed on. right now /about only exposes a single fallback URL so supporting something like Primary => A => B would need API changes as well before we could even represent that chain.
> 
> on the main concern though , restoring back to primary even if an intermediate fallback is down, that's already how it works. probePrimary() talks directly to the primary endpoint and recovery only depends on whether the primary is reachable. The backup's health isn't part of that decision. so if we're currently on the backup and the primary comes back online, we will show the restore prompt regardless of what's happening with the backup.
> 
> he re-check happens on app launch (per the spec), not continuously during a session so we don't interrupt someone while they're using the app.
> 
> if we ever want true N-level fallback chains the implementation already routes through an ordered endpoints() list, so extending it wouldn't be difficult. since that's outside the current scope and would also require /about changes, I'd rather treat it as a separate discussion than fold it into this PR.
> 
> I hope this clears it for you.

The N chain fallback was discussed as in scope of the PR for the same reason as mentioned in my previous comment. It's essential for highest availability. Otherwise if you just probe primary, backup A could be on, but primary and B off, and it would never recover

app/src/main/java/app/revanced/manager/ui/viewmodel/DashboardViewModel.kt
```diff
@@ -43,6 +48,16 @@ class DashboardViewModel(
     val hasUpdate = managerUpdateRepository.hasUpdate
     val updateVersion = managerUpdateRepository.version
 
+    private val patchingWorkActive =
+        WorkManager.getInstance(app)
+            .getWorkInfosForUniqueWorkFlow(PATCHING_WORK_NAME)
+            .map { infos -> infos.any { !it.state.isFinished } }
```
Axelen123: Patching is always cancelled when you leave the screen.

app/build.gradle.kts
```diff
@@ -227,6 +227,10 @@ android {
         buildConfig = true
     }
 
+    testOptions {
+        unitTests.isReturnDefaultValues = true
+    }
```
Axelen123: This seems like a remnant of the tests you mentioned.

app/src/main/java/app/revanced/manager/ui/viewmodel/DashboardViewModel.kt
```diff
@@ -132,4 +147,17 @@ class DashboardViewModel(
...
     }
+
+    fun dismissPrimaryRecoveryPrompt() = endpointState.dismissPrimaryRecoveryPrompt()
+    fun restartApp() {
+        val intent = app.packageManager.getLaunchIntentForPackage(app.packageName)
+            ?.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK or Intent.FLAG_ACTIVITY_CLEAR_TASK)
+            ?: return
+        app.startActivity(intent)
+        Process.killProcess(Process.myPid())
+    }
```
Axelen123: Can't you just call `endpointState.signalPrimaryRecoveryDetected()` instead?

Axelen123 (changes requested):

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
mostafaNazari702: > The fallback can be down too, the functionality has already been discussed on our discord

Not to go against you but i don't remember that. Not to imply that we did not have that convo but i have put a ton of time into the website and forgot. I do however remember you telling me that:
Primary down? => Backup. 
Backup down as well AND we have cached data? => We go to offline mode and use the cache instead so functionality persists even in offline mode/unresponsive API mode.
No cached data and primary/backup down? => No choice but to tell the user "Not available" but we of course also check whether it is true endpoint unavailability or just user being offline for whatever reason.

app/src/main/java/app/revanced/manager/ui/viewmodel/DashboardViewModel.kt
mostafaNazari702: endpointState.primaryRecoveryAvailable should do it as beautifully right?

app/src/main/java/app/revanced/manager/di/HttpModule.kt
```diff
@@ -39,6 +41,11 @@ val httpModule = module {
         install(ContentNegotiation) {
             json(json)
         }
+        install(HttpCache) {
+            publicStorage(
+                FileStorage(context.cacheDir.resolve("api_cache").also { it.mkdirs() })
```
oSumAtrIX:
```suggestion
                FileStorage(context.cacheDir.resolve("api_cache").apply { mkdirs() })
```

app/src/main/java/app/revanced/manager/ui/viewmodel/DashboardViewModel.kt
Axelen123: I don't know exactly what you have in mind, but my point is that if you need to do something after recovering, you should do that immediately instead of restarting the app.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
oSumAtrIX: See https://discord.com/channels/952946952348270622/1485653081038393415/1501176573100036167 for the explanation on the functionality

mostafaNazari702:
Dropped the recovery notification card, restart dialog, restartApp(), the
related strings, and the now-unused lastSessionUsedFallback persistence and
prompt state, basically ALL restart related stuff!

when a session is on the backup a background loop probes the primary every 5 minutes and switches back silently once it is reachable.

app/src/main/java/app/revanced/manager/di/HttpModule.kt
```diff
@@ -39,6 +41,11 @@ val httpModule = module {
         install(ContentNegotiation) {
             json(json)
         }
+        install(HttpCache) {
+            publicStorage(
+                FileStorage(context.cacheDir.resolve("api_cache").apply { mkdirs() })
```
oSumAtrIX:
```suggestion
                FileStorage(context.cacheDir.resolve("http_cache").apply { mkdirs() })
```

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -9,56 +8,119 @@ import app.revanced.manager.network.dto.ReVancedAssetHistory
...
+    }
 
-    suspend fun getInfo() = request<ReVancedInfo>("about")
+    private suspend inline fun <reified T> requestTracked(
+        route: String,
+    ): Pair<APIResponse<T>, EndpointState.ActiveEndpoint?> = withContext(Dispatchers.IO) {
+        var lastFailure: APIResponse<T>? = null
+        for ((endpoint, baseUrl) in endpointState.endpoints()) {
```
oSumAtrIX: use .forEach

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -9,56 +8,119 @@ import app.revanced.manager.network.dto.ReVancedAssetHistory
...
+    /**
+     * Routes a request for a patch-bundle source. Only the official source (the configured
+     * primary endpoint) participates in the primary -> fallback logic; a user-added custom
+     * remote source queries its own server directly and is never redirected to the backup.
+     */
+    private suspend inline fun <reified T> requestForSource(
+        apiUrl: String,
+        route: String,
+    ): APIResponse<T> {
+        val normalized = apiUrl.trimEnd('/')
```
oSumAtrIX: Why is normalization needed

app/src/main/java/app/revanced/manager/ManagerApplication.kt
```diff
@@ -88,6 +93,19 @@ class ManagerApplication : Application() {
         scope.launch(Dispatchers.Default) {
             downloadedAppsRepository.cleanUp()
         }
+
+        // While the session is on the backup endpoint, periodically probe the primary and switch back to it silently as soon as it is reachable again.
+        // Endpoint URLs are resolved so the switch takes effect without restarting the app.
+        scope.launch(Dispatchers.IO) { 
+            while (true) {
```
oSumAtrIX: This is an infinite loop even if on primary

app/src/main/java/app/revanced/manager/ManagerApplication.kt
```diff
@@ -118,4 +136,8 @@ class ManagerApplication : Application() {
             mkdirs()
         }
     }
+
+    private companion object {
+        const val PRIMARY_RECONNECT_INTERVAL_MS = 5 * 60 * 1000L
```
oSumAtrIX:
```suggestion
        const val PRIMARY_RECONNECT_INTERVAL_MS = 5 * 60 * 1000L // 5min.
```

app/src/main/java/app/revanced/manager/ManagerApplication.kt
```diff
@@ -88,6 +93,19 @@ class ManagerApplication : Application() {
         scope.launch(Dispatchers.Default) {
             downloadedAppsRepository.cleanUp()
         }
+
+        // While the session is on the backup endpoint, periodically probe the primary and switch back to it silently as soon as it is reachable again.
+        // Endpoint URLs are resolved so the switch takes effect without restarting the app.
```
oSumAtrIX: What happens in the case of an update check?

app/src/main/java/app/revanced/manager/ManagerApplication.kt
PalmDevs: It's also probably better off as a proper Duration object. Kotlin is deprecating `delay(ms: Long)`.

app/src/main/java/app/revanced/manager/network/api/EndpointState.kt
```diff
@@ -0,0 +1,56 @@
+package app.revanced.manager.network.api
+
+import android.util.Log
+import app.revanced.manager.domain.manager.PreferencesManager
+import app.revanced.manager.util.tag
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+import kotlinx.coroutines.flow.asStateFlow
+
+class EndpointState(
```
oSumAtrIX, line 10: I don't think this class works for chained fallbacks and it is also overcomplicated. You simply need one persistence for a schema:

API(url: String, fallback: String?, restore: API?)

If current API is down -> fallback
If restore != null, in an interval probe all restore APIs recursively, the earliest wins.

mostafaNazari702: Reworked EndpointState into the priority-chain setup you described. Endpoints are now represented as "ApiEndpoint(url, fallback)" and the behavior matches what we discussed on Discord, if i rememeber correctly.

Requests always try the active endpoint first. if that one is down we walk the chain from the top and pick the first reachable endpoint, so higher-priority endpoints always win. for example if we are currently on C and it dies while A or B are available we will move back up to A/B instead of falling further down to D.

I also added the recovery logic for backup endpoints. while we are running on a backup, every 5 minutes we check whether any higher-priority endpoints have come back. If they have, we switch to the highest priority one that's reachable. over time that means we naturally climb back up the chain (e.g C then to B tehn to A) until we're back on the primary then the recovery loop stops doing anything.

I didnt keep an explicit "restore" link on the endpoint node. A doubly-linked immutable structure (A.fallback = B and B.restore = A) is awkward to construct with vals so the node only stores the fallback direction. when we need restore candidates we derive them by walking the chain from the head. Same behavior, just simpler to build and maintain ig.

Also worth noting that /about only exposes a single fallback today, so in practice we are still running with two tiers. The code just no longer assumes that's the limit, so if we add more fallback levels later, it should work without any further changes.

the literal "restore" field can be added back if you'd rather keep the structure closer to the original proposal, since we kinda....shifted a lot during the...journey?

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
mostafaNazari702: Removed the redundant trimEnd and compare directly.

app/src/main/java/app/revanced/manager/ManagerApplication.kt
mostafaNazari702: loop now collects endpointState.activeUrl and only runs while it's non-null (i.e on a backup), on the primary it stays suspended. collectLatest also restarts it after each partial restore so it keeps climbing the chain until it's back on the primary.

app/src/main/java/app/revanced/manager/ManagerApplication.kt
mostafaNazari702: the startup update check could run against a backup. after a successful restore I now re-run the update checks (runUpdateChecks()), silently (showToast = false) so results reflect the recovered endpoint without a jarrring UI change.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
mostafaNazari702: Done

app/src/main/java/app/revanced/manager/di/HttpModule.kt
mostafaNazari702: Applied

app/src/main/java/app/revanced/manager/ManagerApplication.kt
mostafaNazari702: switched to a Duration (5.minutes) and delay(Duration) so it's self-documenting and avoids the deprecated delay(Long). Thanks @PalmDevs.

app/src/main/java/app/revanced/manager/network/api/EndpointState.kt
oSumAtrIX: > C and it dies while A or B are available we will move back up to A/B instead of falling further down to D.

If A and B are up, it should go back to A again

oSumAtrIX: > Also worth noting that /about only exposes a single fallback today, so in practice we are still running with two tiers. The code just no longer assumes that's the limit, so if we add more fallback levels later, it should work without any further changes.

What do you mean? Theres one fallback, but the fallback itself can have another one, its recursive.

app/src/main/java/app/revanced/manager/network/api/EndpointState.kt
```diff
@@ -0,0 +1,77 @@
...
+        val fallback: ApiEndpoint?,
+    ) {
+        fun asSequence(): Sequence<ApiEndpoint> = generateSequence(this) { it.fallback }
+    }
+
+    // URL of the endpoint currently in use, null means the primary (chain head).
+    private val _activeUrl = MutableStateFlow<String?>(null)
+    val activeUrl: StateFlow<String?> = _activeUrl.asStateFlow()
+
+    suspend fun primaryUrl(): String = prefs.api.get().trimEnd('/')
```
oSumAtrIX, line 29: These should be properties not functions as the name suggests.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -10,55 +9,126 @@ import app.revanced.manager.network.dto.ReVancedGitRepository
...
+    suspend fun getDownloaderUpdate(): APIResponse<ReVancedAsset> =
+        request("manager/downloaders${prefs.useDownloaderPrerelease.prereleaseString()}")
 
-    suspend fun getAppHistory() = request<List<ReVancedAssetHistory>>("manager/history${prefs.useManagerPrereleases.prereleaseString()}")
+    suspend fun getContributors(): APIResponse<List<ReVancedGitRepository>> =
+        request("contributors")
 
-    suspend fun getPatchesUpdate() = request<ReVancedAsset>("patches${prefs.usePatchesPrereleases.prereleaseString()}")
+    suspend fun getInfo(): APIResponse<ReVancedInfo> {
+        val (response, servedBy) = requestTracked<ReVancedInfo>("about")
```
oSumAtrIX, line 50: Logic like this should not live in the API class. The API class is only responsible for the network layer.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -10,55 +9,126 @@ import app.revanced.manager.network.dto.ReVancedGitRepository
...
+            endpointState.updateFallbackFromAbout(response.data.api?.fallback)
+        }
+        return response
+    }
 
-    suspend fun getPatchesHistory(apiUrl: String, prerelease: Boolean) =
-        request<List<ReVancedAssetHistory>>(apiUrl, defaultApiVersion, "patches/history${prerelease.prereleaseString()}")
+    
+    // Probes the higher-priority endpoints above the active one and restores the earliest reachable one (the primary wins over any intermedieate backup). returns the restored endpoint or null if none higher than the active endpoint is currently reachable.
+    suspend fun restoreHigherPriorityEndpoint(): EndpointState.ApiEndpoint? = withContext(Dispatchers.IO) {
```
oSumAtrIX, line 59: This is all logic to be put in a different file, not API

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -10,55 +9,126 @@ import app.revanced.manager.network.dto.ReVancedGitRepository
...
-    suspend fun getContributors() = request<List<ReVancedGitRepository>>("contributors")
+    private suspend inline fun <reified T> request(route: String): APIResponse<T> =
+        requestTracked<T>(route).first
 
-    suspend fun getInfo() = request<ReVancedInfo>("about")
+    /**
+     * Routes a request for a patch-bundle source. Only the official source (the configured
+     * primary endpoint) participates in the primary -> fallback logic; a user-added custom
+     * remote source queries its own server directly and is never redirected to the backup.
+     */
```
oSumAtrIX, line 77: This is worth reconsidering.

The primary source should behave like, and technically be, just another source, no different from any other.

Right now, the patches API and the manager API share the same domain, differing only in path (`/manager` vs `/patches`). `/about` isn't actually patches-specific, it belongs to this shared domain, not to patches itself.

Our current fallback logic is built around the manager only. Patches would benefit from the same kind of fallback, but there's a gap: manager can pull `/about` as part of its own API to discover fallback info, while patches has no equivalent. A patch can be imported from an arbitrary location (e.g. `some.com/my/patches`), and there's no metadata there describing a fallback. Because of this, the only correct way to give patches a fallback mechanism is to build one that's independent of the manager's.

Instead add a new field to the patches JSON, mirroring what `/about` already does for manager.

With this in place, the "main" patches source behaves like any other source and can recover on its own. When manager and patches live in the same place, the "main" source's restore field would happen to point to the same location as the manager's fallback, but the two systems remain technically independent.

As a result, the patches model needs a new field, analogous to `/about`, along with matching fallback logic. This means the fallback logic itself needs to be refactored to be generic/abstract, not manager-specific, since manager and patches will now run on identical logic.

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -10,55 +9,126 @@ import app.revanced.manager.network.dto.ReVancedGitRepository
...
+    ): APIResponse<T> =
+        if (apiUrl == endpointState.primaryUrl()) {
+            request(route)
+        } else {
+            directRequest(apiUrl, route)
+        }
+
+    // Issue a request against the active endpoint and returns the first success with the URL that served it. 
+    // If the active endpoint is unreachable, the whole chain is walked from the primary downward so a higher-priority endpoint is preferred over a lower-priority one: i.e recovery up the chain is attempted before falling further down. The serving endpoint becomes active.
+    private suspend inline fun <reified T> requestTracked(
```
oSumAtrIX, line 90: Maybe call this requestWithFallback

app/src/main/java/app/revanced/manager/network/api/ReVancedAPI.kt
```diff
@@ -10,55 +9,126 @@ import app.revanced.manager.network.dto.ReVancedGitRepository
...
+                if (response is APIResponse.Success) {
+                    endpointState.setActive(endpoint)
+                    return@withContext response to endpoint.url
+                }
+                lastFailure = response
+            }
+        lastFailure to null
+    }
+
+    private suspend inline fun <reified T> directRequest(
```
oSumAtrIX, line 113: just request, but all code using the API should probably not use this function, what code still makes use of this?

app/src/main/java/app/revanced/manager/ManagerApplication.kt
```diff
@@ -78,16 +86,37 @@ class ManagerApplication : Application() {
...
+        // collectLatest restarts this block whenever the active endpoint changes: after a partial
+        // restore (e.g. C -> B while the primary is still down) the active URL is still non-null, so
+        // probing resumes for the remaining higher-priority endpoints until the primary is reached,
+        // at which point activeUrl is null and the loop stays idle.
+        scope.launch(Dispatchers.IO) {
+            endpointState.activeUrl
+                .filterNotNull()
+                .collectLatest {
+                    while (true) {
+                        delay(PRIMARY_RECONNECT_INTERVAL)
```
oSumAtrIX, line 108: To avoid an infinite probe, add  a hard limit, if the endpoints are probed e,g., for 3 times (5min*3 = 15min of attempting recovery), give up and stay on the current api. On application restart the same logic applies.

PR #3350, opened 2026-05-13
feat: Add share button for patched apk

secp192k1:
Just adds a share button on the patcher screen.
Pretty useful for installing with another app (wink wink)

It's using the default `output.apk` filename because:
1. Keep changes to a minimum
2. There would need a good chuck of changes for it to be possible
3. It doesn't need to have a pretty name, if the user wanted a pretty name they should use the `Export`/`Save` button anyways

|image | image | image |
|-|-|-|

### DIFF
--- a/app/src/main/java/app/revanced/manager/util/Util.kt
+++ b/app/src/main/java/app/revanced/manager/util/Util.kt
@@ -6,6 +6,7 @@ import android.content.Context
 import android.content.Intent
 import android.content.pm.ApplicationInfo
 import android.graphics.Bitmap
+import android.net.Uri
 import android.os.Handler
 import android.os.Looper
 import android.renderscript.Allocation
@@ -75,6 +76,17 @@ fun Context.openUrl(url: String) {
     startActivity(Intent(Intent.ACTION_VIEW, url.toUri()).apply {
         flags = Intent.FLAG_ACTIVITY_NEW_TASK
     })
+}
+fun Context.shareApk(apkUri: Uri) {
+    startActivity(
+        Intent.createChooser(Intent(Intent.ACTION_SEND).apply {
+            type = APK_MIMETYPE
+            putExtra(Intent.EXTRA_STREAM, apkUri)
+            flags = Intent.FLAG_GRANT_READ_URI_PERMISSION
+        }, null).apply {
+            flags = Intent.FLAG_ACTIVITY_NEW_TASK
+        }
+    )
 }
 
 fun Context.toast(string: String, duration: Int = Toast.LENGTH_SHORT) {
--- a/app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
+++ b/app/src/main/java/app/revanced/manager/ui/screen/PatcherScreen.kt
@@ -16,6 +16,7 @@ import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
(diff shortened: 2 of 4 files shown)


Ushie:
Share button is fine, I'm surprised we didn't have one infact

But if your purpose is what's implied within that wink wink, then an option in settings to use intent-based installation instead of session would be better UX

Ushie:
As for the output name, you could use the same name used by exporting the patch log file

secp192k1:
> [...] an option in settings to use intent-based installation instead of session would be better UX

I guess? Although if the `output.apk` becomes a coherent filename, then a share button would be nice anyways imo

> As for the output name, you could use the same name used by exporting the patch log file

`fun logFileName() = "revanced_patcher_${packageName}_${version}_${System.currentTimeMillis()}.txt"`
Instead of that, which has "patcher" (not "patched") and its huge because of the unix timestamp, not even considering the `.txt` at the end
Perhaps make it the same as the export apk name?
`"${viewModel.packageName}_${viewModel.version}_revanced_patched.apk"`

Ushie:
Sure looks perfect

oSumAtrIX:
Share should replace save, because sharing to files is the standardized way making save redundant

oSumAtrIX:
image
 
For some reason the perceived size is different than the other icons for me

secp192k1:
> image
> 
> For some reason the perceived size is different than the other icons for me

image
image

insane, its really really hard to see but yeah there is a size discrepancy

PalmDevs:
Material Icons moment

app/src/main/java/app/revanced/manager/util/Util.kt
```diff
@@ -76,6 +77,17 @@ fun Context.openUrl(url: String) {
...
+    startActivity(
+        Intent.createChooser(Intent(Intent.ACTION_SEND).apply {
+            type = APK_MIMETYPE
+            putExtra(Intent.EXTRA_STREAM, apkUri)
+            flags = Intent.FLAG_GRANT_READ_URI_PERMISSION
+        }, null).apply {
+            flags = Intent.FLAG_ACTIVITY_NEW_TASK
+        }
+    )
+}
```
validcube, line 90: If I remember Android 17 should complain about implict URI grant that it's soon to be future problem starting with Android 18 regardless if the app is targeting 18 or not.

But I assume you have already fix it here with `FLAG_GRANT_READ_URI_PERMISSION`?

validcube (commented):
Everything looks fine to me (aside from one general comment), although it would be good if the APK name is the same as APK content but if that's too much work we could add a TODO comment or file a feature request for it.

If resolved, I'll go ahead, and approve it, then merge it after 24 hours has passed. (Assuming there are no more review/requested changes after this from other designated reviewers; of course.)

app/src/main/java/app/revanced/manager/util/Util.kt
secp192k1: > But I assume you have already fix it here with `FLAG_GRANT_READ_URI_PERMISSION`?

Correct, additionally it didn't complain in my personal device running `CinnamonBun (CP31.260423.012.A1)` (not canary but still...)

secp192k1:
https://github.com/ReVanced/revanced-manager/commit/3b72fbf58be2ab4b9af09b4379cd345932a86fae
image

https://github.com/ReVanced/revanced-manager/commit/b4116ed6b52789a4a3f282dbd01ffcc8144c6c57
image

validcube (approved):
🥞 LGTM, merge on queue when 24hrs passed.

PalmDevs:
Can we put the share button next to the save button

Axelen123:
The share won't work after the file is deleted, but I guess that won't be a problem as long as the target app consumes the file promptly.

Axelen123 (approved):

secp192k1:
> Can we put the share button next to the save button

yes we can https://github.com/ReVanced/revanced-manager/pull/3350/commits/4266d048bf0783ade21c3d9c94b89b3482a69089

Issue #3060, opened 2026-03-15
bug: Unable to save parched apk


vicokoby:
### Bug description

Patching youtube apk doesn't finish correctly as the patched APK is no saved 

### Patch logs

```shell

```

### Debug logs

[revanced_patcher_com.google.android.youtube_20.40.45_1773603772512.txt](https://github.com/user-attachments/files/26008478/revanced_patcher_com.google.android.youtube_20.40.45_1773603772512.txt)

### Acknowledgements

- [x] I have checked all open and closed bug reports and this is not a duplicate.
- [x] I have chosen an appropriate title.
- [x] All requested information has been provided properly.
- [x] The bug is only related to ReVanced Manager.

azertyis1984:
Same issue here, android 11 armeabi v7a

YandereNoki:
Same issue, android 10

I though it was a ram issue so i tried patching with less and less and the gms core crashed but when i selected only 1 patch (hide ads), it gave me back an error 135

xuefer:
<img width="2048" height="1536" alt="Image" src="https://github.com/user-attachments/assets/c791c0c8-8cfb-4890-95de-3b664f216f51" />

[revanced-manager_logcat_2026-03-20T02_19_10.434.txt](https://github.com/user-attachments/files/26222791/revanced-manager_logcat_2026-03-20T02_19_10.434.txt)

PR #3398, opened 2026-06-04
feat: Various misc QoL enchancements

### LINKED ISSUE #3197: feat: provide "proof of life" when downloading apk bundles
### Feature description

Currently when trying to download/patch apk bundles, for example reddit, the download throbber just spins without any sign that anything is even being done. This makes it very easy for people to assume it's not downloading or doing anything at all as it looks no different than if something is frozen. 

### Motivation

The importance of this is primarily, better UX, and reduced cases of people saying "bundles don't work" "reddit won't download" on the discord. 

### Acknowledgements

- [x] I have checked all open and closed feature requests and this is not a duplicate

### LINKED ISSUE #3330: feat: indicate progress for merging of split APKs
### Feature description

pretty self explanatory

currently, it appears as if it is loading forever, but it does eventually complete, it just doesn't have a progress bar:

Image

### Motivation

Showing progress would (obviously) be better UX. The amount of time it took led me to believe it would just never finish so I just cancelled the patching and assumed it was broken which was wrong

### Acknowledgements

- [ ] I have checked all open and closed feature requests and this is not a duplicate

### LINKED ISSUE #3392: bug: unable to download patches; Out of Memory
### Bug description

Tried to patch YT, it fails downloading the patches, don't know why [revanced_patcher_com.google.android.youtube_20.40.45_1780483389348.txt](https://github.com/user-attachments/files/28547522/revanced_patcher_com.google.android.youtube_20.40.45_1780483389348.txt)

### Patch logs

```shell
App: com.google.android.youtube (Selected: 20.40.45, Suggested: 20.40.45)
Source: ReVanced Patches
Removed: Alternative thumbnails, Change header, Disable fullscreen ambient mode, Downloads, Enable debugging, Hide autoplay preview, Hide info cards, Hide timestamp
Manager version: 2.7.0-dev.4
Use manager pre-releases: true
Manager auto-update: true
Allow changing patch selection: true
Show universal patches: false
Use patches pre-releases: true
Use process runtime: true
Process runtime custom memory limit: 2500MB
Memory limit: 256MB (large: 512MB)
Root permissions: No
RAM: 3,20 GB / 11,84 GB available
Storage: 301 GB / 488 GB available
Android version: 16 (API 36)
Supported architectures: arm64-v8a
Model: CPH2797
```

### Debug logs

[revanced-manager_logcat_2026-06-03T12_45_06.219867.txt](https://github.com/user-attachments/files/28547516/revanced-manager_logcat_2026-06-03T12_45_06.219867.txt)

### Acknowledgements

secp192k1:
### Live updates for patcher screen
<details>
<summary>Implemented live updates for: Decoding resources</summary>

- Ticks `Decoding resources [space] Xs)` every second, replacing in place
- Final `Decoded resources in Xs` replaces the live tick
- Final summary also written to the exported log file

| Live updates: In progress  | Live updates: Finished  | Logs snippet |
|-|-|-|
| image | image | `[INFO] Applying patches`<br>`[INFO] Decoded resources in 31s`<br>`[INFO] Writing patched files...` |
</details>

<details>
<summary>Implemented live updates for: Downloading APK file</summary>

- Ticks `Download APK file [spacer] [size] - Xs)` every second, replacing in place
- Final `Downloaded APK file in Xs [spacer] [size]` replaces the live tick
- Final summary also written to the exported log file
- Fixes #3197
- Fixes #3330 

| Live updates: In progress  | Live updates: Finished  | Logs snippet |
|-|-|-|
| image | image | `[INFO] Applying patches`<br>`[INFO] Decoded resources in 31s`<br>`[INFO] Writing patched files...` |
</details>

### CI: Skip automatic builds for drafts
As the title implies, it just prevents GitHub Actions to build on PR drafts, everything else it unchanged (i.e. keeps building on commits)

### Clear downloaded APK cache

Just a button under `Settings` => `Developer options` to clear the cache of downloaded APKs
<details><summary>Image preview</summary>
image
</details>

### Fix network memory leak

The flow for donwloads is:
1. Read chunk
2. Write chunk
3. Repeat

Every read operation allocates a small byte container, everything writes correctly, but the containers are never freed from memory after use, causing them to persist and accumulate

To fix this, we wrap each chunk inside a `.use { }` blocks for automatic resource management:
- Read chunks: containers are automatically freed from memory once processed
- Output sink: ensure it is properly flushed and closed, preventing the memory leak

Instead of wrapping the main logic in an `if (success)` check, it now throws early if it fails
It keeps the exact same behavior but saves us a level of indentation and its better practice

Fixes #3392

### Dedicated dialogs for patcher kills and native crashes

`ProcessRuntime` throws a generic `Exception` for any non-zero exit code from the patcher subprocess, leaving users with the generic "Process exited with nonzero exit code X"
A good example is on #3060 which on low-RAM devices is almost always Android's LowMemoryKiller, not a real bug

Added signal exit codes to two typed exceptions routed to dedicated `InstallerStatusDialog` kinds:
- `PatcherKilledException` for when the system terminates the process
- `PatcherCrashedException` for when the process crashed natively

And `InstallerModel` now has a `retry()` so the dialog button can a clean install attempt can be tried from the installer screen

### Soft-skip failed patches - WIP!

When a patch fails, the entire process dies, this addition makes it so keeps patching, skipping the failed patch

<details><summary>Image preview</summary>

| image | image |
|-|-|
</details>

### Fix step icons not aligned

<details><summary>Image preview</summary>

| Before | After |
|-|-|
| image | image |

</details>

### Removed useless prefix

The prefix `[INFO]` is useless and shouldn't be there, removed only for the INFO level

### Styling and refactor

<details><summary>`strings.xml` has style tags like `<b>` but it doesn't work, now it does!</summary>

| Screenshot_20260606_175300 | Screenshot_20260606_210058 |
|-|-|
</details>

Now these components animate:
- Notifications in the dashboard
- Battery optimization warning
- Onboarding permission items


oSumAtrIX:
Maybe just say "Decoding resources (Xs)" instead of "Still". However this entire PR has a contract issue. 

The patcher is what emits the arbitrary log strings. There is no contracted guarantee those strings will change. You cant "assume" a specific string

Similarly replaceLineAtIndex isnt guaranteed, since the patcher might emit more or less strings shifting the index.

Ushie:
What's left here? it's best to make sure this doesn't remain in this PR for too long otherwise it's going to stale out like its siblings

secp192k1:
> What's left here? it's best to make sure this doesn't remain in this PR for too long otherwise it's going to stale out like its siblings

Only thing that was left was any future ideas of what to add next, but I 100% agree. Readyyyyy

app/src/main/res/values/strings.xml
```diff
@@ -566,6 +568,10 @@ It’s only compatible with these versions: %2$s</string>
     <string name="installation_invalid_dialog_title">Install invalid</string>
     <string name="installation_storage_issue_dialog_title">Not enough storage</string>
     <string name="installation_timeout_dialog_title">Timed out</string>
+    <string name="installation_patcher_killed_dialog_title">Patcher killed</string>
```
PalmDevs: This sounds like the user invoked an action. I think it should be

`Patcher was killed`

app/src/main/res/values/strings.xml
```diff
@@ -566,6 +568,10 @@ It’s only compatible with these versions: %2$s</string>
     <string name="installation_invalid_dialog_title">Install invalid</string>
     <string name="installation_storage_issue_dialog_title">Not enough storage</string>
     <string name="installation_timeout_dialog_title">Timed out</string>
+    <string name="installation_patcher_killed_dialog_title">Patcher killed</string>
+    <string name="installation_patcher_killed_description">The patcher subprocess was terminated by the system, usually due to low memory. Close other apps or reduce the number of selected patches and try again.</string>
+    <string name="installation_patcher_crashed_dialog_title">Patcher crashed</string>
+    <string name="installation_patcher_crashed_description">The patcher subprocess crashed. This is usually caused by a bug in a native dependency or by a malformed input APK. Try a different version of the app, fewer patches, or report the issue with the exported log.</string>
```
PalmDevs: The user doesn't know what a "dependency" is.

`The patcher subprocess crashed. This may be caused by a bug, or a malformed input APK. Try a different version of the app, fewer patches, or report the issue with exported logs.`

app/src/main/res/values/strings.xml
```diff
@@ -292,6 +292,8 @@ You won’t be able to update apps that were signed with the previous keystore.<
     <string name="patches_force_download">Force download all patches</string>
     <string name="patches_reset">Reset patches</string>
     <string name="downloaders_reset">Reset downloaders</string>
+    <string name="downloaded_apps_clear">Clear downloaded APK cache</string>
```
PalmDevs: Not sure if this is necessarily always an APK. In the future, I think we may support other formats, so it's best to do

`Clear download cache` or `Clear downloaded apps cache`.

Thoughts @oSumAtrIX?

app/src/main/java/app/revanced/manager/util/Util.kt
```diff
@@ -388,4 +394,20 @@ fun blurBackground(
     rs.destroy()
 
     return workingBitmap
-}
\ No newline at end of file
+}
+
+private val boldRegex = Regex("</?b>")
```
PalmDevs, line 400: Is this how other apps do it? Is there a dependency that does this for us?

PalmDevs (changes requested):
Needs some small changes.

Will string changes be reflected to Crowdin? @validcube

validcube:
> Needs some small changes.
> 
> Will string changes be reflected to Crowdin? @validcube

For changes applied to default language strings (english), yes. For changes applied to other languages they will be reverted to original state.

app/src/main/java/app/revanced/manager/util/Util.kt
secp192k1: I'm not really sure how other apps do it I'll be honest

I personally like the regex method because its the cleanest, instead of a bunch of text splits and find index etc..

app/src/main/res/values/strings.xml
secp192k1: Well, regardless, if thats the case for the future, maybe `Clear downloaded assets cache`

oSumAtrIX:
I have asked this before but still have not understood how you are certain about where to inject "[INFO] Decoded resources in 31s" in the logs. What if the patcher adds or removes certain logs? How do you know when decoding starts and ends? This part of the process is not in the hands of manager, and probably even a concern of patcher if at all.

> Soft-skip failed patches - WIP!

Patches should never raise an exception back to caller. Can you provide a stack trace or location where an exception is unhandled?

> The prefix [INFO] is useless and shouldn't be there, removed only for the INFO level

The formatting is not user friendly. Instead it should say:

Info:
Error:
Warning:

or co.

Perhaps the raw log string can also be styled, but a "lazy" string works too for the while.

secp192k1:
> What if the patcher adds or removes certain logs?

We never read the patcher's logs to time this. The timer is just a coroutine ticking every second, totally separate from whatever the patcher prints:
```kt
val heartbeat = Heartbeat(this) { elapsed ->
    emitLog("Decoding resources", "${elapsed}s")
}
```
The patcher could even not print anything and it would still keep ticking

> How do you know when decoding starts and ends?

The start time will be the moment `patcher` is called, and the end time is the first patch result we get back, the first step is to decode the resources which makes it predictable
```kt
patcher { (patch, exception) ->
    heartbeat.complete { elapsed ->
        val message = "Decoded resources in ${elapsed}s"
        logger.info(message)
        emitLog(message)
    }
    // ...
```

And we cant call `complete` multiple times because of this line (`fun complete()` at `Heartbeat.kt`)
```kt
if (!finished.compareAndSet(false, true)) return
```
So later patches dont run this again

> This part of the process is not in the hands of manager, and probably even a concern of patcher if at all.

Yes and that stays true since we arent changing any behaviour of the patcher, just using its steps to get a rough estimate of the ETA

> Patches should never raise an exception back to caller. Can you provide a stack trace or location where an exception is unhandled?

And it doesnt, the exception gets passed as a regular string variable on the result callback, nothing is unhandled
```kt
patcher { (patch, exception) ->

local val patch: Patch
local val exception: PatchException?
```
The only place an exception actually flies is at the very end of that callback, where we re-throw it on purpose to abort the run
```kt
if (exception != null) {
    // ...
    phaseLogger.error("${patch.name} failed:")
    phaseLogger.error(exception.stackTraceToString())
    throw exception
} // ...
```

The soft-skip is only preventing the re-throw when the exception looks like a missing fingerprint, and we add a `return@patcher` to keep the loop going

Ideally, re-throwing would never happen and for each failed patch its treated as a soft-skip / skippable task
After all, it is a WIP, I wasn't sure if some failures leave the APK half-broken since some patches depend on others, making it unsafe to just completely ignore..

oSumAtrIX:
Moved to https://github.com/ReVanced/revanced-manager/pull/3459

PR #3436, opened 2026-06-28
feat: unify importers

mostafaNazari702:


app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+import android.net.Uri
+import app.revanced.manager.network.service.HttpService
+import io.ktor.client.request.url
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.withContext
+import java.io.IOException
+import java.io.InputStream
+
+
+// open streams for URIs of the [schemes] it is registered for.
```
oSumAtrIX: Start comments with upper casing

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+interface ProtocolHandler {
+    val schemes: Set<String>
```
oSumAtrIX: Any reason this is public and not protected/private?

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,36 @@
...
+    override suspend fun getStream(uri: Uri): InputStream = http.getStream {
+        url(uri.toString())
+    }
+}
+
+class FileProtocolHandler(private val app: Application) : ProtocolHandler {
+    override val schemes = setOf("file", "content")
+
+    override suspend fun getStream(uri: Uri): InputStream = withContext(Dispatchers.IO) {
+        app.contentResolver.openInputStream(uri) ?: throw IOException("Cannot open $uri")
```
oSumAtrIX: I dont think this handles opening the file picker yet. This works for the scheme "content" but for "file", it needs to invoke the file picker, let the user select the file, which returns a content url, and then obtain a stream to that url. The file picker requires interaction with UI and to communicate to UI, you will need a Channel. You can inject necessary references in the constructor for that. A UI side handler on the other side of the channel can handle the picker related stuff.

app/src/main/java/app/revanced/manager/network/service/HttpService.kt
```diff
@@ -85,6 +97,49 @@ class HttpService(
         }
     }
 
+    // Make a GET request and returns the response body as a stream.
+    // Caller is responsible for closing the returned stream.
+    // Failures happening while streaming are thrown from the stream's read methods.
+    fun getStream(builder: HttpRequestBuilder.() -> Unit): InputStream {
```
oSumAtrIX: This looks overcomplicated. See https://ktor.io/docs/client-responses.html#writing-the-response-directly-to-a-file (not the file related stuff). However it looks scope bound so you might not be able to return a stream and instead pass it through a callback

app/src/main/java/app/revanced/manager/network/service/HttpService.kt
```diff
@@ -85,6 +97,49 @@ class HttpService(
...
+        val failure = AtomicReference<Throwable?>(null)
+        val stream = PipedInputStream(DEFAULT_BUFFER_SIZE)
+        val output = PipedOutputStream(stream)
+
+        val writer = writerScope.launch {
+            try {
+                streamTo(output, builder)
```
oSumAtrIX: This buffers the file temporarily in memory, works for small files, but still not something we want, a callback approach is likely inevitable

oSumAtrIX:
Next steps after review can concern with registering the handlers or getting rid of the "Local" and "Remote" concept by merging both into one. A local and remote source of patches therefore becomes identical in shape and behaviour.

app/src/main/java/app/revanced/manager/di/ServiceModule.kt
```diff
@@ -1,9 +1,27 @@
...
 import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.module
 
 val serviceModule = module {
     singleOf(::HttpService)
+    single { Channel<FilePickRequest>() }
+    singleOf(::HttpProtocolHandler)
+    singleOf(::ContentProtocolHandler)
+    singleOf(::FileProtocolHandler)
+    single<Map<String, ProtocolHandler>> {
```
oSumAtrIX: Where is this used?

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+    suspend fun <T> getStream(uri: Uri, block: suspend (InputStream) -> T): T
+}
+
+class HttpProtocolHandler(private val http: HttpService) : ProtocolHandler {
+    override suspend fun <T> getStream(uri: Uri, block: suspend (InputStream) -> T) =
+        http.getStream({ url(uri.toString()) }, block)
+}
+
+// Opens content:// URIs, which the platform grants the app access to.
+class ContentProtocolHandler(private val app: Application) : ProtocolHandler {
```
oSumAtrIX: Any reason you dont inject     private val content: ContentProtocolHandler like you did for FileProtocolHandler?

app/src/main/java/app/revanced/manager/di/ServiceModule.kt
oSumAtrIX: I think this is better as a singleton field somewhere, instead of a dependency. You can create a top level field

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+            app.contentResolver.openInputStream(uri) ?: throw IOException("Cannot open $uri")
+        }
+
+        return stream.use { block(it) }
+    }
+}
+
+// A request for the UI to let the user pick a file, answered with a content:// URI.
+class FilePickRequest {
+    val result = CompletableDeferred<Uri?>()
```
oSumAtrIX: Why is this Deferred?

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+        val stream = withContext(Dispatchers.IO) {
+            app.contentResolver.openInputStream(uri) ?: throw IOException("Cannot open $uri")
+        }
+
+        return stream.use { block(it) }
+    }
+}
+
+// A request for the UI to let the user pick a file, answered with a content:// URI.
+class FilePickRequest {
```
oSumAtrIX: Is this how you do Channel to UI?

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+    val result = CompletableDeferred<Uri?>()
+}
+
+// Reading file:// URIs directly requires storage permissions, which the app avoids.
+// The user instead picks the file through the system file picker, which yields
+// a content:// URI the app is allowed to open.
+class FileProtocolHandler(
+    private val pickRequests: Channel<FilePickRequest>,
```
oSumAtrIX:
```suggestion
    private val filePickRequestChannel: Channel<FilePickRequest>,
```

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
```diff
@@ -0,0 +1,56 @@
...
+    private val content: ContentProtocolHandler
```
oSumAtrIX:
```suggestion
    private val contentProtocolHandler: ContentProtocolHandler
```

app/src/main/java/app/revanced/manager/network/service/HttpService.kt
```diff
@@ -85,6 +87,18 @@ class HttpService(
         }
     }
 
+    // Makes a GET request and passes the response body to [block] as a stream.
+    // The stream is only valid inside [block] and is closed automatically afterwards.
+    suspend fun <T> getStream(
+        builder: HttpRequestBuilder.() -> Unit,
```
oSumAtrIX: builder should be second parameter just like in the other functions too

app/src/main/java/app/revanced/manager/network/service/HttpService.kt
oSumAtrIX:
```suggestion
// Additional authors: Aliucord Authors, DiamondMiner88
```

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -72,13 +77,17 @@ import app.revanced.manager.util.deepLinkedComposable
...
+import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.launch
+import org.koin.android.ext.android.inject
 import org.koin.androidx.compose.koinViewModel
 import org.koin.core.parameter.parametersOf
 import java.util.Locale
 import org.koin.androidx.viewmodel.ext.android.getViewModel as getActivityViewModel
 
 class MainActivity : AppCompatActivity() {
+    private val filePickRequests: Channel<FilePickRequest> by inject()
```
oSumAtrIX:
```suggestion
    private val filePickRequestChannel: Channel<FilePickRequest> by inject()
```

oSumAtrIX: Is main activitiy the correct place?

oSumAtrIX: Why is "inject()" used here

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -117,6 +127,26 @@ class MainActivity : AppCompatActivity() {
     }
 }
 
+// Answers file pick requests from the file protocol handler by launching the system file picker.
+@Composable
+private fun FilePickerHost(requests: Channel<FilePickRequest>) {
```
oSumAtrIX:
```suggestion
private fun FilePickerHost(channel: Channel<FilePickRequest>) {
```

app/src/main/java/app/revanced/manager/MainActivity.kt
```diff
@@ -117,6 +127,26 @@ class MainActivity : AppCompatActivity() {
     }
 }
 
+// Answers file pick requests from the file protocol handler by launching the system file picker.
```
oSumAtrIX: You should not specialize this to the file protocol handler. The communication layer between backend and frontend is a separate feature from the protocol handlers, not specific to them.

app/src/main/java/app/revanced/manager/domain/protocol/ProtocolHandler.kt
mostafaNazari702: it's the response slot, sender suspends on it until the UI completes it, i just kept it over a continuation in compose state approach because the pending request has to survive activity recreation: the picker result gets redelivered after rotation and a continuation living in remember{} would be gone by then, hanging the sender!!!

app/src/main/java/app/revanced/manager/ui/component/UiRequestHost.kt
oSumAtrIX, line 1: Calling this file "UiRequestHost" but it only implemen ting file picker related code doesnt look right.

app/src/main/java/app/revanced/manager/ui/component/UiRequestHost.kt
```diff
@@ -0,0 +1,39 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import app.revanced.manager.util.FilePickRequest
+import app.revanced.manager.util.UiRequest
+import app.revanced.manager.util.uiRequestChannel
+
+// The request currently being fulfilled. Lives outside the composition so that a result
+// delivered after the activity is recreated (e.g. a rotation while the file picker
+// is open) still completes the request instead of dropping it.
+private var currentRequest: UiRequest<*>? = null
```
oSumAtrIX: For that remember { } exists.

oSumAtrIX: Maybe a koin single is better here?

app/src/main/java/app/revanced/manager/ui/component/UiRequestHost.kt
oSumAtrIX: Actually the current implementation breaks open closed, simply rename this to FilePickerRequestHost

app/src/main/java/app/revanced/manager/util/UiRequest.kt
```diff
@@ -0,0 +1,36 @@
+package app.revanced.manager.util
+
+import android.net.Uri
+import kotlinx.coroutines.CompletableDeferred
+import kotlinx.coroutines.channels.Channel
+
+// Communication layer for background code that needs the UI to fulfill a request,
+// such as showing a system dialog. UI hosts receive requests from [uiRequestChannel]
+// and complete [UiRequest.result] with the outcome.
+sealed class UiRequest<T> {
```
oSumAtrIX: No super UiRequest class needed, just FilePickerRequest is enough

app/src/main/java/app/revanced/manager/util/UiRequest.kt
```diff
@@ -0,0 +1,36 @@
...
+    // The response slot. The sender suspends on it until the UI completes it.
+    val result = CompletableDeferred<T>()
+}
+
+// A request to let the user pick a file, answered with a content:// URI
+// or null if the picker was cancelled.
+class FilePickRequest : UiRequest<Uri?>()
+
+val uiRequestChannel = Channel<UiRequest<*>>()
```
oSumAtrIX: This becomes filePickerRequestChannel

app/src/main/java/app/revanced/manager/ui/component/UiRequestHost.kt
```diff
@@ -0,0 +1,39 @@
...
+
+// Fulfills UI requests sent by background code. Mounted once at the root of the UI,
+// unconditionally, so that pending activity results are always delivered.
+@Composable
+fun UiRequestHost() {
+    val filePicker = rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) { uri ->
+        (currentRequest as? FilePickRequest)?.result?.complete(uri)
```
oSumAtrIX: You dont need casting anymore when your request turns to FilePickerRequest

app/src/main/java/app/revanced/manager/domain/manager/SourceManager.kt
```diff
@@ -259,7 +258,9 @@ abstract class SourceManager<DB : SourceManager.DatabaseEntity, LOADED, OUTPUT>(
         }
 
     suspend fun reloadApiSources() = dispatchAction("Reload API sources") { state ->
-        this@SourceManager.store.state.value.sources.values.filterIsInstance<APISource<LOADED>>()
+        this@SourceManager.store.state.value.sources.values
```
oSumAtrIX, line 312: The code should not have any reference to any variations of a source

app/src/main/java/app/revanced/manager/di/ServiceModule.kt
```diff
@@ -1,9 +1,20 @@
...
 import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.module
 
 val serviceModule = module {
     singleOf(::HttpService)
-}
\ No newline at end of file
+    singleOf(::UiFilePicker) { bind<FilePicker>() }
+    singleOf(::HttpProtocolHandler)
+    single { ContentProtocolHandler(androidContext().contentResolver) }
```
oSumAtrIX: Cant androidContext().contentResolver be injected automatically?

mostafaNazari702: Yes, it can!

mostafaNazari702:
[move UI requests into a generic communication layer](https://github.com/ReVanced/revanced-manager/pull/3436/commits/a9e80480cc4a252307187c9261a6af09dbdb56db) => [route local imports through the protocol handlers](https://github.com/ReVanced/revanced-manager/pull/3436/commits/844dce762b2af928c1fc7f132c95b2738b7c7d27) = Rebase.

mostafaNazari702:
Need review on my commits [merge local and remote sources into one](https://github.com/ReVanced/revanced-manager/pull/3436/commits/85ef81b064f94e90859d0c3a8651f137d8d0075f) + [provide the content resolver through the koin graph](https://github.com/ReVanced/revanced-manager/pull/3436/commits/9ea6088125fa968eb470c17611b6b720bdf66cac)

app/src/main/java/app/revanced/manager/domain/sources/Source.kt
```diff
@@ -14,16 +22,37 @@ fun interface Loader<T> {
...
+    if (!hasSerializationFailure) return this
+
+    return when (this) {
+        is APIFailure -> UnsupportedRemoteSourceException(this)
+        else -> UnsupportedRemoteSourceException(this)
+    }
+}
+
 /**
  * A [PatchBundle] or [app.revanced.manager.downloader.Downloader] source.
```
oSumAtrIX: This is very far fetched abstraction, if downloader or patches change slightly, the abstraction could fail

app/src/main/java/app/revanced/manager/domain/sources/Source.kt
```diff
@@ -36,18 +65,43 @@ sealed class Source<T>(
     }
 
     val isDefault inline get() = uid == 0
+
+    // A source can update itself when a handler knows how to reach its origin.
```
oSumAtrIX: What does this mean

app/src/main/java/app/revanced/manager/data/room/sources/Source.kt
```diff
@@ -22,10 +23,15 @@ sealed class Source {
...
     companion object {
-        fun from(value: String) = when (value) {
-            Local.SENTINEL -> Local
-            API.SENTINEL -> API
-            else -> Remote(Url(value))
+        fun from(value: String): Source {
+            val uri = Uri.parse(value)
+
+            return when (uri.scheme) {
+                // Rows written before sources were stored as URIs are plain
```
oSumAtrIX: Instead of this, run a database migration on boot, to get rid of this case. There should not be nay "API" vs "Local"

app/src/main/java/app/revanced/manager/domain/sources/Source.kt
```diff
@@ -14,16 +22,37 @@ fun interface Loader<T> {
 
 typealias PatchBundleSource = Source<PatchBundle>
 
+class UnsupportedRemoteSourceException(cause: Throwable? = null) : Exception(cause)
```
oSumAtrIX: There should not be any "Remote" semantic. Locality is handled by the protocol handlers

app/src/main/java/app/revanced/manager/di/ServiceModule.kt
```diff
@@ -1,9 +1,21 @@
...
+import org.koin.android.ext.koin.androidContext
+import org.koin.core.module.dsl.bind
 import org.koin.core.module.dsl.singleOf
 import org.koin.dsl.module
 
 val serviceModule = module {
     singleOf(::HttpService)
-}
\ No newline at end of file
+    singleOf(::UiFilePicker) { bind<FilePicker>() }
```
oSumAtrIX, line 16: What does bind do?

app/src/main/java/app/revanced/manager/ui/component/FilePickerRequestHost.kt
```diff
@@ -0,0 +1,32 @@
...
+import org.koin.compose.koinInject
+
+// Fulfills file pick requests sent by background code. Mounted once at the root
+// of the UI, unconditionally, so pending activity results are always delivered.
+@Composable
+fun FilePickerRequestHost(picker: UiFilePicker = koinInject()) {
+    val launcher = rememberLauncherForActivityResult(ActivityResultContracts.GetContent()) { uri ->
+        picker.currentRequest?.result?.complete(uri)
+        picker.currentRequest = null
+    }
```
oSumAtrIX, line 18: Wouldnt it be better:

if (picker.currentRequest?..result != null) {
 remember ...
}

Then you avoid calling this remember when result is not null

app/src/main/java/app/revanced/manager/ui/screen/settings/DownloadersInfoScreen.kt
```diff
@@ -215,7 +215,7 @@ fun DownloaderInfoScreen(
                     )
                 }
 
-                remote?.endpoint?.takeUnless { source.isDefault }?.let { url ->
+                remote?.uri?.toString()?.takeUnless { source.isDefault }?.let { url ->
```
oSumAtrIX: Not sure but if remote refers to the downloaders locality, this variable needs to be renamed or changed

app/src/main/java/app/revanced/manager/di/ServiceModule.kt
mostafaNazari702: it also registers the same singleton as FilePicker, file protocol handler injects the interface, the UI host injects the actual UiFilePicker that owns the pending request and both use the same object.

app/src/main/java/app/revanced/manager/ui/component/FilePickerRequestHost.kt
mostafaNazari702: Maybe better not. The activity result docs require the launcher callback to be registered on every activity creation. If activity gets recreated while the picker is open the result may be delivered with no callback registered and get lost which then leaves the pending request hanging. remember just keeps the callback registered!

app/src/main/java/app/revanced/manager/data/room/sources/Source.kt
mostafaNazari702: done.

app/src/main/java/app/revanced/manager/domain/repository/PatchBundleRepository.kt
```diff
@@ -65,8 +65,8 @@ class PatchBundleRepository(
         }
 
         val uri = when (source) {
-            // Imported files are copied into app storage and their origin is not recorded.
-            is SourceInfo.Local -> Uri.parse("rvp:${file.absolutePath}")
+            // Imported files are copied into app storage, the source points at that copy.
+            is SourceInfo.Local -> Uri.fromFile(file)
```
oSumAtrIX: No distinction of local, api, or remote should exist

oSumAtrIX: The persistence needs to get rid of it as well, and instead a migration should be added to migrate the old schema to the new

app/src/main/java/app/revanced/manager/domain/manager/SourceManager.kt
```diff
@@ -337,25 +336,25 @@ abstract class SourceManager<DB : SourceManager.DatabaseEntity, LOADED, OUTPUT>(
...
             }.getOrThrow()
-        }.exceptionOrNull()?.toRemoteValidationMessage()
+        }.exceptionOrNull()?.toValidationMessage()
     }
 
-    private fun Throwable.toRemoteValidationMessage() = when (asRemoteSourceException()) {
+    private fun Throwable.toValidationMessage() = when (asSourceException()) {
         // wtf is this? this data is not a bundle, at least something!
-        is UnsupportedRemoteSourceException -> app.getString(R.string.remote_source_url_unsupported)
+        is UnsupportedSourceException -> app.getString(R.string.remote_source_url_unsupported)
```
oSumAtrIX: References to remote still exist.

app/src/main/java/app/revanced/manager/domain/repository/PatchBundleRepository.kt
mostafaNazari702: This is already fixed in 602a23c, it was pushed after this review, right before the "mostafa requested a review from oSuMATrix, "[store sources as plain URLs and migrate old rows on boot](https://github.com/ReVanced/revanced-manager/pull/3436/commits/602a23cf2ff019b439fe92db7f159b65f3c31652)".

The RoomSource sealed class is gone now, we just store a plain Uri in the entities and the converter is simply `Uri.parse()` / `toString()`.

Old rows get migrated when the app starts in `SourceManager.loadFromDb()`: `uid` 0 gets updated to the configured default URL and old local rows without a scheme get rewritten to file:// URLs that point to their private copy.

The migration is safe to run more than once and the schema identity hash didnt change because the only thing that changed was the converter.

app/src/main/java/app/revanced/manager/domain/manager/SourceManager.kt
mostafaNazari702: Fixed.
validateUrl doesn't go through HttpService anymore, It now resolves the protocol through the handler mapjust like everything else so SourceManager no longer has any dependency on HTTP.

I also renamed the string resources from `remote_source_url_*` to `source_url_*`, and `setEndpoint` is now to `setUrl`.

The only place that still uses the old local/remote naming is `ImportSourceDialog` that's getting replaced by the Figma redesign anyway so i didn't think it was worth renaming a bunch of translated strings right before removing them. I can clean it up anyway, but are you truely sure you'd rather have it cleaned up?

app/src/main/java/app/revanced/manager/data/room/bundles/PatchBundleEntity.kt
```diff
@@ -1,15 +1,15 @@
...
-import app.revanced.manager.data.room.sources.Source
 import app.revanced.manager.domain.manager.SourceManager
 
 @Entity(tableName = "patch_bundles")
 data class PatchBundleEntity(
     @PrimaryKey override val uid: Int,
     @ColumnInfo(name = "name") val name: String,
     @ColumnInfo(name = "version") val versionHash: String? = null,
-    @ColumnInfo(name = "source") val source: Source,
+    @ColumnInfo(name = "source") val source: Uri,
```
oSumAtrIX: This should be renamed to url and be of type URL not Uri, migration needs to be added for it

app/src/main/java/app/revanced/manager/data/room/sources/SourceProperties.kt
```diff
@@ -0,0 +1,12 @@
+package app.revanced.manager.data.room.sources
+
+import android.net.Uri
+import androidx.room.ColumnInfo
+
+data class SourceProperties(
```
oSumAtrIX, line 13: Where is this used

mostafaNazari702: Its the partial type returned by `dbGetProps(uid)`, used by `updateDb` to update only the mutable fields without loading the full entity. Both DAOs return it from `getProps`.

mostafaNazari702:
The 6.json alone made the diff massive 
bild

oSumAtrIX:
Does it migrate from the old fields

mostafaNazari702:
> Does it migrate from the old fields

It does, tested on my phone.

app/src/main/java/app/revanced/manager/data/room/sources/SourceProperties.kt
oSumAtrIX: Where is the full scheme, and where is source properties used, the question is still unanswered properly

app/src/main/java/app/revanced/manager/ui/component/sources/ImportSourceDialog.kt
```diff
@@ -28,295 +37,292 @@ import app.revanced.manager.ui.component.AlertDialogExtended
...
-    val inputsAreValid by remember {
-        derivedStateOf {
-            (sourceType == SourceType.Local && local != null) ||
-                    (sourceType == SourceType.Remote && remoteUrl.isNotEmpty() &&
-                            (URLUtil.isHttpUrl(remoteUrl) || URLUtil.isHttpsUrl(remoteUrl)))
-        }
+    // Auto accepts any text and defers the scheme to the handlers, URL expects a
+    // well-formed web address before the network is even touched.
+    val isValidUrl = url.trim().let { URLUtil.isHttpUrl(it) || URLUtil.isHttpsUrl(it) }
+    val inputsAreValid = when (method) {
```
oSumAtrIX, line 100: This doesnt suite open closed principle well. Whats the reason for this implementation

app/src/main/java/app/revanced/manager/ui/component/sources/ImportSourceDialog.kt
```diff
@@ -28,295 +37,292 @@ import app.revanced.manager.ui.component.AlertDialogExtended
...
+                                ),
+                                label = {
+                                    Text(
+                                        stringResource(
+                                            if (method == ImportMethod.Auto) R.string.import_source_input
+                                            else strings.urlLabel
+                                        )
+                                    )
+                                },
+                                placeholder = if (method == ImportMethod.Http) {
```
oSumAtrIX, line 176: Breaks open closed, the set of handlers should map to composeables, so that you can iterate over the map and inject the respective composable into the ui, not manually hardcode them

app/src/main/java/app/revanced/manager/data/room/sources/SourceProperties.kt
mostafaNazari702: The full schema is the entity. `PatchBundleEntity` has `uid`, `name`, `version`, `url`, `auto_update`, `released_at`. DownloaderEntity is the same.

`SourceProperties` is a projection of the columns that can change after a row is created. It is the return type of `getProps` in both DAOs.

It is used by `updateDb` in `SourceManager`: read the props, copy with the changes then rebuild the entity with `entityFromProps`. `createEntity` uses the same path. `uid` is not in it because it is the primary key, not a mutable field.

PR #3433, opened 2026-06-27
fix: Change SELinux context to match stock APK before mounting

PalmDevs:
fixes #3428

PR #3469, opened 2026-07-23
fix: app root mount not applied after reboot (service.sh)

S4adam:
**fixes: https://github.com/ReVanced/revanced-manager/issues/3188** 

as PR https://github.com/ReVanced/revanced-manager/pull/3433/ did not do the trick in my case.

Changes:
- `stock_version` might return several packages so only take the first result (newest one).
- `SIGKILL` the app to apply the mount.

Tested on Youtube and YT Music apps on a Samsung s10e with [ExtremeROM Nexus](https://github.com/ExtremeXT/ExtremeROM/) 2.5.0-a38fc5f7 (beyond0lte).

### DIFF
--- a/app/src/main/assets/root/service.sh
+++ b/app/src/main/assets/root/service.sh
@@ -13,7 +13,7 @@ sleep 5
 
 base_path="$DIR/$package_name.apk"
 stock_path="$(pm path "$package_name" | grep base | sed 's/package://g')"
-stock_version="$(dumpsys package "$package_name" | grep versionName | cut -d "=" -f2)"
+stock_version="$(dumpsys package "$package_name" | grep versionName | head -n1 | cut -d "=" -f2)"
 
 echo "base_path: $base_path"
 echo "stock_path: $stock_path"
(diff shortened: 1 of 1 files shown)


oSumAtrIX:
> stock_version might return several packages so only take the first result (newest one).

When would it return several packages? In case of system-installed and user-installed updates?
Is the first one always the newest one? I suppose Android also always uses the " newest " one? If not, perhaps it uses some "active" one? If so, what is considered the currently active one? The first one?

Naim2000:
Hi, came here to bump since I got this issue and quickly landed on an entirely identical solution (ref: the `head -n1`)

> When would it return several packages? In case of system-installed and user-installed updates?

Well, looking at the output from `dumpsys package`, it lists `Packages:` (currently installed version) and then `Hidden system packages:` ("factory default" version)

> Is the first one always the newest one? I suppose Android also always uses the " newest " one? If not, perhaps it uses some "active" one? If so, what is considered the currently active one? The first one?

`Packages` gets listed before `Hidden system packages`, so most likely yes.
although `Packages` is plural, I have no clue whether multiple packages getting listed here is a reality.. maybe split APKs or whatever?
And, either way, it's between grep pulling multiple versions that will never match one version, or pulling one version that might not match. Nothing is being lost here, I think

duckiye05-pixel:
@S4adam I would like to ask why this has been removed?
`chcon u:object_r:apk_data_file:s0 "$base_path"`

mine just crashed without it

Issue #1993, opened 2024-06-22
feat: Parse app info in the app info screen when selecting a downloadable app


oSumAtrIX:
### Feature description

Right now the package name and a generic icon is shown:

![image](https://github.com/ReVanced/revanced-manager/assets/13122796/4fcd99b0-44f5-418d-a18e-7562e56aced6)

I believe one the app was found online , it doesn't hurt to parse the metadata such as the icon or title for it. 

### Motivation

-

### Acknowledgements

- [X] This issue is not a duplicate of an existing feature request.
- [X] I have chosen an appropriate title.
- [X] The feature request is only related to ReVanced Manager

KobeW50:
When the app is an installed app, this can be done even before the app is found online. The app icon and title are already shown in the app selection screen.

https://github.com/user-attachments/assets/5367931d-62c8-40db-9844-22391a1a1bc4

PR #3515, opened 2026-09-01
feat: parse downloadable app metadata

keeltrace:
## Why

Downloadable apps currently fall back to a package name and generic icon because no APK metadata is available on the app info screen.

## What

- Cache the APK when a downloader resolves an app so Manager can parse its real icon, label, and version.
- Reuse the cached APK during patching instead of downloading it twice.
- Keep the existing downloader source identity and fall back to installed package metadata if the cache disappears.

## Testing

- `git -c core.whitespace=cr-at-eol diff --check`
- `./gradlew :app:assembleRelease`

Closes #1993


keeltrace:
Thanks for all the hard work yall put into revanced; hope these old features requests turned into PR's are useful. Best youtube app ever

